PR #15, opened 2021-06-18
Remove Mutex lock on Threadpool and routes

JackThomson2:
Hi, cool library.

Taking a quick look, I've found some nice wins here which means we don't have to lock at all as we don't access &mut self anymore. I've added the Dashmap library for the concurrent hashmap. 

Any ideas let me know

Cheers Jack


sansyrox (approved):
This looks really good! Thank you! 🚀

PR #17, opened 2021-06-19
Pure tokio

JackThomson2:
Here's another proposal which massively reduces the complexity of the project by leveraging Tokio's multi-threaded runtime

Would probably need some benching marking on your machine, for me it appears to be faster.

Jack


JackThomson2:
In my tests this is 2x as fast

sansyrox:
Hey @JackThomson2 , Thank you for the contribution! 😄 
I will try to run the benchmarks on my machine and will confirm regarding the tokio runtime. But I have also heard that tokio works better than async-std.

However, I do have a question regard this implementation. I created a separate thread for the TCPListener as I am planning to introduce multiple processes as well in the future. 
I planned to share the stream object among multiple process(this is just a plan atm) . What according to you will having everything in the main thread change in this process?

sansyrox:
> In my tests this is 2x as fast

@JackThomson2 , what kind of tests are you using?

JackThomson2:
Hey! 

This should work with multiple processes, the tokio spawn will share the load between the different cores, if you where to use this multiple times it should be fine and tokio should just work it all out! For testing I was timing your ping script and simply testing response times on chrome as well. Very crude tests but looked positive. Any questions let me know

JackThomson2:
@sansyrox I've just pushed an update which will stop slow clients, by spawning on recieve.

sansyrox:
> This should work with multiple processes, the tokio spawn will share the load between the different cores, if you where to use this multiple times it should be fine and tokio should just work it all out!

Just a clarification, now with tokio added, we won't have to spawn a process(something like this https://doc.rust-lang.org/std/process/struct.Command.html) to spread the request across different cores. Tokio will handle it automatically?
If that is the case, I will try to learn about tokio in more detail.

> @sansyrox I've just pushed an update which will stop slow clients, by spawning on receive.

Thank you so much! I just saw it, it looks good! 🚀

JackThomson2:
Yes exactly! A better explanation can be found here https://docs.rs/tokio/1.7.1/tokio/runtime/index.html#multi-thread-scheduler but basically it uses a work stealing algorithm between cores for scheduling.

sansyrox:
@JackThomson2 , this looks really good. Can you change this to a normal PR so we can review it.

- Can you change `src/threadpool.rs` to a different name?

sansyrox:
@JackThomson2 , one last thing before we merge this PR. Can you please fetch the latest changes from the main branch?

JackThomson2:
@sansyrox Won't merging automatically rebase onto main or am I missing something?

sansyrox:
@JackThomson2 , I was worried that I was just showing old changes. I'll merge and check.

sansyrox (approved):
LGTM! 🚀
Amazing work! ✨

JackThomson2:
:raised_hands:

PR #20, opened 2021-06-21
Add support for sync functions

sansyrox:
Method 1 support for sync functions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -575,9 +575,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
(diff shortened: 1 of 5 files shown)


sansyrox:
This approach still allows non-blocking execution of code. 

This will block the sync thread when multiple sync functions are called and while the sync thread is blocked, async functions will still continue to progress as desired.

Future Improvement: reintroduce the threadpool for sync threads!

sansyrox:
Accepted this. Rejected (https://github.com/sansyrox/robyn/pull/21)

PR #21, opened 2021-06-21
Add support for sync functions | Facade Approach

sansyrox:
This is another approach for adding support for async functions


sansyrox:
This approach merely creates a facade of an async function and the code will still be blocking, whereas in (https://github.com/sansyrox/robyn/pull/20) we have a segregation of both kinds of functions and we can spawn sync functions on different threads.

sansyrox:
Rejected this for https://github.com/sansyrox/robyn/pull/20

PR #26, opened 2021-06-24
Build and publish wheels on GitHub Actions

messense:
You need to setup the `PYPI_PASSWORD` secret, use a [api token](https://pypi.org/help/#apitoken) is recommended.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


messense:
Example run: https://github.com/messense/robyn/actions/runs/967265751

sansyrox (approved):
@messense , this looks amazing! 😄 

I will create a stable branch off my main. Can you add a condition such that it only triggers when  pushed on `stable`?

messense:
> I will create a stable branch off my main. Can you add a condition such that it only triggers when pushed on `stable`?

Sure. 
But I don't think it's really needed, currently it builds the wheels on push and PR, only publishes them to PyPI on new git tag push.

messense:
@davidhewitt There is a weird GIL related error on Python 3.6, I'm not sure why, it works on Python 3.7 and later.
See https://github.com/messense/robyn/runs/2902786370?check_suite_focus=true

sansyrox:
> But I don't think it's really needed, currently it builds the wheels on push and PR, only publishes them to PyPI on new git tag push.
@messense , it makes sense then. Thank you for the amazing contribution. I will be merging this PR! 🚀

sansyrox:
Hey @messense ,
Thank you for the amazing work. We just released the latest version at : https://pypi.org/project/robyn/0.4.1/

davidhewitt:
> Maybe related to the use of prepare_freethreaded_python? 

Possibly. There was a lot of changes to python initialization in Python 3.7.

We also changed the code a lot for PyO3 0.14, so if it works for "future" versions then I'm not too worried.

PR #28, opened 2021-06-29
Add asgi support

sansyrox:
This is my first attempt at introducing asgi support. We'll see how this works out.


sansyrox:
This is just a reference. No plans to merge it in the future.

PR #34, opened 2021-07-01
General tidy up

JackThomson2:
I've made a few changes:

 - Clippy compiles with no warnings now
 - Removed the use of unwrap when executing functions, means we don't shutdown if the python runtime errors.
 - More efficient fastpath less need for checking un-needed options around python functions.
 - Match route against hyper route enum == faster & case safe
 - Removed the non standard UPDATE header https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
 

Thanks Jack 🚀


JackThomson2:
Also why are we pushing Wheels on a PR,  could be misunderstanding but this could be very dangerous, no?

sansyrox:
> Also why are we pushing Wheels on a PR, could be misunderstanding but this could be very dangerous, no?
@JackThomson2 , we are not pushing the wells on the PR. These are just build checks. The wheels are only pushed on tag commits.

sansyrox:
Thank you for the PR 😄  I will have a look at it by tonight. (IST)

JackThomson2:
I see, it looked like here it was actually uploading something 

image

src/processor.rs
```diff
@@ -1,69 +1,53 @@
-use tokio::io::AsyncWriteExt;
-use tokio::net::TcpStream;
+use anyhow::Result;
```
sansyrox, line 1: How is `anyhow::Result` different from normal `Result`?

src/processor.rs
```diff
@@ -1,69 +1,53 @@
...
-        Some(_) => {}
-    }
+    };
 
-    let function = function.unwrap();
+    Ok(Response::new(Body::from(contents)))
+}
 
-    let contents = match function {
+#[inline]
```
sansyrox, line 28: One more question, what does an `#[inline]` do in rust?

test.py
```diff
@@ -3,10 +3,14 @@
 
 app = Robyn()
 
+callCount = 0
```
sansyrox, line 6: Why did you add a counter here??

sansyrox (commented):
Hey @JackThomson2 , 
Thank you for the great PR. But I do have some questions before we merge it.
Can you please answer the inline comments?

src/processor.rs
JackThomson2: It captures any of the error types be that: python errors or unwrap errors etc. so it means I can capture them all and we can capture them to display a nice error rather than crashing

src/processor.rs
JackThomson2: It hints to the compiler that the assembly should be placed inside the caller function, this can make calling functions faster especially if in a commonly called function. The slight cost is a marginally larger binary.

test.py
JackThomson2: It was to demo the number of calls increasing and checking the GIL lock was acquiring the global parameters correctly. Was more for a personal testing scenario, could be reverted if needed

sansyrox: No no. That's alright.

sansyrox:
Merged!

PR #36, opened 2021-07-02
Add the basics to add serving of static files

sansyrox:
First implementation to enable the serving of static files

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+  </head>
+  <body>
+    Hello world. How are you?
+  </body>
+</html>
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -1,5 +1,6 @@
 from .robyn import Server
 from asyncio import iscoroutinefunction
+from .responses import static_file
 
 class Robyn:
     """This is the python wrapper for the Robyn binaries.
@@ -8,6 +9,8 @@ class Robyn:
         self.server = Server()
 
     def add_route(self, route_type, endpoint, handler):
+        """ We will add the status code here only
+        """
         self.server.add_route(route_type, endpoint, handler, iscoroutinefunction(handler))
 
     def start(self, port):
@@ -38,8 +41,3 @@ class Robyn:
         def inner(handler):
             self.add_route("PATCH", endpoint, handler)
         return inner
-
-
-
-
-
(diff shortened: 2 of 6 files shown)


JackThomson2:
Just an idea can't we load the file from rust and not touch python at all, I.e add app.serve("localpath", "publication")

sansyrox:
> Just an idea can't we load the file from rust and not touch python at all, I.e add app.serve("localpath", "publication")

@JackThomson2 , I can see three problems in this, do correct me if I am wrong:

1. How will you check the routing ?
2. It will also not allow us to give html response on HTTP methods apart from GET.
3. Non standard syntax and how will we add support for multiple HTML files?

JackThomson2:
Hi so,

1. Routing is given in a way you pass in location on disk and location on api e.g. ```app.serve("./static/index.html", "/")``` We would also need to hand mounting a whole directory so ```app.serve("./static/", "/")``` would load the whole static directory
2. We could add an overload to the function to accept different requests, however that seems an unusual use-case.
3. May be unusual from a python developers POV, but the benefits are massive, we'd be able to achieve true parallelism as we would never touch python. And for handling multiple files I mentioned above in 1. We will allow them to call serve multiple times as we just store them in another concurrent map

Jack

sansyrox:
I can definitely understand the benefits of the approach mentioned by you. The decorator syntax is really loved in the python ecosystem.

We can implement both of the approaches. The current approach(including python) will be used to serve smaller single html files and it will also be used to serve json responses.

We will also allow the way you suggested to serve larger html files and mounting directories.

> We could add an overload to the function to accept different requests,

Can you please share a snippet on how overloading is done in rust? I could not find it anywhere.

> we should use the built-in async file reading so it wont block

Are you talking about this? https://docs.rs/async-fs/1.5.0/async_fs/

JackThomson2:
Interesting, I've never really seen HTML returned from a POST request, I always thought POST as submitting data to the server and GET getting data from the server. For overloading, i tend to have two methods with similar names and the one with less parameters fills in the default when calling the other. I guess this may be more appropriate as we're using tokio: https://docs.rs/tokio/1.8.0/tokio/fs/index.html

Jack

JackThomson2:
Dont you need at be setting correct MIME types, as a browser won't like static files returned this way

sansyrox:
@JackThomson2 , even though it worked well on my browser, I believe that it will be a good practice if we set mime types.

Screenshot 2021-07-05 at 3 20 04 PM

JackThomson2:
I think errors would start to appear when using JS or other file types.

sansyrox:
> I think errors would start to appear when using JS or other file types.
That could be true. I will implement it in a new PR.

sansyrox:
Also, will we have to manage application/json and text/json separately ?

PR #44, opened 2021-07-09
Serving static file using async functions

### LINKED ISSUE #43: Try to serve individual static files using vanilla rust
I think I may know a way to do that.

Explore the possibility to do that.

sansyrox:
Fixes #43

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,8 +5,11 @@ mod types;
 
 use server::Server;
 
+use tokio::fs;
+
 // pyO3 module
 use pyo3::prelude::*;
+use pyo3::types::PyString;
 use pyo3::wrap_pyfunction;
 
 #[pyfunction]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -16,11 +19,26 @@ pub fn start_server() {
     Server::new();
 }
 
+#[pyfunction]
+pub fn async_static_files(py: Python, file_name: String) -> PyResult<PyObject> {
+    pyo3_asyncio::tokio::into_coroutine(py, async move {
+        let contents = fs::read(file_name.clone()).await.unwrap();
+        let foo = String::from_utf8_lossy(&contents);
+        Ok(Python::with_gil(|py| {
```
sansyrox, line 27: Surprisingly this approach is coming out to be slower than the python response.
This is probably because of locking and unlocking of gil.

sansyrox:
This approach is slower even for large files.

But this approach helped me find that the html service files are not robust enough.
```
fn read_file(file_path: &str) -> String {
    let mut file = File::open(file_path).unwrap();
    let mut buf = vec![];
    file.read_to_end(&mut buf).unwrap();
    String::from_utf8_lossy(&buf).to_string()
}

```

This should be the real implementation.

sansyrox:
This was a bad idea. Clearly slower in all the cases. It helped us identify a flaw so not too bad.

We will add the option to mount the folder once we switch to actix.

PR #60, opened 2021-08-02
Add the ability to add body in PUT, PATCH and DELETE

sansyrox:
Need to test it before merging. PyO3 not compiling on my machine.

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -2,7 +2,6 @@
 import os
 import argparse
 import asyncio
-import inspect
 
 from .robyn import Server
 from .responses import static_file, jsonify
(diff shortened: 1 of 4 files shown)


robyn/__init__.py
```diff
@@ -129,6 +135,12 @@ def patch(self, endpoint):
         :param endpoint [str]: [endpoint to server the route]
         """
         def inner(handler):
+            sig = inspect.signature(handler)
+            params = len(sig.parameters)
+            if params != 1:
```
sansyrox, line 140: On further thought, this should not be a mandate.
One should be able to create a route without worrying about the params.

PR #61, opened 2021-08-02
Create a request object

sansyrox:
This request creates a request object instead of referring to individual variables.

Also, set basis for middleware additions in the future.


JackThomson2:
We should probably build the Hashmap outside the GIL to try and keep concurrency as high as possible

sansyrox:
@JackThomson2 , yes absolutely. I will make the changes.

PR #64, opened 2021-08-04
Add the static directory serving

sansyrox:
Integrate static directory serving.


src/server.rs
```diff
@@ -1,9 +1,10 @@
 use crate::processor::{apply_headers, handle_request};
 use crate::router::Router;
 use crate::types::Headers;
+use actix_files::Files;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::{Relaxed, SeqCst};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
sansyrox, line 7: Use https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html instead of a generic mutex

sansyrox: Fix ordering of imports

PR #70, opened 2021-08-08
Pyo3 update

awestlake87:
Hey, the fix was a bit more complicated than I anticipated. Essentially, since you're using `run_forever`, there was no way of getting the event loop reference via `get_running_loop` as opposed to `get_event_loop` (which will fetch the global event loop). I went ahead and scoped the event loop for your `index` handler so that it has access to the event loop via task-local data.

I've been improving the docs on 0.14, but this problem probably isn't explicitly addressed. The "Event Loop References and Thread-Awareness" section in the docs may provide some intuition for why 0.14 acts this way.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -12,8 +12,6 @@ use pyo3::prelude::*;
 use pyo3::types::PyAny;
 
 // hyper modules
-use pyo3_asyncio::run_forever;
-
 static STARTED: AtomicBool = AtomicBool::new(false);
 
 #[pyclass]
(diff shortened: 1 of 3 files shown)


src/server.rs
```diff
@@ -44,16 +42,29 @@ impl Server {
         let router = self.router.clone();
         let headers = self.headers.clone();
 
+        let asyncio = py.import("asyncio").unwrap();
+
+        let event_loop = asyncio.call_method0("new_event_loop").unwrap();
```
ShadowJonathan, line 47: are we sure this is on a python thread? (that will keep executing the event loop with the control on python's side, before we start throwing tasks into it)

ShadowJonathan: Oh nevermind, i just saw the `run_forever` pass to python and then asyncio down below, yeah, that should make this alright.

src/server.rs
```diff
@@ -44,16 +42,29 @@ impl Server {
...
+            .unwrap();
+        let event_loop_hdl = PyObject::from(event_loop);
+
         thread::spawn(move || {
             //init_current_thread_once();
             actix_web::rt::System::new().block_on(async move {
                 let addr = format!("127.0.0.1:{}", port);
 
                 HttpServer::new(move || {
+                    let event_loop_hdl = event_loop_hdl.clone();
```
ShadowJonathan, line 59: Why not just move the value? (and not clone it)

awestlake87: `HttpServer::new` expects a `Fn` not a `FnOnce`, so when `event_loop_hdl` gets moved into the router closure later on, `event_loop_hdl` needs to remain valid for the next time the `HttpServer::new` closure is called.

ShadowJonathan: Ah, didn't know that, then this LGTM 👍

sansyrox:
Thank you @awestlake87 for this. Do the changes even work with `v0.13.*` ?

awestlake87:
The 0.14 API is pretty different from 0.13, so it won't work with both. If you need it to work with both 0.13 and 0.14, you can delay this PR and continue using the 0.13 API with `#[allow(deprecated)]` since the 0.13 conversions work the same in 0.13 and 0.14.

sansyrox:
@awestlake87 , no no. I was asking as the Cargo.toml was not updated. I will update it and check if the tests pass.

sansyrox:
@awestlake87 , what does `pyo3-asyncio::tokio::scope` do?

awestlake87:
It allows the conversions to get a reference to the current event loop via task-local data. I've got a breakdown of its motivation and usage [here](https://docs.rs/pyo3-asyncio/0.14.0/pyo3_asyncio/#event-loop-references)

ShadowJonathan:
More specifically;

> `pyo3_asyncio::<runtime>::scope` - Store the event loop in task-local data when executing the given Future.

> `pyo3_asyncio::<runtime>::scope_local` - Store the event loop in task-local data when executing the given !Send Future.

PR #77, opened 2021-08-21
Add default port and a variable url

sansyrox:
This PR adds a default port and the possibility of a variable URL, even 0.0.0.0


sansyrox:
Works well!

PR #78, opened 2021-08-23
Add the base of http requests

sansyrox:
This pr adds the base of http compliance. Need to add more specifications per route!

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -130,3 +130,48 @@ class Robyn:
            
         return inner
 
+    def head(self, endpoint):
+        """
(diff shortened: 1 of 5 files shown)


sansyrox:
@JackThomson2 @messense , I tried adding directory serving in actix here(https://github.com/sansyrox/robyn/pull/78/files#diff-48d27698cc708c7efe770fd897e4885efaa0a15cae30d54936068603ba03bbd9R71) . However, it is not allowing me serve request apart from GET. 

Do you folks know where I might be going wrong?

PR #82, opened 2021-09-06
Add the basics of python testing

sansyrox:


sansyrox:
Add requests in the requirements!

sansyrox:
I am thinking of wrapping it inside a docker-compose file. Testing will be simpler and I am more familiar with that syntax than testing with circleCI

PR #128, opened 2021-12-11
Add help messages and simplify 'dev' option

Kludex:
This PR adds the help messages (my own choice of words, but feel free to modify).

What's the difference between processes and workers here? I didn't inspect the code yet, but in the ASGI context it doesn't make sense.

Also, I'd recommend for us to use Click. It will make the writing of the CLI more pleasant.

### DIFF
--- a/robyn/argument_parser.py
+++ b/robyn/argument_parser.py
@@ -1,13 +1,35 @@
 import argparse
 
+
 class ArgumentParser(argparse.ArgumentParser):
     def __init__(self):
-        self.parser = argparse.ArgumentParser(description="Robyn, a fast async web framework with a rust runtime.")
-        self.parser.add_argument('--processes', type=int, default=1, required=False)
-        self.parser.add_argument('--workers', type=int, default=1, required=False)
-        self.parser.add_argument('--dev', default=False, type=lambda x: (str(x).lower() == 'true'))
+        self.parser = argparse.ArgumentParser(
+            description="Robyn, a fast async web framework with a rust runtime."
+        )
+        self.parser.add_argument(
+            "--processes",
+            type=int,
+            default=1,
+            required=False,
+            help="Choose the number of processes. [Default: 1]",
+        )
+        self.parser.add_argument(
+            "--workers",
+            type=int,
+            default=1,
+            required=False,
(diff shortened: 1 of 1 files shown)


robyn/argument_parser.py
```diff
@@ -1,13 +1,35 @@
...
+            "--workers",
+            type=int,
+            default=1,
+            required=False,
+            help="Choose the number of workers. [Default: 1]",
+        )
+        self.parser.add_argument(
+            "--dev",
+            dest="dev",
+            action="store_true",
```
sansyrox, line 26: Thank you for contributing. But can you tell me what these flags do?

sansyrox (commented):
These changes look good but I have a question with the contribution.

robyn/argument_parser.py
Kludex: Yes! So, the change here is removing the `dev` in the arguments and add in the options. The reason is that no one will ever do `--dev=false` as it's the default.

This change just makes `--dev` standalone possible. The `action` argument on the highlighted above makes it store a `True` value on the `dev` variable.

sansyrox: Oh, alright. Makes sense. Thank you! :D

sansyrox:
> What's the difference between processes and workers here? I didn't inspect the code yet, but in the ASGI context it doesn't make sense.

Workers are the `tokio`s way of implementing concurrency. This is mostly threads.

Whereas is the processes are the sharing the listening TCP socket across multiple python processes. So, you can imagine having multiple runtimes running in parallel.

This is still experimental to see if they add any performance benefits or not. That's why the defaults are set to 1 at the moment.

sansyrox:
> Also, I'd recommend for us to use Click.

This is a great suggestion. I didn't know about this library. I will integrate this in the coming releases. :D

Kludex:
I see. On `uvicorn` workers and processes are analogous. We probably want to mention that somewhere at some point. :+1:

sansyrox:
> I see. On `uvicorn` workers and processes are analogous. We probably want to mention that somewhere at some point. 👍

Thanks for the feedback. Will do 🖖

sansyrox (approved):
LGTM! 🚀

PR #134, opened 2021-12-15
Add async support in WS

### LINKED ISSUE #116: Async web socket support
This answer on stack overflow looks helpful.

https://stackoverflow.com/questions/64434912/how-to-correctly-call-async-functions-in-a-websocket-handler-in-actix-web

sansyrox:
**Description**

This PR fixes #116

### DIFF
--- a/src/web_socket_connection.rs
+++ b/src/web_socket_connection.rs
@@ -81,21 +81,27 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for MyWs {
                         println!("{:?}", handler);
                         let fut = Python::with_gil(|py| {
                             let handler = handler.as_ref(py);
-                            pyo3_asyncio::tokio::into_future(handler.call0().unwrap()).unwrap()
+                            let coro = handler.call0().unwrap();
+                            pyo3_asyncio::tokio::into_future(coro)
                         });
-                        tokio::spawn(async {
-                            let output = fut.await.unwrap();
+                        // let fut = async move {
+                        //     println!("Hello world");
+                        // }
+                        // tokio::spawn(async {
+                        //     let output = fut.await.unwrap();
 
-                            Python::with_gil(|py| {
-                                println!("Bruhhh moment");
-                                let contents: &str = output.extract(py).unwrap();
-                                println!("{:?}", contents);
-                            });
-                        });
-                        let f = async move {};
-                        let fut = actix::fut::wrap_future::<_, Self>(f);
-                        // fut.spawn(self);
-                        ctx.spawn(fut);
+                        //     Python::with_gil(|py| {
+                        //         println!("Bruhhh moment");
+                        //         let contents: &str = output.extract(py).unwrap();
+                        //         println!("{:?}", contents);
+                        //     });
+                        // });
+                        // let f = async move {};
+                        let f = async move {
+                            fut.unwrap().await.unwrap();
+                        };
+                        let x = actix::fut::wrap_future::<_, Self>(f);
+                        ctx.spawn(x);
                         // ctx.add_stream("Hello, world");
                         return ctx.text("Async Functions are not supported in WS right now.");
                     }
--- a/src/web_socket_connection.rs
+++ b/src/web_socket_connection.rs
@@ -1,6 +1,7 @@
 use crate::types::PyFunction;
 
-use actix::{Actor, StreamHandler};
+use actix::prelude::*;
+use actix::{Actor, AsyncContext, StreamHandler};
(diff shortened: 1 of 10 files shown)


src/web_socket_connection.rs
```diff
@@ -70,9 +75,34 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for MyWs {
...
                     }),
-                    PyFunction::CoRoutine(_handler) => {
-                        println!("Async functions are not supported in WS right now.");
+                    PyFunction::CoRoutine(handler) => {
+                        println!("{:?}", handler);
+                        let fut = Python::with_gil(|py| {
+                            let handler = handler.as_ref(py);
+                            let coro = handler.call0().unwrap();
+                            pyo3_asyncio::tokio::into_future(coro)
+                        });
```
sansyrox, line 86: Hi @awestlake87 ,

I hope everything is going well.

I trying to convert a python coroutine to a rust future and I am getting the following error:


```
sys:1: RuntimeWarning: coroutine 'connect' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
thread 'actix-rt|system:0|arbiter:0' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'RuntimeError'>, value: RuntimeError('no running event loop'), traceback: None }', src/web_socket_connection.rs:101:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I am using the same method for conversion as I am doing here: https://github.com/sansyrox/robyn/blob/main/src/processor.rs#L84

I am unable to figure out why this is giving this error.

Can you please help? 

Thanks in advance.

ShadowJonathan:
I would not recommend doing ctx.spawn, websocket messages may be handled out of order due to scheduling.

sansyrox:
> I would not recommend doing ctx.spawn, websocket messages may be handled out of order due to scheduling.

@ShadowJonathan , `ctx.spawn` is being used for async functions. So, ig out of order execution may work. 
Do you have any other suggestions otherwise?

ShadowJonathan:
It's mostly so that message B doesn't get processed after message A, I suggest doing the `.await` on the asyncio coroutine inside the ws object handler function.

sansyrox:
@ShadowJonathan , I tried converting this function(https://github.com/sansyrox/robyn/pull/134/files#diff-0d0da7574dadb5452e2f359138c914a901f38425138f1c2facc37f403b98ab7fR38) to an async function but it is not supported by Actix at the moment :(

ShadowJonathan:
Ooooh wait, that explains it some more, yes, then this approach should be fine enough.

Maybe warn users that websocket message handling can be concurrent, and doesn't wait for each handler to complete before spawning the next one.

*Or*, you could maybe use [`wait`](https://docs.rs/actix/0.12.0/actix/trait.AsyncContext.html#tymethod.wait) instead.

sansyrox:
> Ooooh wait, that explains it some more, yes, then this approach should be fine enough.

Perfect!

>Maybe warn users that websocket message handling can be concurrent, and doesn't wait for each handler to complete before spawning the next one.

This makes sense. If someone wants to have synchronous execution, they can use regular sync functions whereas the async functions can be used when the order does not matter. This would give a more concurrent feel imo.

ShadowJonathan:
(I recommend trying out that `wait` function though, it seems plenty perfect enough for your usecase)

sansyrox:
> (I recommend trying out that wait function though, it seems plenty more perfect for your usecase)

But won't it block the execution of other calls while the current call is completed?

ShadowJonathan:
As far as i see it, it blocks the current task until the asyncio future is completed, so it'll wait until the python async function has handled this message first before it moves on to the next.

Documentation says the actor wont receive any new messages during this time.

sansyrox:
> Documentation says the actor wont receive any new messages during this time.

Hmm.. This is something I didn't want to happen for `async` functions in WebSockets. Otherwise, I would've used `pyo3-asyncio::tokio::run_until_complete` method. 

I am taking inspiration from expressjs or fastifyjs where async functions may execute out of order to provide a more non-blocking experience. 

Sync functions are always executed in order.

ShadowJonathan:
Just looked up a bunch more documentation from actix, i think it's fine to `wait`, the actor will receive messages inbetween these `handle` calls, i think, but it is up to you.

If you're doing `spawn`, then i really recommend noting this down in documentation.

sansyrox:
Now, that I think if the order is really important in async functions. I can just add a param called `non_blocking=False`

PR #144, opened 2022-01-01
Fix dev server

sansyrox:
**Description**

This PR fixes #

This doesn't work atm. We will have to figure out a different approach or read the docs better.


sansyrox:
Old

PR #146, opened 2022-01-03
URL queries

patchgamestudio:
**Description**

This PR adds support for URL form queries, which do not match REST-like patterns such as `/path/<id>`. The Rust side already has access to query strings for requests to paths such as `/path?foo=bar`. This data is made available on the Python side as a dict object similar to headers under the key name `queries`.

Docs and tests have been updated to reflect changes.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,2 +1,3 @@
 flake8==4.0.1
 black==21.12b0
+websockets==10.1
--- a/src/server.rs
+++ b/src/server.rs
@@ -235,11 +235,14 @@ async fn index(
     mut payload: web::Payload,
     req: HttpRequest,
 ) -> impl Responder {
-    let split = req.query_string().split("&");
     let mut queries = HashMap::new();
-    for s in split {
-        let params = s.split_once("=").unwrap_or(("", ""));
-        queries.insert(params.0, params.1);
+    
+    if req.query_string().len() > 0 {
+        let split = req.query_string().split("&");
+        for s in split {
+            let params = s.split_once("=").unwrap_or((s, ""));
+            queries.insert(params.0, params.1);
+        }
     }
 
     match router.get_route(req.method().clone(), req.uri().path()) {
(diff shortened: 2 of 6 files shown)


dev-requirements.txt
```diff
@@ -1,2 +1,3 @@
 flake8==4.0.1
 black==21.12b0
+websockets==10.1
```
sansyrox, line 3: Why did you add this here?

sansyrox (commented):
Hi @patchgamestudio ,

Thank you for your contribution 😄 

I have a few inline comments, the PR looks good otherwise.

src/server.rs
```diff
@@ -234,6 +235,16 @@ async fn index(
...
+    let mut queries = HashMap::new();
+    
+    if req.query_string().len() > 0 {
+        let split = req.query_string().split("&");
+        for s in split {
+            let params = s.split_once("=").unwrap_or((s, ""));
+            queries.insert(params.0, params.1);
+        }
+    }
+
```
sansyrox, line 247: Do we want the queries hashmap to be available in every HTTP request or just the GET requests methods?

I usually use this kind of params in GET requests only. But I am not sure about the standard of WEB requests.

dev-requirements.txt
patchgamestudio: While trying to build and test robyn, I ran into an issue where the `websockets` dependency hadn't been met. I looked around, and it gets called in `robyn/integration_tests/test_web_sockets.py`. Since it isn't necessary for normal operation, it seemed more appropriate to add it to `dev-requirements.txt`. I suppose it isn't really necessary for this to be included for the feature, but I felt like it was worth slipping it in to help with long term CI/CD.

src/server.rs
patchgamestudio: Personally, I think it's generally only necessary for GET requests. That being said, this project doesn't exist for just me. I expect developers will want to use features like this in unorthodox ways. There didn't seem to be a good reason to restrict queries solely to GET requests, so I left it open to all types.

dev-requirements.txt
sansyrox: Ah, that makes sense. Thank you! :D

src/server.rs
sansyrox: > I expect developers will want to use features like this in unorthodox ways

Absolutely! Robyn is based on making unorthodox tradeoffs for better performance. I love it. :D

sansyrox (approved):
Great work! :D

PR #149, opened 2022-01-11
Use print instead of input after starting server

klaa97:
**Description**

This super short PR fixes the usage of `input` after starting the server and replaces it with `print`.

The main problem solved is the necessity of having a TTY attached in case we use `input` - for example, with the previous `input`, Docker was failing if running without interactive mode.

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -88,7 +88,7 @@ class Robyn:
                 )
                 p.start()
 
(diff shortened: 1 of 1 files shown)


sansyrox (approved):
LGTM! 🚀 

Great work! ✨

sansyrox:
@klaa97 , just created a new release of robyn for this. https://pypi.org/project/robyn/

PR #154, opened 2022-01-15
Add Event handlers

### LINKED ISSUE #153: Consider adding startup events
What do you think about the idea of adding a startup event? To make an example, I may wanna initialize a connection pool to my database at startup, and re-use connections across requests. I cannot do that though in Python, since the asyncio loop running will not be the same as the one used to process the requests... a possible solution is to register a Pyo3 function in the Rust server during initialization in Py, and call it before you start the webserver

Some examples of how it is done in the py servers: https://www.starlette.io/events/

Originally posted by @klaa97 [here](https://gitter.im/robyn_/community#:~:text=5-,replies,-Sanskar%20Jethi%20%40sansyrox).

sansyrox:
**Description**

This PR fixes #153

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -21,13 +21,12 @@ mp.allow_connection_pickling()
 
 
 class Robyn:
-    """This is the python wrapper for the Robyn binaries.
-    """
+    """This is the python wrapper for the Robyn binaries."""
+
     def __init__(self, file_object):
         directory_path = os.path.dirname(os.path.abspath(file_object))
         self.file_path = file_object
         self.directory_path = directory_path
-        self.server = Server(directory_path)
         self.parser = ArgumentParser()
         self.dev = self.parser.is_dev()
         self.processes = self.parser.num_processes()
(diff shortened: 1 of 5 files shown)


sansyrox:
@klaa97 , this PR adds events in Robyn. Could you please have a look at this PR once?

robyn/__init__.py
```diff
@@ -51,25 +51,44 @@ def add_route(self, route_type, endpoint, handler):
...
         self.web_sockets[endpoint] = ws
 
-    def start(self, url="127.0.0.1", port=5000):
+    def _add_event_handler(self, event_type: str, handler):
+        print(f"Add event {event_type} handler")
+        if event_type.lower() not in {"startup", "shutdown"}:
+            return
+
+        is_async = asyncio.iscoroutinefunction(handler)
+        if event_type.lower() == "startup":
```
klaa97, line 80: Very minor, but for clarity, I would consider to:
- Extract the "startup"/"shutdown" string either in an Enum (like Class(str, Enum) or as constants
- DIrectly use "startup"/"shutdown" as keys for the dictionary `event_handlers`

klaa97 (approved):
The logic looks perfect to me! 👏

robyn/__init__.py
sansyrox: Thank you. Updated :D

PR #170, opened 2022-03-07
Modify headers

### LINKED ISSUE #167: [BUG] Unable to modify headers in middlewares
**Description**

After request and before request should also have the ability to alter the request. 

**Expected Behavior**

Fix that.

**Screenshots**

**Environment:**
- Host OS: 
- Browser: 

---

---

sansyrox:
**Description**

This PR fixes #167

### DIFF
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -28,7 +28,6 @@ class Router(BaseRouter):
                     "body": res["body"],
                     **res
                 }
-            print("Setting the response", response)
         else:
             response = {"status_code": "200", "body": res, "type": "text"}
 
@@ -37,7 +36,6 @@ class Router(BaseRouter):
     def add_route(self, route_type, endpoint, handler):
         async def async_inner_handler(*args):
             response = self._format_response(await handler(*args))
-            print(f"This is the response in python: {response}")
             return response
 
         def inner_handler(*args):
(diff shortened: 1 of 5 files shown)


robyn/router.py
```diff
@@ -28,16 +28,17 @@ def _format_response(self, res):
                     "body": res["body"],
                     **res
                 }
-            print("Setting the response", response)
         else:
             response = {"status_code": "200", "body": res, "type": "text"}
 
+        if response["body"] is None:
+            response["body"] = {}
+
```
sansyrox, line 36: not needed

src/processor.rs
```diff
@@ -37,7 +37,7 @@ pub fn apply_headers(response: &mut HttpResponseBuilder, headers: &Arc<Headers>)
 pub async fn handle_request(
     function: PyFunction,
     number_of_params: u8,
-    headers: &Arc<Headers>,
+    headers: Vec<HashMap<String, String>>,
```
sansyrox, line 40: maybe, just passing hashmap would work here.

sansyrox: works

robyn/router.py
sansyrox: you were correct

PR #184, opened 2022-04-06
Add experimental io-uring

### LINKED ISSUE #177: Optional build feature for io_uring
**Current Behavior**

Things are fast.

**Desired Behavior**

Things could be even faster 😹

If you haven't read much on io_uring I'd check it out. But what I'm thinking here is there could be an optional feature when building Robyn if you want to use io_uring

Since Robyn is using actix-web and they have an expiremental feature for io_uring you could have a similar feature here.

https://github.com/sansyrox/robyn/blob/main/Cargo.toml#L25
https://github.com/actix/actix-web/blob/09cffc093cd755d09a40a14f073f49015aa6e7ad/actix-web/Cargo.toml#L63

sansyrox:
**Description**

This PR fixes #177 

To run `maturin develop --cargo-extra-args="--features=io-uring"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ If you're feeling curious. You can take a look at a more detailed architecture [
 
 1. Add more routes in the `integration_tests/base_routes.py` file(if you like).
 
-2. Run `maturin develop`
+2. Run `maturin develop` or `maturin develop --cargo-extra-args="--features=io-uring"` (if you want to run the experimental version).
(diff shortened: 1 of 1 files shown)


Kludex:
How were the conclusions about this?

sansyrox:
@Kludex , unfortunately, we didn't see any significant speedups. :(

PR #210, opened 2022-06-13
Add a const router

### LINKED ISSUE #48: [Proposal] Const Requests
I wonder if we should off an option for a constant response which won't change, this would allow us to never touch python for ultra fast path solutions.

sansyrox:
**Description**

This PR fixes #48

### DIFF
--- a/src/executors/mod.rs
+++ b/src/executors/mod.rs
@@ -126,27 +126,14 @@ pub async fn execute_middleware_function<'a>(
 
 pub async fn execute_function(
     function: Py<PyAny>,
-    // queries: Rc<RefCell<HashMap<String, String>>>,
     number_of_params: u8,
     is_async: bool,
 ) -> Result<HashMap<String, String>> {
     let mut request: HashMap<String, String> = HashMap::new();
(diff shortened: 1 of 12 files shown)


src/executors/mod.rs
```diff
@@ -124,6 +124,81 @@ pub async fn execute_middleware_function<'a>(
...
+
+            // this makes the request object to be accessible across every route
+            let coro: PyResult<&PyAny> = match number_of_params {
+                0 => handler.call0(),
+                1 => handler.call1((request,)),
+                // this is done to accomodate any future params
+                2_u8..=u8::MAX => handler.call1((request,)),
+            };
+            pyo3_asyncio::tokio::into_future(coro?)
+        })?;
```
sansyrox, line 157: need to make this part generic for every kind of request.

sansyrox:
@JackThomson2 , I have implemented const requests(https://github.com/sansyrox/robyn/issues/48) . Would you like to have a look at the implementaiton?

Also, thanks again for the great suggestion :D

PR #217, opened 2022-06-27
Remove unnucessary `.pyi`

Kludex:
All of these are useless. You can just add the type annotation on the `.py` files.

When I've mentioned `.pyi` it was meant to be added only on the bindings, but when I suggested, I didn't say it as "do it", it was more like: "check if it makes sense". For the python interface, there's no need for `.pyi`, and for the bindings, I just don't know...


robyn/argument_parser.py
```diff
@@ -38,20 +38,20 @@ def __init__(self):
...
-    def log_level(self):
+    def log_level(self) -> str:
         return self.args.log_level
 
     @property
-    def is_dev(self):
+    def is_dev(self) -> bool:
         _is_dev = self.args.dev
-        if _is_dev and (self.num_processes() != 1 or self.workers() != 1):
+        if _is_dev and (self.num_processes != 1 or self.workers != 1):
```
sansyrox, line 55: Thank you for this suggestion. This is actually a code error and for some reason the code tests did not detect it.

sansyrox: I chose to add types in the library as everyone else loves them.

PR #237, opened 2022-07-16
Add Rust CI

### LINKED ISSUE #236: [Feature Request] add clippy in ci
**Current Behavior**

Floppy not present

**Desired Behavior**

Add clippy 
---
**Screenshots / Mockups**

**Alternatives**

---

AntoineRR:
**Description**
- Fix Rust clippy warnings
- Add a CI for the Rust part of the project, running cargo check, cargo test, cargo fmt, and cargo clippy

**Notes**
- I added more than just clippy, let me know if you want to remove some stuff
- There are two different CI systems in the project : Github Actions and Circle CI. I added the Rust CI to the Github Actions CI because that's what I am the most familiar with. Is there a reason why Circle CI is used for testing the Python part of Robyn? If you prefer having the Rust CI on Circle CI I think I can figure it out and move it there :)

This PR fixes #236

### DIFF
--- a/.github/workflows/rust-CI.yml
+++ b/.github/workflows/rust-CI.yml
@@ -1,0 +1,64 @@
+on: [push, pull_request]
+
+name: Rust Continuous integration
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: check
+
+  test:
+    name: Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+
+  fmt:
+    name: Rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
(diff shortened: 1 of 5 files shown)


.github/workflows/rust-CI.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
```
sansyrox, line 32: @AntoineRR , are these the replacements of the `circle ci`?

These are `cargo test` and `cargo check` , right?

`cargo test` is already being done in the `.circleci/config.yml` file

sansyrox (commented):
Thank you for the contributions! :D

I have an inline comment.

.github/workflows/rust-CI.yml
AntoineRR: Thanks for your feedback :)

From what I understand of CircleCI, you indeed defined a `test` command in `.circleci/config.yml` on line 12, that would run `cargo test`. However, this command is currently not used in any of the workflows. To use the `test` command, it should be added under `steps` inside a `job` (example from the [doc](https://circleci.com/docs/reusing-config#invoking-reusable-commands): the command `sayhello` is invoked inside the `myjob` job, which has to be run by a workflow). As I am not very familiar with this, I may have misunderstood something, so feel free to correct me if that's the case.

I also chose to add the commands to github workflows as it is easier to specify the CI should run on push AND on pull requests. To move `cargo test`, `cargo fmt` and `cargo clippy` to circle CI, I think I would just have to uncomment lines 56-57 of `.circleci/config.yml` to invoke the `lint-test-build` job defined in the rust orb :) Let me know what suits you best!

sansyrox: @AntoineRR , that makes sense. I anyway want to move away from CircleCI. This will be a good first step.

Thanks!

sansyrox (approved):
LGTM! :rocket:

PR #262, opened 2022-08-06
ci: split the ci pipelines for better readability

sansyrox:
**Description**

This PR fixes #


sansyrox:
Split not possible as it `needs` the build artifacts from other file and can't be linked.

PR #278, opened 2022-09-05
fix: add proper kill process to conftest. #249

### LINKED ISSUE #249: [BUG] The --dev mode spawns more servers without clearing previous ones.
**Description**

The `--dev` mode is going in infinite recursion when a file is changed in the directory.

To check, you can do
`lsof -i tcp:5000`

**Expected Behavior**

Handle a base case that might have been missed here.

**Environment:**
- Host OS: every 

---

---

guilefoylegaurav:
**Description**

This PR fixes #249 

### Changes
* fix: fix problem of process persistence post test session
* replace `process.terminate()` with `kill()` equivalent
* scope pytest fixtures to **session** level

### Documentation
No documentation update is required

### How I've tested my work 
* **25/25** tests passed (Debian 10)
* **23/25** tests passed (Windows 10) (`global_index_request()` `dev_index_request()` failed) 
* `lsof -i tcp:5000` outputs nothing post test session

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,40 +1,67 @@
+import signal
+import sys
+from typing import List
 import pytest
 import subprocess
 import pathlib
 import os
 import time
 
+def spawn_process(command: List[str]) -> subprocess.Popen:
+    if sys.platform.startswith("win32"):
+        command[0] = "python"
+        process = subprocess.Popen(command, shell=True, creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
+        return process
+    process = subprocess.Popen(command, preexec_fn=os.setsid)
+    return process 
+    
 
-@pytest.fixture
+
+def kill_process(process: subprocess.Popen) -> None:
+    if sys.platform.startswith("win32"):
+        process.send_signal(signal.CTRL_BREAK_EVENT)
+        process.kill()
+        return     
+    os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+    
+        
+       
+        
+@pytest.fixture(scope="session")
 def session():
     os.environ["ROBYN_URL"] = "127.0.0.1"
     current_file_path = pathlib.Path(__file__).parent.resolve()
     base_routes = os.path.join(current_file_path, "./base_routes.py")
-    process = subprocess.Popen(["python3", base_routes])
+    command = ["python3", base_routes]
+    process = spawn_process(command)
     time.sleep(5)
     yield
-    process.terminate()
+    kill_process(process)
+    
 
 
-@pytest.fixture
+
+@pytest.fixture(scope="session")
 def default_session():
     current_file_path = pathlib.Path(__file__).parent.resolve()
     base_routes = os.path.join(current_file_path, "./base_routes.py")
-    process = subprocess.Popen(["python3", base_routes])
+    command = ["python3", base_routes]
+    process = spawn_process(command)
     time.sleep(5)
     yield
-    process.terminate()
+    kill_process(process)
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def global_session():
     os.environ["ROBYN_URL"] = "0.0.0.0"
     current_file_path = pathlib.Path(__file__).parent.resolve()
     base_routes = os.path.join(current_file_path, "./base_routes.py")
-    process = subprocess.Popen(["python3", base_routes])
+    command = ["python3", base_routes]
+    process = spawn_process(command)
     time.sleep(1)
     yield
-    process.terminate()
+    kill_process(process)
 
 
 @pytest.fixture(scope="session")
@@ -43,8 +70,8 @@ def dev_session():
     os.environ["ROBYN_PORT"] = "5001"
     current_file_path = pathlib.Path(__file__).parent.resolve()
     base_routes = os.path.join(current_file_path, "./base_routes.py")
-    process = subprocess.Popen(["python3", base_routes, "--dev"])
+    command = ["python3", base_routes, "--dev"]
+    process = spawn_process(command)
     time.sleep(5)
     yield
-    process.terminate()
-
+    kill_process(process)


integration_tests/conftest.py
```diff
@@ -1,40 +1,67 @@
...
+from typing import List
 import pytest
 import subprocess
 import pathlib
 import os
 import time
 
+def spawn_process(command: List[str]) -> subprocess.Popen[bytes]:
+    if sys.platform.startswith("win32"):
+        command[0] = "python"
```
sansyrox, line 12: Do you think replacing "python" with "py" might fix the failing windows tests?

guilefoylegaurav: * I tried replaced `python` with `py` - that did not help
* Apparently `python base_routes.py --dev` does not work in Windows 10 as well so the corresponding test obviously fails, because inside **dev_event_handler.py** we have this:

`self.processes.append(subprocess.Popen(["python3", self.file_name], start_new_session=False))`

* I tried adding a conditional that replaces `python3` with `python` in win32 platform - the dev server starts, but then for some reason it fails to find the modules, including Robyn

sansyrox: > Apparently the command python base_routes.py --dev does not work in Windows 10 as well because inside dev_event_handler.py we have this:

This should be the case as we dropped dev server support for windows. But it shouldn't be failing. Can you share the error log please?

integration_tests/conftest.py
```diff
@@ -1,40 +1,67 @@
+import signal
+import sys
+from typing import List
 import pytest
 import subprocess
 import pathlib
 import os
 import time
 
+def spawn_process(command: List[str]) -> subprocess.Popen[bytes]:
```
sansyrox, line 10: @guilefoylegaurav , it should just be `subprocess.Popen` instead of `subprocess.Popen[bytes]` here.

integration_tests/conftest.py
```diff
@@ -1,40 +1,67 @@
...
 
-@pytest.fixture
+
+def kill_process(process: subprocess.Popen[bytes]) -> None:
+    if sys.platform.startswith("win32"):
+        process.send_signal(signal.CTRL_BREAK_EVENT)
+        process.kill()
+        return     
+    os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+    return
```
sansyrox, line 26: can you remove the extra return statement here, please?

integration_tests/conftest.py
```diff
@@ -1,40 +1,67 @@
+import signal
```
sansyrox, line 1: Also, I see a few formatting issues.

Can you please run the pre-commit hook mentioned here: https://github.com/sansyrox/robyn#%EF%B8%8F-to-develop-locally

sansyrox:
@guilefoylegaurav , I hope all's good. Just following up on this PR. Do let me know if you require any more insights from me.

guilefoylegaurav:
My apologies @sansyrox for semi-ghosting you. Was very inappropriate of me. 
Had gone on a trip to Shillong. Returned this Monday.

sansyrox:
@guilefoylegaurav , not a problem :D

I hope you had an amazing trip :sparkles:

integration_tests/conftest.py
guilefoylegaurav: [windows_test.log](https://github.com/sansyrox/robyn/files/9560577/windows_test.log)

This is the log file obtained

guilefoylegaurav: > This should be the case as we dropped dev server support for windows. But it shouldn't be failing. Can you share the error log please?

I suspect that `test_dev_index_request()` would still fail anyways, on Windows?

* Trying to run `python base_routes.py --dev` on Windows 10 does not work (in my machine at least), and throws up errors instead
* `test_dev_index_request()` tries to run the aforementioned command, expects a server to get created, and then tries makes a GET request to the server. However, the command seems to fail.

guilefoylegaurav: Also, `test_global_index_request()` is also failing in Windows. 

```
def test_global_index_request(global_session):
    BASE_URL = "http://0.0.0.0:5000"
    res = requests.get(f"{BASE_URL}")
    assert(os.getenv("ROBYN_URL") == "0.0.0.0")
    assert(res.status_code == 200)
```
* I feel `BASE_URL` should be `http://172.0.0.1` instead? [Reference](url)

>  0.0.0.0 means "all IPv4 addresses on the local machine". It is a meta-address which is non-routable.
>  If you want to access the server locally, i.e. client on the same machine as the server, either use the IP address 127.0.0.1 (loopback Internet protocol) or the equivalent named domain name (localhost)
> r = requests.get('http://127.0.0.1:80')
> r = requests.get('http://localhost:80')

* If this change is made, `test_global_index_request()` would pass in both Windows and Linux. While `requests.get('http://0.0.0.0:8080')` is working fine on Linux, on Windows, it would throw a `requests.exceptions.ConnectionError`.

sansyrox: > I suspect that test_dev_index_request() would still fail anyways, on Windows?

Trying to run python base_routes.py --dev on Windows 10 does not work (in my machine at least), and throws up errors instead (traceback included in the previous message)
test_dev_index_request() tries to run the aforementioned command, expects a server to get created, and then tries makes a GET request to the server. However, the command seems to fail.


@guilefoylegaurav , you're correct about this. This needs to be fixed. I will create an issue for this. Thank you!

sansyrox: > I feel BASE_URL should be http://127.0.0.1 instead? [Reference](https://github.com/sansyrox/robyn/pull/url)

This should be "0.0.0.0" only as this test is indeed for a session that is available throughout the IP. We have the local ip adress test already availble.

sansyrox (approved):
LGTM! :+1:

guilefoylegaurav:
Thanks for approving this PR @sansyrox 
Just before we proceed forward, I wanted to know:
* Would you appreciate the idea of making `test_dev_index_request()` a conditional test for now - say on win32 platform, we skip running this particular test because `--dev` mode does not work in Windows at present?

* What is your say on `test_global_index_request()` failing in win32

sansyrox:
Hey @guilefoylegaurav ,

> Would you appreciate the idea of making test_dev_index_request() a conditional test for now - say on win32 platform, we skip running this particular test because --dev mode does not work in Windows at present?

> What is your say on test_global_index_request() failing in win32

Since these tests were not failing before(read a month back), I don't think that we should be disabling them. Instead, we should be focusing on fixing them. Let be in the PRs.


> Apparently some (2) checks were not successful - should that be a cause of concern?

Those are clippy errors. But I don't know how to fix them at the moment. We integrated clippy in our system a few weeks back. So, still getting used to it :sweat_smile: . If you have any input, I all ears :smile:

guilefoylegaurav:
My bad, my commits were seriously messed up. 
Cleaned the mess up and squashed all that into a single meaningful commit. 

I really apologize for this

sansyrox (approved):
LGTM! 👍

PR #282, opened 2022-09-13
Don't run sync functions in pool

JackThomson2:
Don't run synchronous functions in the tokio threadpool, this is just adding overhead. We need to acquire the GIL anyway so it won't be run async anyway. Server will still accept connections as this is done in a worker thanks to Actix design

This PR fixes #

### DIFF
--- a/src/executors/mod.rs
+++ b/src/executors/mod.rs
@@ -262,27 +262,24 @@ pub async fn execute_http_function(
         }
 
         PyFunction::SyncFunction(handler) => {
-            tokio::task::spawn_blocking(move || {
-                Python::with_gil(|py| {
-                    let handler = handler.as_ref(py);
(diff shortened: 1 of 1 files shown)


sansyrox:
@JackThomson2 , I just had a look at the PR. It looks good.

But I have one question. Does it have any effect on the performance?

JackThomson2:
Yeah it is much faster than previously, I'll make another PR to remove async from the function in techempowered as well should help scores go up massively

sansyrox:
@JackThomson2 , that sounds great. Thank you! :D

I am not sure how Techempowered benchmarks are scored. But if it is not a dynamic request, do  you think const requests would help: https://sansyrox.github.io/robyn/#/features?id=const-requests ?

sansyrox (approved):
LGTM! :sparkles:

PR #286, opened 2022-10-02
Feature add support for env variables

### LINKED ISSUE #97: [Feature] Add support for Env variables and a robyn.yaml config file
**Current Behavior**

Right now, Robyn only allows config variables to be passed as command line arguments. 

**Desired Behavior**

Having it detect a config file at the root of the project will allow a cleaner invocation, which is the ideology of the project.

**High Level Implementation**
- Add a file detector and parser function (ideally on the rust side) and parse the file contents
- Parse the file contents and look for any overrides through the cli
- Set the config variables in the appropriate places
- Start the application.

Shending-Help:
**Description**

there are two files here 
definitions.py => that file's whole purpose is to define where the root of the project is so we can find the config files placed there and access them even in nested files
import_vars.py => which parses the content of the configuration file and then check for overrides and if not it sets the env vars that can be imported and accessed any where

This PR fixes #97


robyn/import_vars.py
```diff
@@ -0,0 +1,27 @@
...
+
+# parse the configuration file returning a list of tuples (key, value) containing the environment variables
+def parser():
+    """Parse the configuration file"""
+    with open(CONFIG_PATH, 'r') as f:
+        for line in f:
+            if line.startswith('#'):
+                continue
+            yield line.strip().split('=')  
+
```
sansyrox, line 12: I would recommend writing the function as the following

```python

def parser(config_path=CONFIG_PATH):
    ...

```

Declaring a function like this will allow it to be more testable and not have us mock the variables if we need to write any unit tests.

This is called reducing coupling. You can have a look at this concept here: https://www.youtube.com/watch?v=eiDyK_ofPPM

robyn/import_vars.py
```diff
@@ -0,0 +1,27 @@
...
+# check for the environment variables set in cli and if not set them
+
+def load_vars(variables = parser()):
+    """Main function"""
+    for var in variables:
+        if var[0] in os.environ:
+            continue
+    else:
+        os.environ[var[0]]=var[1]
```
sansyrox, line 22: Something similar for this function

```python3

def load_variables(variables = None):
    variables = parser() if variables is None
    ...
```

robyn/import_vars.py
```diff
@@ -0,0 +1,27 @@
...
+                continue
+            yield line.strip().split('=')  
+
+
+# check for the environment variables set in cli and if not set them
+
+def load_vars(variables = parser()):
+    """Main function"""
+    for var in variables:
+        if var[0] in os.environ:
```
sansyrox, line 19: Do you think we would want to override the existing environment variables or not?

robyn/import_vars.py
```diff
@@ -0,0 +1,27 @@
+import os
```
sansyrox, line 1: Maybe name this file something like `environment_populator` or `env_populator`? Or something similar according to you.

As a good practice is to keep the namespace or filename in accordance to what it does. And I assume the purpose of this file is to have helper functions for environment population.

definitions.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+# Path to the root of the project
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+
```
sansyrox, line 5: We can either move the contents of the file to `import_vars` only or if you like to keep the constants in a separate file, we can create a submodule.

Let me know if you need any help with that.

sansyrox (commented):
Hi @Shending-Help ,

Great work with the PR! :sparkles: 

I have added a few inline comments.

I would like to add an integration test in this PR, but we can address that once the comments have been addressed.

Great work again! :smile:

robyn/import_vars.py
Shending-Help: I thought the priority is for environment variables that are set in the terminal or manually by the user not the ones that are in the config file so i made the code check for variables that already exist and not update those

definitions.py
Shending-Help: the idea was that this file defines where the root of the project is, because if i tried to define the root in a nested file it would be dependent on the path of the file I guess

robyn/env_populator.py
```diff
@@ -0,0 +1,33 @@
...
+    """Main function"""
+    
+    variables = parser() 
+
+    if variables is None:
+        return
+
+    for var in variables:
+        if var[0] in os.environ:
+            print("Variable {} already set".format(var[0]))
```
sansyrox, line 25: We can use `logger` module here. Similar to (https://github.com/sansyrox/robyn/blob/main/robyn/__init__.py#L21) .

We will be able to control the logs using the `log-level`.

Second, instead of `format`, we can use `f-strings` as we support `Python >=3.7`. https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings

definitions.py
sansyrox: We can use Pathlib(https://docs.python.org/3/library/pathlib.html) to resolve the path. And I feel that we can move these constants and env_populator to a module of their own as the constants are only being used in the `env_populator.py`.

robyn/import_vars.py
sansyrox: Okay, makes sense. ✨

definitions.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+# Path to the root of the project
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+
+# Path to the environment variables
+CONFIG_PATH = os.path.join(ROOT_DIR, 'configuration.conf')
```
sansyrox, line 7: I convention for the configuration file that I had in my mind was something like `robyn.env`. What do you think about it?

Shending-Help: i think robyn.env is great i've changed it to that

robyn/env_populator.py
Shending-Help: should logs be in debug level or info level? and I changed the syntax such that f-strings are used instead

sansyrox: @Shending-Help , these logs should be `debug` level in my opinion.

sansyrox: Or info works too. There is a very thin line and doesn't really matter that much.

robyn/__init__.py
```diff
@@ -39,7 +40,7 @@ def __init__(self, file_object: str) -> None:
         self.headers = []
         self.directories = []
         self.event_handlers = {}
-
+        self.load_var = load_vars()
```
sansyrox, line 43: One last thing before we can proceed with the tests, `self.load_var` is `None`.  So, we can just call the `load_vars()` function.

sansyrox (commented):
All looks good. Just one final change.

sansyrox (approved):
LGTM! Great job! 🚀

PR #288, opened 2022-10-06
testing env support

Shending-Help:
**Description**

This PR tests the parser function in env_populator.py from the previous PR, and also tests if variables are set correctly

This PR fixes #287

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -75,3 +75,15 @@ def dev_session():
     time.sleep(5)
     yield
     kill_process(process)
+
+@pytest.fixture(scope="session")
+def test_session():
+    os.environ["ROBYN_URL"] = "127.0.0.1"
+    os.environ["ROBYN_PORT"] = os.environ["PORT"]
+    current_file_path = pathlib.Path(__file__).parent.resolve()
+    base_routes = os.path.join(current_file_path, "./base_routes.py")
+    command = ["python3", base_routes, "--dev"]
+    process = spawn_process(command)
+    time.sleep(5)
+    yield
+    kill_process(process)    
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -79,7 +79,7 @@ def dev_session():
 @pytest.fixture(scope="session")
 def test_session():
     os.environ["ROBYN_URL"] = "127.0.0.1"
-    os.environ["ROBYN_PORT"] = os.environ["PORT"]
+    os.environ["ROBYN_PORT"] = "5000"
     current_file_path = pathlib.Path(__file__).parent.resolve()
     base_routes = os.path.join(current_file_path, "./base_routes.py")
     command = ["python3", base_routes, "--dev"]
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -76,14 +76,3 @@ def dev_session():
     yield
     kill_process(process)
 
-@pytest.fixture(scope="session")
-def test_session():
-    os.environ["ROBYN_URL"] = "127.0.0.1"
-    os.environ["ROBYN_PORT"] = "5000"
-    current_file_path = pathlib.Path(__file__).parent.resolve()
-    base_routes = os.path.join(current_file_path, "./base_routes.py")
-    command = ["python3", base_routes, "--dev"]
-    process = spawn_process(command)
-    time.sleep(5)
-    yield
-    kill_process(process)    
--- a/integration_tests/test_env_populator.py
+++ b/integration_tests/test_env_populator.py
@@ -1,4 +1,4 @@
-from integration_tests.conftest import test_session
+# from integration_tests.conftest import test_session
 from robyn.env_populator import load_vars, parser
 import pathlib
 import os
@@ -21,7 +21,7 @@ def env_file():
     os.unsetenv("PORT")
 
 # this tests if a connection can be made to the server with the correct port imported from the env file
-def test_env_population(test_session, env_file):
+def test_env_population(dev_session, env_file):
     dir = path / "test_dir"
     env_file = dir / "robyn.env"
     load_vars(variables = parser(config_path = env_file))
--- a/integration_tests/test_env_populator.py
+++ b/integration_tests/test_env_populator.py
@@ -12,7 +12,8 @@ path = pathlib.Path(__file__).parent
 #create robyn.env before test and delete it after test
 @pytest.fixture
 def env_file():
-    CONTENT = "ROBYN_PORT=8080" + "\n" + "ROBYN_URL=127.0.1.1"
+    CONTENT = """ROBYN_PORT=8080 
+    ROBYN_URL=127.0.1.1"""
     dir = path / "test_dir"
     env_file = dir / "robyn.env"
     env_file.write_text(CONTENT)
--- a/integration_tests/test_env_populator.py
+++ b/integration_tests/test_env_populator.py
@@ -12,7 +12,7 @@ path = pathlib.Path(__file__).parent
 #create robyn.env before test and delete it after test
 @pytest.fixture
 def env_file():
-    CONTENT = "PORT=8080"
+    CONTENT = "ROBYN_PORT=8080" + "\n" + "ROBYN_URL=127.0.1.1"
     dir = path / "test_dir"
     env_file = dir / "robyn.env"
     env_file.write_text(CONTENT)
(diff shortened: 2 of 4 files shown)


integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,31 @@
...
+import pathlib
+import os
+
+# content of test_tmp_path.py
+CONTENT = "PORT=8080"
+
+path = pathlib.Path(__file__).parent  
+
+#create a directory before test and delete it after test
+def test_create_file():
```
sansyrox, line 11: You can use fixtures(https://docs.pytest.org/en/6.2.x/fixture.html) instead of depending on the execution order of tests.

You would write the test like this

```
@pytest.fixture
def env_file():
    dir = path / "test_dir"
    env_file = dir / "robyn.env"
    env_file.write_text(CONTENT)
    yield
    env_file.unlink()

def test_parser(env_file):
     ...
```

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,31 @@
...
+    d = path / "test_dir"
+    p = d / "robyn.env"
+    p.write_text(CONTENT)
+
+def test_parser():
+    d = path / "test_dir"
+    p = d / "robyn.env"
```
sansyrox, line 18: Also, can you please change the single letter variable names to words? Like in the comment above?

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,31 @@
+from robyn.env_populator import load_vars, parser
+import pathlib
+import os
+
+# content of test_tmp_path.py
+CONTENT = "PORT=8080"
```
sansyrox, line 6: and possibly move this variable in the fixture?

sansyrox (changes requested):
Hi @Shending-Help ,

The test structure looks great. I just have some suggestions regarding the implementation.

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,31 @@
...
+    d = path / "test_dir"
+    p = d / "robyn.env"
+    assert list(parser(config_path = p)) == [['PORT', '8080']]
+
+def test_load_vars():
+    d = path / "test_dir"
+    p = d / "robyn.env"
+    load_vars(variables = parser(config_path = p))
+    assert os.environ['PORT'] == '8080'
+
```
sansyrox, line 26: These tests are good but more towards the unit test side of things. An integration test would be making a request(maybe like a get request) and getting back a specific environment variable.

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,41 @@
...
+def test_parser(env_file):
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    assert list(parser(config_path = env_file)) == [['PORT', '8080']]
+
+def test_load_vars(env_file):
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    load_vars(variables = parser(config_path = env_file))
+    assert os.environ['PORT'] == '8080'
```
sansyrox, line 31: @Shending-Help , these feel more like a unit test. So, it would make sense to move them to a different folder. We can create a new unit test folder if you really want them. 

```
-unit_tests
-integration_test
-robyn
```

integration_tests/conftest.py
```diff
@@ -74,4 +74,16 @@ def dev_session():
...
     time.sleep(5)
     yield
-    kill_process(process)
\ No newline at end of file
+    kill_process(process)
+
+@pytest.fixture(scope="session")
+def test_session():
+    os.environ["ROBYN_URL"] = "127.0.0.1"
+    os.environ["ROBYN_PORT"] = os.environ["PORT"]
```
sansyrox, line 82: Try setting `ROBYN_URL` and `ROBYN_PORT` directly through `load_vars()`?

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_get_request(test_session, env_file):
```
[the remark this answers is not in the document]
Shending-Help: okay I will change it to that

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,41 @@
...
+
+#create robyn.env before test and delete it after test
+@pytest.fixture
+def env_file():
+    CONTENT = "PORT=8080"
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    env_file.write_text(CONTENT)
+    yield
+    env_file.unlink()
```
sansyrox, line 20: Just realized that unlinking the file will not be enough here. We will also have to unset the environment variables.

Shending-Help: will do

integration_tests/conftest.py
Shending-Help: but shouldn't that be left to the user, it won't always be the case that a user will provide a PORT and a URL in robyn.env  but even if he did those vars will be loaded automatically and he can just assign them to a variable and use them

Shending-Help: but that wouldn't be the case for other variables such as token or db credentials so should just auto invoke the variables that are relevant for robyn and other variables get left for the user to decide what to do with them?

sansyrox: No no. What I mean is if you set `ROBYN_PORT` in the test then it will be automatically present in the environment. That way you don't need to set that variable in the fixture.

If that still doesn't make sense then we can hop on a quick call.

Shending-Help: i think i understand what you mean i will get on it and check with you afterwards

sansyrox: Perfect!

Shending-Help: i think this should be it, I changed the content written in robyn.env to include `ROBYN_PORT` and `ROBYN_URL` and it worked, i also deleted the fixture i created in conftest.py since it is of no use now.

integration_tests/test_env_populator.py
```diff
@@ -12,7 +12,7 @@
 #create robyn.env before test and delete it after test
 @pytest.fixture
 def env_file():
-    CONTENT = "PORT=8080"
+    CONTENT = "ROBYN_PORT=8080" + "\n" + "ROBYN_URL=127.0.1.1"
```
sansyrox, line 15: You can use python multiline strings(https://www.w3schools.com/python/gloss_python_multi_line_strings.asp) here.

unit_tests/test_env_pop.py
```diff
@@ -0,0 +1,31 @@
...
+#create robyn.env before test and delete it after test
+@pytest.fixture
+def env_file():
+    CONTENT = "PORT=8080"
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    env_file.write_text(CONTENT)
+    yield
+    env_file.unlink()
+    os.unsetenv("PORT")
```
sansyrox, line 20: This should be ROBYN_PORT and ROBYN_URL now

unit_tests/test_env_pop.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_parser(env_file):
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    assert list(parser(config_path = env_file)) == [['PORT', '8080']]
```
sansyrox, line 25: Same here

unit_tests/test_env_pop.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_load_vars(env_file):
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    load_vars(variables = parser(config_path = env_file))
+    assert os.environ['PORT'] == '8080'
```
sansyrox, line 31: same here

integration_tests/test_env_populator.py
```diff
@@ -0,0 +1,32 @@
...
+#create robyn.env before test and delete it after test
+@pytest.fixture
+def env_file():
+    CONTENT = "ROBYN_PORT=8080" + "\n" + "ROBYN_URL=127.0.1.1"
+    dir = path / "test_dir"
+    env_file = dir / "robyn.env"
+    env_file.write_text(CONTENT)
+    yield
+    env_file.unlink()
+    os.unsetenv("PORT")
```
sansyrox, line 21: same here

sansyrox (commented):
All looks good. Just one final change before we merge it.

integration_tests/test_env_populator.py
Shending-Help: done

integration_tests/test_env_populator.py
Shending-Help: done

unit_tests/test_env_pop.py
Shending-Help: done

sansyrox (approved):
All looks good to me! Great work! :D

Issue #283, opened 2022-09-15
[BUG] The --dev flag not working on Windows


sansyrox:
**Description**

`--dev` flag not working on windows

```
Traceback (most recent call last): File "C:\Users\Gaurav Bhattacharjee\OneDrive\Desktop\Projects\rob\robyn\integration_tests\base_routes.py", line 209, in <module> app.start(port=ROBYN_PORT, url=ROBYN_URL) File "\\?\C:\Users\Gaurav Bhattacharjee\OneDrive\Desktop\Projects\rob\robyn\robyn\__init__.py", line 163, in start event_handler.start_server_first_time() File "\\?\C:\Users\Gaurav Bhattacharjee\OneDrive\Desktop\Projects\rob\robyn\robyn\dev_event_handler.py", line 14, in start_server_first_time self.processes.append(subprocess.Popen(["python3", self.file_name], start_new_session=False)) File "C:\Users\Gaurav Bhattacharjee\AppData\Local\Programs\Python\Python310\lib\subprocess.py", line 969, in __init__ self._execute_child(args, executable, preexec_fn, close_fds, File "C:\Users\Gaurav Bhattacharjee\AppData\Local\Programs\Python\Python310\lib\subprocess.py", line 1438, in _execute_child hp, ht, pid, tid = _winapi.CreateProcess(executable, args, FileNotFoundError: [WinError 2] The system cannot find the file specified
```

To replicate this issue: `pytest -s integration_tests` . You will see two tests failing.

**Expected Behavior**

The tests should not fail. 😄 

**Screenshots**

**Environment:**
- Host OS:  Windows
---

[Optional] **Additional Context**

---

guilefoylegaurav:
Hey @sansyrox  I am interested in taking this up :) Will look into this issue soon

sansyrox:
Thanks @guilefoylegaurav . Let me know if you require any suport/information from my end :smile:

guilefoylegaurav:
@sansyrox I am sorry, could not update you on this - stuck with mid-terms for 4 days. Shall get back to this soon.

sansyrox:
Hi @guilefoylegaurav , 
No worries. Take your time 😄 

Also, all the best with your mid-terms! ✨

guilefoylegaurav:
[...]

PR #290, opened 2022-10-07
fix: enable hot reload on windows

guilefoylegaurav:
**Description**

Concerned with issue/bug #283

### Changes
* fix(dev_event_handler.py): use platform specific python3 alias to spawn processes 

### Documentation
No documentation update is required

### How I've tested my work 
* **26/26** tests passed (Debian 10)
* **25/26** tests passed (Windows 10) (`global_index_request()` failed) 

### Remarks
* Although this minute code change gets the `dev_index_request()` test passed on windows, for some reason hot reload fails to work on my system. 

     * **Windows:** Inside `dev_event_handler.py`, function `on_any_event()` gets called every time I make a change in the `base_routes.py` file but I cannot see changes being reflected on the browser. 
     * **Debian:**  Inside `dev_event_handler.py`, function `on_any_event()` does not get called when I make a change in the `base_routes.py` file and thus I cannot see changes being reflected on the browser. 
     * I just want to make sure if this is a problem that everyone else is experiencing or is it only me?

* Failure of `global_index_request()` test persists

### DIFF
--- a/robyn/dev_event_handler.py
+++ b/robyn/dev_event_handler.py
@@ -1,4 +1,5 @@
 import subprocess
+import sys
 
 from watchdog.events import FileSystemEventHandler
 
@@ -7,11 +8,14 @@ class EventHandler(FileSystemEventHandler):
     def __init__(self, file_name) -> None:
         self.file_name = file_name
         self.processes = []
+        self.python_alias = "python3" if not sys.platform.startswith("win32") else "python"
+        self.shell = True if sys.platform.startswith("win32") else False
+        
 
     def start_server_first_time(self) -> None:
         if self.processes:
             raise Exception("Something wrong with the server")
-        self.processes.append(subprocess.Popen(["python3", self.file_name], start_new_session=False))
+        self.processes.append(subprocess.Popen([self.python_alias, self.file_name], shell = self.shell, start_new_session=False))
 
     def on_any_event(self, event) -> None:
         """
@@ -22,5 +26,5 @@ class EventHandler(FileSystemEventHandler):
 
         if len(self.processes) > 0:
             for process in self.processes:
-                process.terminate()
-        self.processes.append(subprocess.Popen(["python3", self.file_name], start_new_session=False))
+                process.terminate()         
+        self.processes.append(subprocess.Popen([self.python_alias, self.file_name], shell = self.shell, start_new_session=False))


sansyrox:
Hi @guilefoylegaurav , thanks for your PR.

I don't have an access to a windows machine till the 26th so can't really tell you about the windows issue.

Dev server reflects the right changes on my osx machine but is not killed properly on KeyboardInterrupt. I will check my linux PC to give you updates about the debian state.

robyn/dev_event_handler.py
```diff
@@ -11,7 +12,8 @@ def __init__(self, file_name) -> None:
     def start_server_first_time(self) -> None:
         if self.processes:
             raise Exception("Something wrong with the server")
-        self.processes.append(subprocess.Popen(["python3", self.file_name], start_new_session=False))
+        command = "python3" if not sys.platform.startswith("win32") else "python"
```
sansyrox, line 15: instead of this we can create a private computed python property that gives us a python alias depending on the platform. What do you think?

guilefoylegaurav: Correct. That would be much better.

sansyrox:
@guilefoylegaurav , I think you pushed the same commit.

guilefoylegaurav:
Noticing that there were some updates to the upstream, I pulled the updates down from upstream to the fork's main and rebased by feature branch with the main and push that back, actually

sansyrox:
@guilefoylegaurav , are you working on converting duplicate code to a `property`?

guilefoylegaurav:
My apologies
I am, just pushed the updates

sansyrox:
Just one last thing @guilefoylegaurav , have you run the `pre-commit` instructions in the README here?

https://github.com/sansyrox/robyn#%EF%B8%8F-to-develop-locally

guilefoylegaurav:
@sansyrox 
I did not run it until now 
Ran it now and it did not reveal any formatting issues

I was under the impression that the hooks would run automatically on commit: https://[pre-commit](https://pre-commit.com/).com/

sansyrox:
Ah, okay. This PR looks good. But I will do a final review tomorrow morning 😄 

Also, the link that you shared is dead in the last comment.

guilefoylegaurav:
Thanks @sansyrox 
My bad, I updated the link
I am embarrassed beyond words - I messed the link up while copying it seems

robyn/dev_event_handler.py
```diff
@@ -7,11 +8,14 @@ class EventHandler(FileSystemEventHandler):
     def __init__(self, file_name) -> None:
         self.file_name = file_name
         self.processes = []
+        self.alias = "python3" if not sys.platform.startswith("win32") else "python"
```
sansyrox, line 11: One last nit: can you name it as `python_alias`?

sansyrox:
> I am embarrassed beyond words - I messed the link up while copying it seems. I will take care from now on to ensure that this does not happen again

Not to worry :smile: Happens with all of us 😄

robyn/dev_event_handler.py
guilefoylegaurav: Done

sansyrox (approved):
Great work! LGTM :fire:

PR #292, opened 2022-10-13
 Hotfix detecting robyn.env

Shending-Help:
**Description**

This PR fixes an issue with detecting robyn.env in a project, if the server is started somewhere in a nested file or the user placed robyn.env somewhere different than expected of him robyn.env won't be found and thus i added a file detector function that starts searching from an entry point where robyn server started and goes from the bottom up in the directories searching for robyn.env

### DIFF
--- a/robyn/env_populator.py
+++ b/robyn/env_populator.py
@@ -3,13 +3,6 @@ import logging
 from pathlib import Path
 
 
-
-# # Path to the root of the project
-# ROOT_DIR = Path(__file__)
-
-# # Path to the environment variables
-# CONFIG_PATH = ROOT_DIR / 'robyn.env'
-
 #set the logger that will log the environment variables imported from robyn.env and the ones already set
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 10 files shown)


sansyrox:
Hey @Shending-Help , Thanks for the PR. :smile: 

I have a suggestion regarding the implementation approach.

We should only be detecting `robyn.env` at the top level of the project. So, we need to search only in the parent directory of the file that is being executed instead of doing a recursive `os.walk` in the code.

Shending-Help:
@sansyrox  I made it so the root of the project gets passed to the load_vars() function and pass it the directory at which Robyn gets initialized

robyn/env_populator.py
```diff
@@ -1,21 +1,20 @@
...
 
+
 # parse the configuration file returning a list of tuples (key, value) containing the environment variables
-def parser(config_path=CONFIG_PATH):
+def parser(config_path=None, project_root = None):
+    """Find robyn.env file in root of the project"""
+    if config_path is None:
+        config_path = Path(project_root) / "robyn.env"
+
     """Parse the configuration file"""
```
sansyrox, line 18: @Shending-Help , this should be the docstring of the function as this is the main purpose.

You can make the comment below  as a `#`comment.
> """Find robyn.env file in root of the project"""

Docstrings are reflected in the function metadata and it doesn't make a lot of sense to add two of them.

Shending-Help: @sansyrox  fixed it

robyn/env_populator.py
```diff
@@ -26,11 +24,11 @@ def parser(config_path=CONFIG_PATH):
 
 
 # check for the environment variables set in cli and if not set them
-def load_vars(variables = None):
+def load_vars(variables = None, root = None):
```
sansyrox, line 27: Two last nits: let's call the variable as `project_root` only as we want the `**kwargs` to be related to the function name,

Second, can you please use the `pre-commit` hooks like mentioned in here https://github.com/sansyrox/robyn#%EF%B8%8F-to-develop-locally

Thanks again :smile:

sansyrox (approved):
LGTM! Great work @Shending-Help 🔥

PR #299, opened 2022-10-28
add route type enum

### LINKED ISSUE #254: [Feature Request] Refactor Code to include better types
**Current Behavior**

Right now, there are a lot of keys as `Strings` in hashmaps. And a lot of strings are being passed where enums would be a better choice.
I lot of problems are being faced here. https://github.com/sansyrox/robyn/pull/253

**Desired Behavior**

Refactor the codebase.

---
**Screenshots / Mockups**

**Alternatives**

---

suhailmalik07:
**Description**

This PR fixes #254 

I'm working on this. 

@sansyrox It will be great, If you can point me to the places from where you think code can be converted to enum and types?

### DIFF
--- a/src/routers/middleware_router.rs
+++ b/src/routers/middleware_router.rs
@@ -9,7 +9,7 @@ use matchit::Router;
 
 use anyhow::{bail, Error, Result};
 
-use super::types::MiddlewareRouteType;
+use crate::routers::types::MiddlewareRoute;
 
 /// Contains the thread safe hashmaps of different routes
 
@@ -27,10 +27,10 @@ impl MiddlewareRouter {
     }
 
     #[inline]
-    fn get_relevant_map(&self, route: MiddlewareRouteType) -> Option<&RwLock<Router<(PyFunction, u8)>>> {
+    fn get_relevant_map(&self, route: MiddlewareRoute) -> Option<&RwLock<Router<(PyFunction, u8)>>> {
         match route {
-            MiddlewareRouteType::BeforeRequest => Some(&self.before_request),
-            MiddlewareRouteType::AfterRequest => Some(&self.after_request),
+            MiddlewareRoute::BeforeRequest => Some(&self.before_request),
+            MiddlewareRoute::AfterRequest => Some(&self.after_request),
         }
     }
 
@@ -38,7 +38,7 @@ impl MiddlewareRouter {
     // Inserts them in the router according to their nature(CoRoutine/SyncFunction)
     pub fn add_route(
         &self,
-        route_type: MiddlewareRouteType, // we can just have route type as WS
+        route_type: MiddlewareRoute,
         route: &str,
         handler: Py<PyAny>,
         is_async: bool,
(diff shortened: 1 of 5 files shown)


src/routers/middleware_router.rs
```diff
@@ -9,6 +9,8 @@ use matchit::Node;
 
 use anyhow::{bail, Error, Result};
 
+use super::router::RouteType;
```
sansyrox, line 12: I don't think `super` is required anymore in Rust.

src/routers/router.rs
```diff
@@ -10,6 +10,22 @@ use matchit::Node;
...
+impl RouteType {
+    pub fn from_str(input: &str) -> RouteType {
+        match input {
+            "BEFORE_REQUEST" => RouteType::BeforeRequest,
+            "AFTER_REQUEST" => RouteType::AfterRequest,
+            _ => panic!("Invalid route type enum."),
+        }
+    }
+}
+
```
sansyrox, line 28: These routes are just for the middleware router and not the global router. So, I think something like `MiddlewareRouteType` would be more relevant.

Also, we usually create a `types` folder in the namespace. e.g. `routes/types.py`

sansyrox (commented):
Hey @suhailmalik07 . 👋 

Thank you for the PR. 😄 

I have a few inline comments.

src/routers/middleware_router.rs
suhailmalik07: another option is to write like this.
```
use crate::routers::types::MiddlewareRouteType;
```

src/routers/router.rs
suhailmalik07: updated PR with given details. Thanks for pointing it out.

src/routers/middleware_router.rs
sansyrox: @suhailmalik07 , yep. That is the preferred way.

src/routers/middleware_router.rs
```diff
@@ -25,19 +27,21 @@ impl MiddlewareRouter {
     }
 
     #[inline]
-    fn get_relevant_map(&self, route: &str) -> Option<&RwLock<Node<(PyFunction, u8)>>> {
+    fn get_relevant_map(
+        &self,
+        route: MiddlewareRouteType,
```
sansyrox, line 32: just one last nit: `MiddlewareRoute` would make more sense than `MiddlewareRouteType`.

Reason:

If you ask the type of route then it should be `MiddlewareRoute` and not type = `MiddlewareRouteType.`

sansyrox (commented):
Thanks again @suhailmalik07 for the quick updates. I just have two final suggestions before we can merge this PR 😄

src/routers/middleware_router.rs
suhailmalik07: update with this.

src/routers/middleware_router.rs
suhailmalik07: Sure, updated this also.

sansyrox (approved):
LGTM! Great work!

sansyrox:
@suhailmalik07 , there are some conflicting files. Can you please rebase the latest changes?

suhailmalik07:
> @suhailmalik07 , there are some conflicting files. Can you please rebase the latest changes?

Rebased 🙂

sansyrox (approved):
LGTM! 🚀 🟢

PR #301, opened 2022-11-01
feat: Replace PyFunction with a decorator

Jamyw7g:
Signed-off-by: wuyaoping <jamykp@gmail.com>

**Description**

*This is a DRAFT to talk about the new feature.*

We can see, there are many code like `fn add_route(..., handler: Py<PyAny>, is_async: bool, ...)` in project. As a result, it have to add a lot of code to handle different inputs.

I have a new idea, how about replace it with decorator *(SEE BELOW)*. After that, all the `Handler` is a coroutine, which can handle as a `Future` in Rust code.

@sansyrox It will be a huge project, so I want to discuss it first

```Python
from asyncio.coroutines import iscoroutine as__iscoroutine
from functools import wraps as __wraps
from typing import Callable


def handler(fn: Callable):
    """Wrap any callable to awaitable.
    """
    @__wraps(fn)
    async def async_inner(*args, **kwargs):
        ret = fn(*args, **kwargs)
        if __iscoroutine(ret):
            return await ret
        else:
            return ret
    return async_inner
```

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,6 +1,7 @@
 use dashmap::DashMap;
 use pyo3::prelude::*;
 
+// Replace PyFunction with a decorator
 #[derive(Debug, Clone)]
 pub enum PyFunction {
     CoRoutine(Py<PyAny>),


sansyrox:
@Jamyw7g , the reason we allow sync functions to be executed without being converted to async functions is that synchronous function execution is a little bit faster. 

Wrapping it in an async function will add some overhead and will deem our sync pipeline useless. We can benchmark the changes to see if this will cause any significant delays.

Jamyw7g:
@sansyrox OK! I need some time to do tests.

sansyrox:
@Jamyw7g , no worries 😄 Lmk if you need any help.

Jamyw7g:
@sansyrox It is not a good practice yet. thx!

sansyrox:
@Jamyw7g , no worries. Thank you for making an attempt though 😄

PR #307, opened 2022-11-11
Router refactor

AntoineRR:
**Description**

This PR refactors the `routers` module of the Rust part of Robyn.
I noticed some improvements could be made there so here is my take on it.

Noticeable changes:
- Rename `Router` to `DynRouter`
- Introduce a `Router` trait to unify the behavior of `ConstRouter`, `DynRouter`, and `MiddlewareRouter`
- Get rid of the `get_relevant_map` methods (mapping is directly done at structure creation)
- Some parts of the code were made less verbose

I tried to keep the same logic as before. I think this could pave the way for further performance improvements for the router  :slightly_smiling_face:

### DIFF
--- a/src/routers/const_router.rs
+++ b/src/routers/const_router.rs
@@ -13,7 +13,6 @@ use actix_web::http::Method;
 
 use anyhow::{Error, Result};
 
-use super::RouteType;
 use super::Router;
 
 type RouteMap = RwLock<matchit::Router<String>>;
@@ -59,13 +58,8 @@ impl Router<String, Method> for ConstRouter {
         Ok(())
     }
 
-    fn get_route(
-        &self,
-        route_method: RouteType<Method>,
-        route: &str, // check for the route method here
-    ) -> Option<String> {
-        // need to split this function in multiple smaller functions
-        let table = self.routes.get(&route_method.0)?;
+    fn get_route(&self, route_method: Method, route: &str) -> Option<String> {
+        let table = self.routes.get(&route_method)?;
         let route_map = table.read().ok()?;
 
         match route_map.at(route) {
--- a/src/routers/const_router.rs
+++ b/src/routers/const_router.rs
@@ -1,125 +1,109 @@
+use std::collections::HashMap;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::sync::RwLock;
 // pyo3 modules
 use crate::executors::execute_function;
+use anyhow::Context;
 use log::debug;
 use pyo3::prelude::*;
 use pyo3::types::PyAny;
 
 use actix_web::http::Method;
-use matchit::Router;
 
-use anyhow::{bail, Error, Result};
+use anyhow::{Error, Result};
+
+use super::RouteType;
+use super::Router;
+
+type RouteMap = RwLock<matchit::Router<String>>;
 
 /// Contains the thread safe hashmaps of different routes
-
 pub struct ConstRouter {
-    get_routes: Arc<RwLock<Router<String>>>,
-    post_routes: Arc<RwLock<Router<String>>>,
-    put_routes: Arc<RwLock<Router<String>>>,
-    delete_routes: Arc<RwLock<Router<String>>>,
-    patch_routes: Arc<RwLock<Router<String>>>,
-    head_routes: Arc<RwLock<Router<String>>>,
-    options_routes: Arc<RwLock<Router<String>>>,
-    connect_routes: Arc<RwLock<Router<String>>>,
-    trace_routes: Arc<RwLock<Router<String>>>,
+    routes: HashMap<Method, Arc<RouteMap>>,
 }
 
-impl ConstRouter {
-    pub fn new() -> Self {
-        Self {
-            get_routes: Arc::new(RwLock::new(Router::new())),
-            post_routes: Arc::new(RwLock::new(Router::new())),
-            put_routes: Arc::new(RwLock::new(Router::new())),
-            delete_routes: Arc::new(RwLock::new(Router::new())),
-            patch_routes: Arc::new(RwLock::new(Router::new())),
-            head_routes: Arc::new(RwLock::new(Router::new())),
-            options_routes: Arc::new(RwLock::new(Router::new())),
-            connect_routes: Arc::new(RwLock::new(Router::new())),
-            trace_routes: Arc::new(RwLock::new(Router::new())),
-        }
-    }
-
-    #[inline]
-    fn get_relevant_map(&self, route: Method) -> Option<Arc<RwLock<Router<String>>>> {
-        match route {
-            Method::GET => Some(self.get_routes.clone()),
-            Method::POST => Some(self.post_routes.clone()),
-            Method::PUT => Some(self.put_routes.clone()),
-            Method::PATCH => Some(self.patch_routes.clone()),
-            Method::DELETE => Some(self.delete_routes.clone()),
-            Method::HEAD => Some(self.head_routes.clone()),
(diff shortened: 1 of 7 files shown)


sansyrox:
@AntoineRR , great work. I will try to review this PR by today evening 😄

src/routers/const_router.rs
```diff
@@ -1,120 +1,71 @@
...
-
-            self.get_relevant_map(method)
-        } else {
-            None
-        }
-    }
-
-    /// Checks if the functions is an async function
-    /// Inserts them in the router according to their nature(CoRoutine/SyncFunction)
+impl Router<String, Method> for ConstRouter {
```
sansyrox, line 26: Hey @AntoineRR ,

I am bit unfamiliar with `traits` in rust. Can you please explain this statement to me. And the benefits associate with it?

src/routers/mod.rs
```diff
@@ -1,5 +1,28 @@
+use anyhow::Result;
+use pyo3::{Py, PyAny};
+
 pub mod const_router;
 pub mod middleware_router;
 pub mod router;
 pub mod types;
 pub mod web_socket_router;
+
+pub struct RouteType<T>(pub T);
```
sansyrox, line 10: what does the `(pub T)` mean here??

src/routers/mod.rs
```diff
@@ -1,5 +1,28 @@
...
+        route_type: &str,
+        route: &str,
+        handler: Py<PyAny>,
+        is_async: bool,
+        number_of_params: u8,
+        event_loop: Option<&PyAny>,
+    ) -> Result<()>;
+
+    /// Checks if the functions is an async function
+    /// Inserts them in the router according to their nature(CoRoutine/SyncFunction)
```
sansyrox, line 26: One nit: `get` function won't insert the function in the router.

src/routers/router.rs
```diff
@@ -8,89 +8,34 @@ use pyo3::types::PyAny;
...
-            };
+type RouteMap = RwLock<MatchItRouter<(PyFunction, u8)>>;
 
-            self.get_relevant_map(method)
-        } else {
-            None
-        }
-    }
+/// Contains the thread safe hashmaps of different routes
+pub struct DynRouter {
```
sansyrox, line 18: Is `DynRouter` a `HTTPRouter`?

src/server.rs
```diff
@@ -394,24 +389,25 @@ async fn index(
...
-                    route_params,
-                    queries.clone(),
-                )
-                .await;
-                debug!("Middleware contents {:?}", x);
-                x
-            }
-            None => HashMap::new(),
-        };
+    let tuple_params = match middleware_router
```
sansyrox, line 392: Do you have a better name suggestion for `tuple_params`?

sansyrox (commented):
Thanks for the PR @AntoineRR :D . I have a few inline comments.

src/routers/const_router.rs
AntoineRR: Sure !

## About traits

Traits are a way in Rust to share behavior between structs. There is no inheritance in Rust as in a truly object oriented language like C++, so traits are more similar to interfaces in Java for example.

Here, defining the `Router` trait allows to specify what methods should be in a router : `add_route` and `get_route`. For every router we want to define, we have to implement this trait using the line above. This will force us to provide a definition for the `add_route` and `get_route` method that is specific to our struct, which is `ConstRouter` here.

## About the generics

I didn't want to change the code too much so I kept the same return types and parameter types in every router struct. The return type of the `get_route` method is different across the routers (`DynRouter` and `MiddlewareRouter` return an `Option<((PyFunction, u8), HashMap<String, String>)>` while `ConstRouter` return an `Option<String>`), so I used a generic type to unify their behavior in the trait. Similarly, the `route_type` parameter of the `get_route` method can either be a `MiddlewareRoute` or a `Method`. I needed two different generic types to handle those differences: this is `T` and `U` in the definition of the trait : `pub trait Router<T, U>`. For `ConstRouter`, `T` is `String` and `U` is `Method`.

## Benefits of using traits

With this PR, `DynRouter`, `ConstRouter` and `MiddlewareRouter` all implement the `Router` trait (I wanted to include the `WebSocketRouter` too but it seems too different from the others). This means the compiler is aware those share some behavior. We could now do some cool things like:
- Creating a `Vec<dyn Router>` containing the routers
- Create generic methods that take a type implementing the `Router` trait as parameter -> `fn method<T: Router>(router: T)`
- Eventually provide a default implementation for the `add_route` and `get_route` methods if we can make it the same for all structs (which I may try to do later)
- If a new method needs to be added for every router, its implementation can be in the `Router` trait instead of in each struct
- Maybe other things I forgot?

I hope this makes things clearer! In this PR I didn't take advantage of the trait for now, I wanted to do small iterations on the code instead of a really big one. I will try to improve things further if I have some time :wink:

src/routers/mod.rs
AntoineRR: I used this struct as a wrapper for the `Method` and `MiddlewareRoute` types. RouteType is a struct that can contain any other type. `pub T` is an unnamed field of this struct, which is public and of type `T`. It can be accessed like this in an instance of type `RouteType`: `route_type.0`.

However now that I think about it it is probably unnecessary here, and I could just use a generic type instead of wrapping the parameter in this struct. I will push a new commit removing it soon

src/routers/mod.rs
AntoineRR: Oops I copied the wrong comment, will fix that

src/routers/router.rs
AntoineRR: What do you mean by that?
This is the struct that was previously called `Router`. To my understanding it stores the routes that have parameters, and I called it `DynRouter` to oppose it to `ConstRouter`. I'm not familiar with routers so I may have not understood everything, let me know if you prefer me to rename it

src/server.rs
AntoineRR: I'm not sure about it, but as I think the purpose of the middleware is to modify the request before it is routed, I called it `modified_request`?

src/routers/router.rs
sansyrox: @AntoineRR , the vanilla `Router`(now `DynRouter`) just serves HTTP Requests. I am not sure what you mean by `DynRouter`. I was thinking of something like `HTTPRouter` or `HTTPRequestRouter`.

What do you think?

AntoineRR: It makes sense! I renamed it to `HttpRouter` and changed the file name as well.

src/routers/mod.rs
sansyrox: Ah, that makes sense. Thank you for explaining 😄

src/routers/const_router.rs
sansyrox: Thank you @AntoineRR . This an excellent explanation 😄

src/server.rs
sansyrox: `modified_request` is much better than `tuple_params` Thanks :D

src/routers/const_router.rs
AntoineRR: You're welcome ! 😉

sansyrox (approved):
LGTM! 🚀  Great work. 

But I am holding this PR to allow (https://github.com/sansyrox/robyn/pull/297) to merge.

AntoineRR:
Ok thanks! I'll solve the eventual merge conflicts once the other PR is merged

sansyrox:
Looks like the other PR is gonna take some time. 

Great job on this one!

PR #318, opened 2022-11-21
Various improvements around the index method

AntoineRR:
**Description**

Hello, I'm back with a new suggestion for the Rust code of Robyn.
I tried to simplify the code around the `index` function in `server.rs`:
- Removed unnecessary `Rc<RefCell<>>`
- Removed some `clone()` calls (this is mostly possible due to the use of the `to_object` method, which takes a reference to the object, instead of the `into_py` method which takes ownership of the object)
- Simplified some chunks of code

Let me know if this is ok for you :slightly_smiling_face:

### DIFF
--- a/src/executors/mod.rs
+++ b/src/executors/mod.rs
@@ -2,9 +2,7 @@
 /// i.e. the functions that have the responsibility of parsing and executing functions.
 use crate::io_helpers::read_file;
 
-use std::cell::RefCell;
 use std::collections::HashMap;
-use std::rc::Rc;
 use std::sync::Arc;
 
 use actix_web::{http::Method, web, HttpRequest};
@@ -25,8 +23,8 @@ pub async fn execute_middleware_function<'a>(
     payload: &mut web::Payload,
     headers: &HashMap<String, String>,
     req: &HttpRequest,
-    route_params: HashMap<String, String>,
-    queries: Rc<RefCell<HashMap<String, String>>>,
+    route_params: &HashMap<String, String>,
+    queries: &HashMap<String, String>,
     number_of_params: u8,
 ) -> Result<HashMap<String, HashMap<String, String>>> {
     // TODO:
@@ -55,20 +53,14 @@ pub async fn execute_middleware_function<'a>(
     // request object accessible while creating routes
     let mut request = HashMap::new();
 
-    let mut queries_clone: HashMap<String, String> = HashMap::new();
-
-    for (key, value) in (*queries).borrow().clone() {
-        queries_clone.insert(key, value);
-    }
-
     match function {
         PyFunction::CoRoutine(handler) => {
             let output = Python::with_gil(|py| {
                 let handler = handler.as_ref(py);
-                request.insert("params", route_params.into_py(py));
-                request.insert("queries", queries_clone.into_py(py));
+                request.insert("params", route_params.to_object(py));
+                request.insert("queries", queries.to_object(py));
                 // is this a bottleneck again?
-                request.insert("headers", headers.clone().into_py(py));
+                request.insert("headers", headers.to_object(py));
                 // request.insert("body", data.into_py(py));
 
                 // this makes the request object to be accessible across every route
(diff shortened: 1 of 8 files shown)


sansyrox:
Thank you for the PR @AntoineRR 😄 

Looks great at the first glance. I will give it a thorough review by tomorrow evening. 😄

src/executors/mod.rs
```diff
@@ -55,20 +53,14 @@ pub async fn execute_middleware_function<'a>(
...
-    }
-
     match function {
         PyFunction::CoRoutine(handler) => {
             let output = Python::with_gil(|py| {
                 let handler = handler.as_ref(py);
-                request.insert("params", route_params.into_py(py));
-                request.insert("queries", queries_clone.into_py(py));
+                request.insert("params", route_params.to_object(py));
+                request.insert("queries", queries.to_object(py));
```
sansyrox, line 61: @AntoineRR , I have one question - What benefit does `to_object` bring over `into_py`?

AntoineRR: From the docs of pyo3 [here](https://pyo3.rs/v0.17.3/conversions/traits.html):
[ToPyObject](https://docs.rs/pyo3/0.17.3/pyo3/conversion/trait.ToPyObject.html) is a conversion trait that allows various objects to be converted into [PyObject](https://docs.rs/pyo3/0.17.3/pyo3/type.PyObject.html). IntoPy<PyObject> serves the same purpose, except that it consumes self.

I used `to_object` because it doesn't consume self. This means I was able to use references instead of taking ownership of the object, and I didn't need to clone the object to use it twice in the code.

sansyrox: Ah, alrighty. Thank you! :D


I have a feeling that this will result in some performance increase!

AntoineRR: It may lead to performance increase, but I am not sure about it because the Rust compiler is smart enough to remove unnecessary `clone` calls most of the time ^^
Btw how do you test the performances of Robyn ? For testing the Rust code it is possible to write benchmarks using the [criterion](https://crates.io/crates/criterion) crate, I could try to do it if you are interested :)

sansyrox: @AntoineRR , I just compile it and run it through `oha`. (https://github.com/hatoo/oha) 

I had a look at `criterion` before but was unable to figure out how would I bench the requests. What approach would you take for this?

AntoineRR: Ok, I used oha to run some performance tests too, and saw no difference on my machine sadly :( But I think it still makes the code more readable ^^
To benchmark with criterion, we would have to first add routes to the routers, then generate a few requests we should test, and I guess we could benchmark the index method synchronously by sending all the requests to it? If there is enough variations in the requests it would be representative of the average speed of the Rust core of Robyn.

sansyrox: @AntoineRR , that sounds complicated. What do you think about python/bash script that runs a dummy app and runs oha on it.

Your approach will give us more granular results, but then I have a question how much of it can we actually control? As we use `pyo3` and then the next steps would be writing unsafe CFFI ourselves.

Let me know what you think of it. And feel free to correct me if I am wrong or very wrong 😆

AntoineRR: Adding a python or bash script to run performance tests on Robyn as a whole would be a good idea I think, it would let us easily check if a change in the code has an impact on performances!

I don't think it would require that much work to write a benchmark with criterion, but my pyo3 knowledge is quite limited for now. I may try it and submit a PR if I find it useful enough tho ^^

For now a script running oha is surely good enough, and easier to setup! I can open an issue for this if you are interested

sansyrox: > For now a script running oha is surely good enough, and easier to setup! I can open an issue for this if you are interested

Yes, please. Thank you! :D

> I don't think it would require that much work to write a benchmark with criterion, but my pyo3 knowledge is quite limited for now. I may try it and submit a PR if I find it useful enough tho ^^

Perfect. Do lmk if I can be of any help here. :D

sansyrox (approved):
LGTM! Great work! :D

src/executors/mod.rs
AntoineRR: Thanks, I'll let you know if I got something to work with criterion ;)
Here is the issue for a follow up on the oha benchmark: https://github.com/sansyrox/robyn/issues/320

PR #322, opened 2022-11-27
fix: factorizing code

AntoineRR:
**Description**

This PR is a big refactor of Robyn's core. It aims at simplifying the code by factorizing it and introducing new useful structs. Here is a description of what I've done to achieve this:
- Create a `Request` struct that wraps the parameters of a request (`queries`, `headers`, `method`, `params`, and `body`)
- Use the `Request` struct in request handlers and executors instead of using each parameters of a request separately
- Create a `FunctionInfo` struct to describe functions coming from Python. This struct is made available in the Python code too and wraps `handler`, `is_async` and `number_of_params`
- The `FunctionInfo` struct replace `handler`, `is_async` and `number_of_params` in every function call
- The `execute_function` function that was used by the const router has been removed as it was very similar to `execute_http_function`. Const router now uses `execute_http_function` (not sure about this one, let me know if I should restore `execute_function` which used `tokio::task::spawn_blocking` while `execute_http_function` doesn't)

This a big PR and it affects a lot of modules. It may need some time to be fully reviewed, do not hesitate to ask me questions about it and ask me to restore some parts if I made mistakes :)

### DIFF
--- a/robyn/processpool.py
+++ b/robyn/processpool.py
@@ -3,7 +3,7 @@ import sys
 from typing import Dict, Tuple
 
 from robyn.events import Events
-from robyn.robyn import Server, SocketHeld
+from robyn.robyn import FunctionInfo, Server, SocketHeld
 from robyn.router import MiddlewareRoute, Route
 from robyn.ws import WS
 
@@ -34,7 +34,7 @@ def spawn_process(
     routes: Tuple[Route, ...],
     middlewares: Tuple[MiddlewareRoute, ...],
     web_sockets: Dict[str, WS],
-    event_handlers: Dict[Events, list],
+    event_handlers: Dict[Events, FunctionInfo],
     socket: SocketHeld,
     workers: int,
 ):
@@ -67,26 +67,18 @@ def spawn_process(
         server.add_header(key, val)
 
     for route in routes:
-        route_type, endpoint, handler, is_async, number_of_params, const = route
-        server.add_route(
-            route_type, endpoint, handler, is_async, number_of_params, const
-        )
+        route_type, endpoint, function, const = route
+        server.add_route(route_type, endpoint, function, const)
 
     for route in middlewares:
-        route_type, endpoint, handler, is_async, number_of_params = route
-        server.add_middleware_route(
-            route_type, endpoint, handler, is_async, number_of_params
-        )
+        route_type, endpoint, function = route
+        server.add_middleware_route(route_type, endpoint, function)
 
     if "startup" in event_handlers:
-        server.add_startup_handler(
-            event_handlers[Events.STARTUP][0], event_handlers[Events.STARTUP][1]
-        )
+        server.add_startup_handler(event_handlers[Events.STARTUP])
 
     if "shutdown" in event_handlers:
-        server.add_shutdown_handler(
-            event_handlers[Events.SHUTDOWN][0], event_handlers[Events.SHUTDOWN][1]
-        )
+        server.add_shutdown_handler(event_handlers[Events.SHUTDOWN])
 
     for endpoint in web_sockets:
         web_socket = web_sockets[endpoint]
--- a/robyn/processpool.py
+++ b/robyn/processpool.py
@@ -67,8 +67,8 @@ def spawn_process(
         server.add_header(key, val)
 
     for route in routes:
-        route_type, endpoint, function, const = route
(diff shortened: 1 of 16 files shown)


AntoineRR:
No worries, take as much time as you need ^^

robyn/robyn.pyi
```diff
@@ -1,13 +1,17 @@
...
-from typing import Callable, Optional, Tuple
+from typing import Callable, Optional
 
 class SocketHeld:
     def __init__(self, url: str, port: int):
         pass
     def try_clone(self) -> SocketHeld:
         pass
 
+class FunctionInfo:
```
sansyrox, line 11: @AntoineRR , I think we will benefit more from making it a `dataclass`https://www.pythontutorial.net/python-oop/python-dataclass/

AntoineRR: Good idea yes! I'm adding it

robyn/processpool.py
```diff
@@ -67,26 +67,18 @@ def spawn_process(
         server.add_header(key, val)
 
     for route in routes:
-        route_type, endpoint, handler, is_async, number_of_params, const = route
-        server.add_route(
-            route_type, endpoint, handler, is_async, number_of_params, const
-        )
+        route_type, endpoint, function, const = route
```
sansyrox, line 70: If we are refactoring this code, can we please rename the `const` variable to `is_const` or `is_const_function`?

robyn/processpool.py
```diff
@@ -67,26 +67,18 @@ def spawn_process(
...
+        server.add_route(route_type, endpoint, function, const)
 
     for route in middlewares:
-        route_type, endpoint, handler, is_async, number_of_params = route
-        server.add_middleware_route(
-            route_type, endpoint, handler, is_async, number_of_params
-        )
+        route_type, endpoint, function = route
+        server.add_middleware_route(route_type, endpoint, function)
```
sansyrox, line 75: Also, I love how the routes and the functions are kept segregated ✨

src/types.rs
```diff
@@ -1,10 +1,49 @@
...
+}
+
+#[derive(Default)]
+pub struct Request {
+    pub queries: HashMap<String, String>,
+    pub headers: HashMap<String, String>,
+    pub method: Method,
+    pub params: HashMap<String, String>,
+    pub body: Bytes,
+}
```
sansyrox, line 36: @AntoineRR what `method` does it give to the request object on `Request:default()`?

src/routers/const_router.rs
```diff
@@ -44,7 +42,7 @@ impl Router<String, Method> for ConstRouter {
             event_loop.context("Event loop must be provided to add a route to the const router")?;
 
         pyo3_asyncio::tokio::run_until_complete(event_loop, async move {
-            let output = execute_function(function, number_of_params, is_async)
+            let output = execute_http_function(&Request::default(), function)
```
sansyrox, line 45: Good catch! Thanks!

sansyrox:
All looks good from the first glance. I have one last question - is `execute_function` used anywhere now?

I will just try a final run and test before approving 😄

src/types.rs
AntoineRR: From actix_web:
```
impl Default for Method {
    #[inline]
    fn default() -> Method {
        Method::GET
    }
}
```
The default method is a GET then. It shouldn't matter for the const router, I just wanted a request to use the `execute_http_function` method, but it shouldn't be used in the case of the const router

src/routers/const_router.rs
AntoineRR: You're welcome ;)
I think it makes sense to only have one method to execute the function, as we will only have to change this one for an impact on `HttpRouter` and `ConstRouter` at the same time!

robyn/processpool.py
AntoineRR: Makes sense, I changed it to `is_const` where it is relevant.

AntoineRR:
I'm glad you like it!
To answer your question, **no**, `execute_function` was only used in the const router. As I replaced this call with a call to `execute_http_function`, I removed the function completely.
This PR removes many lines of code so another test is certainly not too much ^^ I'll be here for any other adjustements, let me know if you need me :)

sansyrox (approved):
LGTM! 👍 🚀 

Great work!

Issue #323, opened 2022-11-28
[BUG] Custom headers not sent when const=False


cirospaciari:
**Description**

Custom header is only sent when const=True
**Expected Behavior**

Custom header must always be sent even without const=True

**Environment:**
- Host OS: Debian GNU/Linux bookworm/sid x86_64 Kernel 6.0.0-2-amd64
- Browser:  Chrome 107.0.5304.110

---

[Optional] **Additional Context**

When using const=False no Custom headers are sent.
```python
app.add_header("Server", "Robyn")
app.add_header("Content-Type", "text/plain")
```

```python
@app.get('/plaintext')
def plaintext() -> str:
    return "Hello, world!"
```
When using const=True Date headers are sent.
```python
@app.get('/plaintext', const=True)
def plaintext() -> str:
    return "Hello, world!"
    
```
I just find this, when trying to run TechEmPower Benchmarks without const

---

sansyrox:
Thank you @cirospaciari for opening this issue 😄 

Can you please elaborate on what you mean by the Date headers?

cirospaciari:
Sorry actually i miss understood the issue, whats is happens is:
After adding this custom headers
```python
app.add_header("Server", "Robyn")
app.add_header("Content-Type", "text/plain")
```
Withou const=True the headers are not sent.
Date header is fine

sansyrox:
@cirospaciari , this shouldn't happen. Will have to investigate. Thanks!

PR #331, opened 2022-12-08
fix: allow response headers  and fix headers not working in const requests

### LINKED ISSUE #325: [Feature Request] Allow the ability of sending the headers from the same route
**Current Behavior**

Currently, you cannot send the custom headers in the response from you route. Add the ability to do that.

**Desired Behavior**

Something along these lines, from our discussion on discord with @cirospaciari

```
@app.get('/plaintext')
def plaintext() -> str:
    return { "body": "Hello, world!", "headers": { "Content-Type": "plain/text", "Custom": "My Custom Value" }}  
```

---
**Screenshots / Mockups**

**Alternatives**

---

sansyrox:
**Description**
fix: allow response headers  and fix headers not working in const requests
This PR fixes #325 , #323


sansyrox:
@AntoineRR @cirospaciari , do you folks want to give it a final green light before we release a new version?

I really appreciate all your help 😄

cirospaciari:
> @AntoineRR @cirospaciari , do you folks want to give it a final green light before we release a new version?
> 
> I really appreciate all your help smile

I think will work.
But my issue is with ```jsonify``` basically this is serializing and deserializing the headers without any need, in CPython this is really slow, even if Robyn can get all raw power of actix web (3.6 mi req/s in TechEmPower) will be limited to JSON serializer performance, the better JSON serializer performs about 1.5 mi req/s in TechEmPower, also is adding a lot of CPU usage, copying, and dynamic allocations. Rewriting now will be better than adding this to main and cause incompatibility issues in the future.

AntoineRR:
I agree with @cirospaciari, we shouldn't use `jsonify` for the headers.
Although I think it is already handled like this in Robyn currently (see [this](https://github.com/sansyrox/robyn/blob/main/integration_tests/base_routes.py#L206) test route from the main branch for example) so there will already be an incompatibility issue. Can you confirm this @sansyrox ?
Also, refactoring to allow something else than a string may be a big refactor for issue #336, maybe it would be good to merge this and start a new PR to remove jsonify ?

cirospaciari:
> I agree with @cirospaciari, we shouldn't use `jsonify` for the headers.
> Although I think it is already handled like this in Robyn currently (see [this](https://github.com/sansyrox/robyn/blob/main/integration_tests/base_routes.py#L206) test route from the main branch for example) so there will already be an incompatibility issue. Can you confirm this @sansyrox ?
> Also, refactoring to allow something else than a string may be a big refactor for issue #336, maybe it would be good to merge this and start a new PR to remove jsonify ?


If already will be an incompatible issue so, merging and fix in a new PR is acceptable, but refactoring should be priority, I need to benchmark it first but the performance penalty will be huge without refactoring.

sansyrox:
>  but refactoring should be priority, I need to benchmark it first but the performance penalty will be huge without refactoring.

Agreed.

> Although I think it is already handled like this in Robyn currently (see [this](https://github.com/sansyrox/robyn/blob/main/integration_tests/base_routes.py#L206) test route from the main branch for example) so there will already be an incompatibility issue.

Yep, yep.

> Also, refactoring to allow something else than a string may be a big refactor for issue https://github.com/sansyrox/robyn/issues/336, maybe it would be good to merge this and start a new PR to remove jsonify ?

I have been working this for a few hours now. It is indeed a big refactor. :(

I think it makes sense to merge this one and refactor based on a priority.

PR #346, opened 2023-01-02
feat: return Response from routes

AntoineRR:
**Description**

This PR makes `Response` a `pyclass`.
There is no more need to convert a `Pydict` to a `Response` in the `execute_http_function` method, as the result will already be a `Response`.
Can you tell me what you think of it @sansyrox ? :slightly_smiling_face:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ dist/
 
 # DS_Store
 *.DS_Store
-.vscode/settings.json
+.vscode
 
 # python venv
 venv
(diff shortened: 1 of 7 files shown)


sansyrox:
@AntoineRR , this looks great :D

I will try running it locally before a final merge. 

Before that, I have one question - Is there any change in how the files are rendered in the output?

sansyrox:
@AntoineRR , I just had a look. This looks great!

sansyrox (approved):
LGTM! 👍

AntoineRR:
Thanks for reviewing and merging it!
Sorry I didn't see your question earlier, but I think I didn't change the way static files are handled anyway

PR #349, opened 2023-01-05
fix: better handling of route return type

AntoineRR:
**Description**

I figured there are issues when you try to return something other than a string or a dict in a route. With this PR, you can return anything and most notably directly a `Response`.
We may want to add it to the doc if it seems relevant to you @sansyrox ?

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -97,8 +97,44 @@ from robyn import jsonify
 
 @app.post("/jsonify")
 async def json(request):
-    print(request)
     return jsonify({"hello": "world"})
+```
+
+## Format of the Response
+Robyn supports several kind of Response for your routes
+
+#### Dictionary
+Robyn accepts dictionaries to build a response for the route:
+
+```python
+@app.post("/dictionary")
+async def dictionary(request):
+    return {
+        "status_code": 200,
+        "body": "This is a regular response",
+        "type": "text",
+        "headers": {"Header": "header_value"},
+    }
+```
+
+#### Response object
+Robyn provides a `Response` object to help you build a valid response.
+
+```python
+from robyn.robyn import Response
+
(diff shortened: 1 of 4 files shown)


sansyrox:
@AntoineRR , thank you for your PR. Can we still return a dictionary in this PR?

AntoineRR:
Yes sure!
Before, we could return only a `str` or `dict`, now you can still do that but also return an `int` or a `Response` for example.

sansyrox (approved):
Looks good to me! 🔥 

I am just wondering where will we add the documentation regarding this 🤔

AntoineRR:
Hey @sansyrox, thanks for your review. I added docs regarding the types returned from a route in the `features` section. I think it fits here, is it ok for you?

docs/features.md
```diff
@@ -89,16 +89,42 @@ async def json(request):
...
+@app.post("/json")
 async def json(request):
     print(request)
-    return jsonify({"hello": "world"})
+    return {
+        "status_code": 200,
+        "body": "This is a regular response",
+        "type": "text",
+        "headers": {"Header": "header_value"},
+    }
```
sansyrox, line 107: Hey @AntoineRR ,

The return type of json should be 

```python
return {
        "status_code": 200,
        "body":  jsonify({"hello": "world"},
        "headers": {"Content-Type": "application/json"},
    }
```

or just

```
```

It will be returning a Plain Text otherwise

AntoineRR: Oh my bad you're right, I will change this :)

AntoineRR: @sansyrox it should be ok now :slightly_smiling_face:

sansyrox (approved):
LGTM! Great work! 🔥

PR #352, opened 2023-01-09
Replace the default port with 8080

sansyrox:
OSX has some service running on port 5000 by default. It makes it annoying to work on Robyn to keep changing the default port number all the time.

**Description**

This changes the default PORT of Robyn.

### DIFF
--- a/integration_tests/test_env_populator.py
+++ b/integration_tests/test_env_populator.py
@@ -12,7 +12,7 @@ path = pathlib.Path(__file__).parent
 # create robyn.env before test and delete it after test
 @pytest.fixture
 def env_file():
-    CONTENT = """ROBYN_PORT=8080
+    CONTENT = """ROBYN_PORT=8081
     ROBYN_URL=127.0.0.1"""
     env_path = path / "robyn.env"
     env_path.write_text(CONTENT)
--- a/robyn/robyn.pyi
+++ b/robyn/robyn.pyi
@@ -21,6 +21,10 @@ class Response:
     headers: dict[str, str]
     body: str
 
+    def set_file_path(file_path):
+        pass
+
+
 class Server:
     def __init__(self) -> None:
         pass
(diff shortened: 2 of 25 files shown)


integration_tests/test_env_populator.py
```diff
@@ -12,7 +12,7 @@
 # create robyn.env before test and delete it after test
 @pytest.fixture
 def env_file():
-    CONTENT = """ROBYN_PORT=8080
+    CONTENT = """ROBYN_PORT=8081
```
AntoineRR, line 15: Why using 8081 here?

AntoineRR (commented):
Hi @sansyrox !

This looks fine to me and runs well on my linux machine. I just noticed there are a few places where port 5001 was used in the tests. Do you want to keep it like this? Maybe we should replace it with 8081 for consistency?

integration_tests/test_env_populator.py
sansyrox: @AntoineRR , as I wanted to test it on something other than the default port. Given that 8080 will be the default port from now on.

sansyrox: >I just noticed there are a few places where port 5001 was used in the tests. Do you want to keep it like this?

Sure, this sounds like a good idea. 😄

sansyrox:
@AntoineRR , all done. 😄

robyn/robyn.pyi
```diff
@@ -21,6 +21,10 @@ class Response:
     headers: dict[str, str]
     body: str
 
+    def set_file_path(file_path):
```
AntoineRR, line 24: Can you add `str` typing to the parameter here?

AntoineRR (approved):
Sorry I forgot to add one comment about the typing. Otherwise this LGTM, approved :wink:

robyn/robyn.pyi
sansyrox: @AntoineRR , yep sure 👍

sansyrox: done!

Issue #351, opened 2023-01-06
[Feature Request] Improve json serialization and deserialization speed


sansyrox:
**Current Behavior**

I have seen various faster libraries

**Desired Behavior**

We can use this:

- MSG Spec https://github.com/jcrist/msgspec

- Orjson 
- Or write our own

Need to assess the speed benefits

---
**Screenshots / Mockups**

**Alternatives**

---

kliwongan:
Has anyone started to work on this issue yet? I've been lurking around for some time and I wanted to try and see if I could get something started.

sansyrox:
Hey @kliwongan 👋 

You may proceed 😄  Which approach out of the three do you have in mind?

kliwongan:
Hello @sansyrox 

You mentioned that you wanted to compare the speeds of the three approaches; maybe I should try the first two (since the third seems a bit more challenging) and compare the performances first? I'll consult more contributors (and yourself) if I have any problems or questions.

kliwongan:
Also, I believe the msgspec approach is kind of similar to the FastAPI schema validation approach, so there may be complications?

sansyrox:
@kliwongan . sounds great! Let me know if you face any difficulties 😄

sansyrox:
JSON serialization was switched to orjson across the Python and Rust paths (jsonify.py, ws.py, pydantic responses, and the Rust executor), delivering the speedup requested here. Closing as resolved.

PR #355, opened 2023-01-13
Implement easy access query parameters and request body validation based on query parameters

### LINKED ISSUE #289: [Feature Request] Implement type defined easy access query parameters.
**Current Behavior**

_Originally from discord_

I'm talking about the type definition, and the controller receiving a predictable variable, exactly like the example above.
Everything in Robyn happens through a request dictionary, just like flask. meaning you don't know the keys, don't know their types, and don't have autocomplete
```
# This is Robyn
@app.post("/jsonify/:id") # the param ID
async def json(request): # you get a "request" dictionary, without any types
    print(request["params"]["id"]) # you access the param through the dict key, you don't validate the key type

# This is FastAPI
@app.post("/jsonify/{id}") # the param ID
async def json(id: int): # you excplitly tell the param and it's type (auto completion and type safety)
    print(id) # you already have the ID in hands, (auto complete)
```
**Desired Behavior**

Add type defined query params.
---
**Screenshots / Mockups**

**Alternatives**

---

kliwongan:
**Description**

This PR fixes #289 and may help with #351 (since msgspec encoding is apparently much faster than the standard json lib)
Query validation is optional, may find another way to indicate whether to verify (or not) instead of saving it in
the FunctionInfo, or force the change in general (when the "full" type-checking and schema validation is implemented)

The general idea of the fix is:
When a request is handled, call on the Python GIL (possible bottleneck) and
1. Get the signature of the model using the features in the inspect module. The current PR supports custom user-defined objects in the handler function signature and imported values from other files that have not been loaded yet (using ForwardRefs)
2. Generate a msgspec.Struct based model for schema validation
3. Compare the model against the received request (by decoding the JSON result into the model object) to see if the schema validation fails; if it does, an InternalServer error is raised.
4. If schema validation succeeds, we convert the object generated by the model validation into a dictionary and pass it to the handler function as kwargs and get the output of the handler function as usual

Further extensions
1. Currently, the user defined classes needs an __init__ definition (we can skip this by decorating the classes with @dataclass), however I wish to fix this by somehow extracting the members and their type annotations from the class type itself using getmembers and/or other tools in the inspect module
2. Implement process/thread-safe caching of Encoder and Decoder objects for the msgspec encode and decode calls. Since a call to either method requires some setup, multiple calls can be made faster if we use an Encoder/Decoder object for the respective calls.

Please do not merge to the main branch until item number 1 above is fixed!

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -97,7 +97,7 @@ class Robyn:
             return
 
         is_async = asyncio.iscoroutinefunction(handler)
-        self.event_handlers[event_type] = FunctionInfo(handler, is_async, 0)
+        self.event_handlers[event_type] = FunctionInfo(handler, is_async, 0, False)
 
     def startup_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.STARTUP, handler)
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -18,6 +18,7 @@ from robyn.responses import jsonify, serve_file, serve_html
 from robyn.robyn import FunctionInfo, SocketHeld
 from robyn.router import MiddlewareRouter, Router, WebSocketRouter
 from robyn.ws import WS
+from robyn.dependencies import get_signature, check_params_dependencies
 from robyn.env_populator import load_vars
 
 logger = logging.getLogger(__name__)
@@ -44,7 +45,7 @@ class Robyn:
         load_vars(project_root=directory_path)
         self._config_logger()
 
-    def _add_route(self, route_type, endpoint, handler, is_const=False):
+    def _add_route(self, route_type, endpoint, handler, is_const=False, validate_params=False):
         """
         [This is base handler for all the decorators]
 
@@ -55,7 +56,7 @@ class Robyn:
 
         """ We will add the status code here only
         """
-        return self.router.add_route(route_type, endpoint, handler, is_const)
+        return self.router.add_route(route_type, endpoint, handler, is_const, validate_params)
 
     def before_request(self, endpoint: str) -> Callable[..., None]:
         """
@@ -204,7 +205,7 @@ class Robyn:
 
         return inner
 
-    def post(self, endpoint: str):
+    def post(self, endpoint: str, validate: bool = False):
         """
         The @app.post decorator to add a get route
 
@@ -212,11 +213,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("POST", endpoint, handler)
+            return self._add_route("POST", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def put(self, endpoint: str):
+    def put(self, endpoint: str, validate: bool = False):
         """
         The @app.put decorator to add a get route
 
@@ -224,11 +225,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("PUT", endpoint, handler)
+            return self._add_route("PUT", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def delete(self, endpoint: str):
+    def delete(self, endpoint: str, validate: bool = False):
         """
         The @app.delete decorator to add a get route
 
@@ -236,11 +237,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("DELETE", endpoint, handler)
+            return self._add_route("DELETE", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def patch(self, endpoint: str):
+    def patch(self, endpoint: str, validate: bool = False):
         """
         [The @app.patch decorator to add a get route]
 
@@ -248,11 +249,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("PATCH", endpoint, handler)
+            return self._add_route("PATCH", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def head(self, endpoint: str):
+    def head(self, endpoint: str, validate: bool = False):
         """
         The @app.head decorator to add a get route
 
@@ -260,11 +261,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("HEAD", endpoint, handler)
+            return self._add_route("HEAD", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def options(self, endpoint: str):
+    def options(self, endpoint: str, validate: bool = False):
         """
         The @app.options decorator to add a get route
 
@@ -272,11 +273,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("OPTIONS", endpoint, handler)
+            return self._add_route("OPTIONS", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def connect(self, endpoint: str):
+    def connect(self, endpoint: str, validate: bool = False):
         """
         The @app.connect decorator to add a get route
 
@@ -284,11 +285,11 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("CONNECT", endpoint, handler)
+            return self._add_route("CONNECT", endpoint, handler, validate_params=validate)
 
         return inner
 
-    def trace(self, endpoint: str):
+    def trace(self, endpoint: str, validate: bool = False):
         """
         The @app.trace decorator to add a get route
 
@@ -296,7 +297,7 @@ class Robyn:
         """
 
         def inner(handler):
-            return self._add_route("TRACE", endpoint, handler)
+            return self._add_route("TRACE", endpoint, handler, validate_params=validate)
 
         return inner
 
--- a/robyn/dependencies.py
+++ b/robyn/dependencies.py
@@ -1,0 +1,38 @@
+
+from typing import Any, Callable, List, Dict
+from inspect import signature, Signature, Parameter
+from .utils import decode_bytearray
+import json
+
+def get_signature(call: Callable[..., Any]) -> Signature:
+    """
+    Gets the dependencies of the function wrapped by the routing decorators
+    (stores the name of the )
+    """
+    sign = signature(call)
+    params = [Parameter(
+        name = param.name,
+        kind = param.kind,
+        default = param.default,
+        # Add ForwardRef analysis here as the annotation value
+        annotation = param.annotation,
+    ) for param in sign.parameters.values()]
+
+    return Signature(params)
+
+def check_params_dependencies(expected: Signature, request: Dict[Any, Any]):
+        """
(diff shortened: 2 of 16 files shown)


sansyrox:
Thank you for the PR @kliwongan . The PR looks great on my first look. But I will have a deeper look today morning 😄

src/types.rs
```diff
@@ -14,16 +14,18 @@ pub struct FunctionInfo {
     pub handler: Py<PyAny>,
     pub is_async: bool,
     pub number_of_params: u8,
+    pub validate_params: bool,
```
sansyrox, line 17: Hey @kliwongan ,

Thanks for you PR 😄 . I have a few comments. 

I want to understand the reasoning behind naming the variable as `validate_params` , are we expecting a hashmap or an object in the future? If not, something like `to_validate` or `validation` would be more relevant.

robyn/__init__.py
```diff
@@ -204,99 +205,99 @@ def inner(handler):
...
+    def trace(self, endpoint: str, validate: bool = False):
         """
         The @app.trace decorator to add a get route
 
         :param endpoint str: endpoint to server the route
         """
 
         def inner(handler):
-            return self._add_route("TRACE", endpoint, handler)
+            return self._add_route("TRACE", endpoint, handler, validate_params=validate)
```
sansyrox, line 300: Also, is there a reason for passing `validate_params` as kwargs? And we can also name the variables in the outer and the inner functions for more consistency.

robyn/utils.py
```diff
@@ -0,0 +1,5 @@
+def decode_bytearray(req):
+    """
+    Decodes a bytearray into a UTF-8 string
+    """
+    return bytearray(req["body"]).decode("utf-8")
```
sansyrox, line 5: I would suggest keeping this in the `dependencies` file only as this is only used there for now.

sansyrox: Also, smart implementation here! Love it!

robyn/dependencies.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Gets the dependencies of the function wrapped by the routing decorators
+    (stores the name of the )
+    """
+    sign = signature(call)
+    params = [Parameter(
+        name = param.name,
+        kind = param.kind,
+        default = param.default,
+        # Add ForwardRef analysis here as the annotation value
```
sansyrox, line 17: What is a ForwardRef? I am not sure if I understand it completely.

src/server.rs
```diff
@@ -371,6 +372,35 @@ async fn index(
...
+                // Import dependency extraction
+                let robyn = py.import("robyn").unwrap();
+
+                // Extract the route's dependencies from the FunctionInfo
+                let dependencies = robyn.call_method1("get_signature", (handler,)).unwrap();
+
+                // Check if the request params match with the dependencies
+                let check_dependencies = robyn.call_method1("check_params_dependencies", (dependencies, request_hashmap,)).unwrap();
+            
+                cont = check_dependencies.extract().unwrap();
```
sansyrox, line 395: This kinda defeats the purpose of writing Robyn in Rust if we are importing a custom defined function from Python function. Have you considered writing `get_signature` and `check_params_dependenencies` in rust?

src/executors/mod.rs
```diff
@@ -19,12 +21,30 @@ fn get_function_output<'a>(
...
+        // Use Python to convert back to PyDict after deserializing
+        let json = py.import("json").unwrap();
+
+        // Extract the route's dependencies from the FunctionInfo
+        let json_dict = json.call_method1("loads", (json_str,)).unwrap();
+
+        // Get kwargs from the PyDict
+        let kwargs: &PyDict = json_dict.try_into().unwrap();
+
+        let test = handler.call((), Some(kwargs)).unwrap();
```
sansyrox, line 37: Let's give this variable a better name than `test`? Something like response/output?

src/executors/mod.rs
```diff
@@ -19,12 +21,30 @@ fn get_function_output<'a>(
...
-    match function.number_of_params {
+    if function.validate_params {
+        let body = request.body.clone().to_vec();
+        let json_str = str::from_utf8(&body).unwrap();
+
+        // Use Python to convert back to PyDict after deserializing
+        let json = py.import("json").unwrap();
+
+        // Extract the route's dependencies from the FunctionInfo
+        let json_dict = json.call_method1("loads", (json_str,)).unwrap();
```
sansyrox, line 32: I suppose you were also unable to find a crate that would allow us to convert a string to a hashmap or a pydict. I think we will have to write a `macro!` for this. I was afraid things might come to this. Maybe Pydantic v2 could help us here?

src/executors/mod.rs
```diff
@@ -1,6 +1,7 @@
 /// This is the module that has all the executor functions
 /// i.e. the functions that have the responsibility of parsing and executing functions.
 use crate::types::{FunctionInfo, Request, Response};
+use std::str;
```
sansyrox, line 4: We don't need to import this, we can directly use `std::str:from_utf8()`

src/executors/mod.rs
sansyrox: We can leave it in if you're not very familiar with macros! I'm not very either. I'll have to read more of it.

sansyrox (approved):
Hey @kliwongan ,

Thank you for the great work! 😄 

I have a few inline comments. Do let me know if you have issues with any of them and I would love to discuss more 😄

robyn/utils.py
kliwongan: Thanks!

robyn/__init__.py
kliwongan: I used kwargs because is_const comes before validate_params in _add_route (at least I kept it like that). Perhaps I didn't need to do that?

src/types.rs
kliwongan: Yes, I agree. I simply wanted the user to have the choice of taking in a raw request, or define the request body using the function signature, hence the boolean variable.

robyn/dependencies.py
kliwongan: This explains it perfectly: https://peps.python.org/pep-0484/#forward-references
Basically, type hints that use classes defined in other files cannot be resolved immediately because of how (circular) imports work in Python (although I don't know the specifics of that). Therefore, ForwardRefs are used when these annotations cannot be immediately resolved. I saw these being used when I was reverse engineering FastAPI's dependency injection system here: https://github.com/tiangolo/fastapi/blob/5905c3f740c8590f1a370e36b99b760f1ee7b828/fastapi/dependencies/utils.py#L248 in the get_typed_annotation function.
Since the callable's signature is retrieved and checked in a different file (inevitably) than it was defined in, any classes defined in the annotation would "external" relative to the function that was retrieving their signature. Thus we need to use ForwardReferences in virtually all cases where we have custom defined classes in the type hint. I am planning on working on this in a follow up PR to this one. I'm also going to try and eliminate external dependencies (like Pydantic) from this mechanism/implementation.

src/server.rs
kliwongan: I've considered perhaps writing both functions in Rust, and if prompted to, I'd probably find a way to do it.
However, it's a little hard to do so since I have to consider working with ForwardRefs in get_signature. The reason I didn't do so for check_params_dependencies is because I have plans to use msgspec not only for json encoding but also for schema validation, since the two features work hand in hand with each other, and it is also possible to define custom classes that represent JSON nesting that do not need to extend from any msgspec's classes, so the user doesn't actually need to learn how to use msgspec (while FastAPI requires the user to define Pydantic models).

src/executors/mod.rs
kliwongan: I'll read more of it too!

src/server.rs
kliwongan: However, I realize that like in other sections of the code, I could simply import the libs in Rust and write the function in Rust where I'd just call the specific Python functions using the call_method1. However, I'm not sure if this is what you meant. Either way, I'll try my best.

src/types.rs
kliwongan: However, I wasn't sure whether or not to embed this in the FunctionInfo struct or not, as for middleware-related and websocket-related functions, validation is not really applicable?

robyn/dependencies.py
kliwongan: I might be wrong about needing to use ForwardRefs, though. I will investigate further.

src/executors/mod.rs
kliwongan: I've just realized that pythonize (by the author of PyO3) exists. If we convert the string to a serde_json::Value, we can use   pythonize to convert the Value enum/object into a PyDict (since Value implements the Deserialize trait). I am going to try this approach.

src/types.rs
sansyrox: @kliwongan , you can make it an `Optional<bool>`

robyn/__init__.py
sansyrox: @kliwongan , we can change that. :)

>  is_const comes before validate_params in _add_route (at least I kept it like that). Perhaps I didn't need to do that?

src/types.rs
kliwongan: You mean an Option<bool>? I was thinking of setting the default argument of validate as False (since dataclasses can detect those things) since an Option in Rust requires unwrapping it and checking for None, which adds an additional layer of checking, but I'm not sure how that would work with PyO3.

src/executors/mod.rs
sansyrox: Perfect @kliwongan 😄 Let me know if you need any help from my end 😄

src/server.rs
sansyrox: @kliwongan , this is exactly what I meant. As creating a binding for a library and then reimporting them feels like circular imports, I am not sure if they will cause any issues or not but this way allows us to keep the code style homogenous throughout. 

> However, I realize that like in other sections of the code, I could simply import the libs in Rust and write the function in Rust where I'd just call the specific Python functions using the call_method1.

src/executors/mod.rs
kliwongan: @sansyrox Update, I found out that msgspec's schema validation is very nice, so I am using that to pass in the query parameters (instead of Pythonizing, I get the output of the model decode/schema validation in the form of the objects/default classes for the function signature) however I have tested the Pythonize approach and that works. I am also thinking of moving the query param validation in get_function_output to put all of the robyn Rust imports in one place. This, and making sure ForwardRefs work properly will take time, but I will complete this in time.

kliwongan:
Please go to the other PR, I accidentally worked on another branch so as to not contaminate the other one!

PR #362, opened 2023-01-16
test: speed up tests

AntoineRR:
**Description**

This PR speeds up tests by a lot. On my computer, running the tests took around 40s and it now runs in around 5s.
To achieve this, I decreased the sleep time when waiting for a server, and changed the librairy used for the websocket client.
Ideally, we should wait for a signal that let us know the server has started but I could not figure out how to do that for now.
Also, I removed the `Arc` around `task_locals` in the Rust code as it seems it wasn't needed.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,7 @@
 -r requirements.txt
 flake8==4.0.1
 black==21.12b0
-websockets==10.1
+websocket-client==1.4.2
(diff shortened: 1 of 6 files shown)


sansyrox (approved):
The speedup is crazy! Wow 🔥

sansyrox:
Great work Antoinne! ❇️

PR #364, opened 2023-01-17
Move pre-commit hooks to use Ruff

patrick91:
**Description**

Ruff is a much faster linter (written in rust), it also supports
sorting imports (replacing isort) and fixing issues.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,14 +5,10 @@ repos:
     hooks:
       - id: black
         args: [--line-length=160]
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.224
     hooks:
-      - id: flake8
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.11.4
-    hooks:
-      - id: isort
+      - id: ruff
 
 ci:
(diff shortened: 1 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -5,14 +5,10 @@ repos:
     hooks:
       - id: black
         args: [--line-length=160]
```
patrick91, line 7: is 160 really what we want? :)

patrick91: ah, btw, the default_stages at the top might not be what we'd like, @sansyrox was there a particular reason it is only set to push?

sansyrox: @patrick91 

> is 160 really what we want? :)

This was just something random. I just feel that 70 characters is a bit too less :p

> ah, btw, the default_stages at the top might not be what we'd like, @sansyrox was there a particular reason it is only set to push?

@patrick91 , I am sorry but I don't understand what you mean by this. Can you please explain a bit more? 😅

sansyrox:
Thank you @patrick91 😄 . I have a few questions on the PR.

.pre-commit-config.yaml
sansyrox: > ah, btw, the default_stages at the top might not be what we'd like, @sansyrox was there a particular reason it is only set to push?

Now, I understand. It was mistake 🤣

PR #368, opened 2023-01-21
test: improve pytest fixtures

AntoineRR:
**Description**

Found a way to actually wait for the server to start instead of calling `time.sleep` in the pytest fixtures. This improves the test speed again, from 5s currently to 0.5s on my machine. This PR with the previous one make the tests around 80x faster than previously! :partying_face:
It should also be safer because we wait for the server to be reachable, if it is not reachable in 5s then the test will fail.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import signal
+import socket
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


sansyrox:
Thanks for the PR @AntoineRR 😄 

The PR looks good from the first glance but I will give it a thorough review in a few hours 😄

sansyrox (approved):
The speed upgrade is just massive! Wow! 🔥

PR #370, opened 2023-01-23
chore: run tests on linux, macos and windows and release builds on ta…

### LINKED ISSUE #310: [BUG] Fix CI/CD pipeline
**Description**

- The `Preview release` job fails for every PR.
- Also, when this job runs on master, it displays the following logs for every build:
```
WARNING  Skipping robyn-0.18.3-cp310-cp310-macosx_10_7_x86_64.whl because it    
         appears to already exist
```

If I understand correctly, the `Preview Release` job should be run for PRs only. In PRs, it seems the job is failing because of an authentication error:
```
ERROR    HTTPError: 403 Forbidden from https://test.pypi.org/legacy/            
         Invalid or non-existent authentication information. See                
         https://test.pypi.org/help/#invalid-auth for more information.
```
Maybe the `TEST_PYPI_PASSWORD` secret is not available in PRs which makes sense from a security point of view (see https://stackoverflow.com/questions/73866908/github-actions-main-repository-secret-not-picked-up-from-pull-request-build)

- The Rust CI displays warnings:
```
Node.js 12 actions are deprecated.
```
This is because the [actions-rs](https://github.com/actions-rs/toolchain) toolchain project is not maintained anymore.

**Expected Behavior**

Do not run the `Preview Release` job on push? Find a way to retrieve the secrets on PRs or maybe remove the job completely?

I know how to get rid of the warnings in the Rust pipeline, but would like some guidance on the `Preview Release` job purpose and fix if possible.

AntoineRR:
**Description**

This PR fixes #310

This is a proposal for a new CI/CD pipeline:
- On push and pull requests -> Run rust CI + Tests on Linux / Windows / Macos for python 3.7, 3.8, 3.9, 3.10, 3.11
- On tag creation -> Build for all OS and Python versions and upload to pypi

It seems like the tests do not pass on windows for now, adding this CI could be a good first step to fix them.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 requests==2.26.0
 pytest==6.2.5
 maturin
-uvloop
+uvloop; platform_system!="Windows"
 watchdog
 multiprocess==0.70.12.2
--- a/robyn/test-requirements.txt
+++ b/robyn/test-requirements.txt
@@ -2,7 +2,7 @@ pytest==6.2.5
 maturin==0.12.11
 watchdog
 requests==2.26.0
-uvloop==0.17.0
+uvloop==0.17.0; platform_system!="Windows"
 multiprocess==0.70.12.2
 websocket-client==1.4.2
 jinja2==3.0.2
--- a/.github/workflows/python-CI.yml
+++ b/.github/workflows/python-CI.yml
@@ -1,17 +1,18 @@
-# CI to build the Python code
+# CI to test Robyn on major Linux, MacOS and Windows
 
 on: [push, pull_request]
 
 name: Python Continuous integration
 
 jobs:
-  test:
-    name: Test
-    runs-on: ubuntu-latest
+  tests:
     strategy:
+      fail-fast: false
       matrix:
+        os: ["windows", "ubuntu", "macos"]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
-
+    name: ${{ matrix.os }} tests with python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}-latest
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 requests==2.26.0
 pytest==6.2.5
 maturin
-uvloop
+uvloop; platform_system!="Windows"
```
sansyrox, line 4: niceee.

robyn/test-requirements.txt
```diff
@@ -2,7 +2,7 @@ pytest==6.2.5
 maturin==0.12.11
 watchdog
 requests==2.26.0
-uvloop==0.17.0
+uvloop==0.17.0; platform_system!="Windows"
```
sansyrox, line 5: nicee..

sansyrox:
> It seems like the tests do not pass on windows for now

I might have a hunch why this is happening. The `sockets` package is not super compatible with windows. Read the first line here(https://docs.python.org/3/library/socket.html)

Just one question , how is the proposal different from our current way? Don't we still build  a package on every PR and push to main?

And release on tags?

requirements.txt
AntoineRR: I didn't know that was a thing until I worked on this PR :smile:

AntoineRR:
Ok thanks, that's a first step for investigating the comptibility with Windows.

There is a difference, as with this PR there are a lot less jobs running on push and PRs (see the pipeline of this PR for reference). Also, previously we built a release package on every push and PRs, but didn't do anything with it. The only time when the builds were used was when a tag was released that would trigger the "Release" job for an upload to Pypi.

In summary:

Previous pipeline:
- Push / PR => Run Rust CI + Run tests on Linux with each Python version + Build a Release package for each OS and each Python version
- Tag creation => All of the above + Publish all Release to Pypi

This pipeline:
- Push / PR => Run Rust CI + Run tests on all OS with each Python version
- Tag creation => All of the above + Build a Release package for each OS and each Python version + Publish all Release to Pypi

I didn't test the tag creation pipeline, but kept the same jobs as before so it should work fine

sansyrox:
@AntoineRR , in this PR , are we depending on the tests to pass to build the packages?

AntoineRR:
@sansyrox no, the test and release workflow are independant of each other

sansyrox (approved):
This looks good! Now we have more information about windows builds!

PR #375, opened 2023-01-27
Changed Response to use body: bytes

### LINKED ISSUE #374: [BUG] Can't send raw bytes
**Description**

I can't return bytes without getting an error that bytes can't be converted to PyString.
Digging into the code I think this is due to the `struct Response` using String instead of bytes.

**Expected Behavior**

Web servers should be able to return `bytes`.

**Screenshots**

**Environment:**

- Host OS:
- Browser:

---

I figured i'd take a crack at it in a PR.

---

madhavajay:
**Description**
- Added ActixBytesWrapper to conform to pyclass and MessageBody

This PR fixes #374 

I know that my code is not optimal, but I wanted to see what the problem was and if i could get it to work.

I don't have a lot of experience with Rust or Maturin / pyo3 and I get really confused on the lifetimes, copying of bytes etc.

Was there a different goal in mind for allowing Robyn to return bytes?
One thing that is a problem is, if the Response() struct `body` contains a string I can't detect it in the `_format_response` method because the rust struct has no getters. This means a python user can use a string in body and it will fail.

It would be good to be able to detect this, or maybe the Rust struct needs to take a Union type and detect the string and convert it then?

I look forwards to feedback on my PR! 🦀


sansyrox:
Thank you for the PR @madhavajay 😄 The pr looks nice from the first glance but I will do a thorough review tomorrow morning 😄

src/types.rs
```diff
@@ -1,13 +1,75 @@
...
 
+
+#[derive(Debug, Clone)]
+#[pyclass]
+pub struct ActixBytesWrapper(Bytes);
+
+impl ActixBytesWrapper {
+    pub fn new(bytes: &PyBytes) -> Self {
+        Self(Bytes::from(bytes.as_bytes().to_vec()))
+    }
```
AntoineRR, line 30: Here is how you could handle both `bytes` and `str` from Python:
```suggestion
    pub fn new(raw_body: &PyAny) -> PyResult<Self> {
        let value = if let Ok(v) = raw_body.downcast::<PyString>() {
            v.to_string().into_bytes()
        } else if let Ok(v) = raw_body.downcast::<PyBytes>() {
            v.as_bytes().to_vec()
        } else {
            return Err(PyValueError::new_err(
                "Could not convert specified body to bytes",
            ));
        };
        Ok(Self(Bytes::from(value)))
    }
```

src/types.rs
```diff
@@ -100,10 +162,11 @@ impl Response {
         // we should be handling based on headers but works for now
         self.response_type = "static_file".to_string();
         self.file_path = Some(file_path.to_string());
-        self.body = match read_file(file_path) {
+        let string = match read_file(file_path) {
             Ok(b) => b,
             Err(e) => return Err(exceptions::PyIOError::new_err::<String>(e.to_string())),
         };
+        self.body = ActixBytesWrapper(Bytes::from(string.clone()));
```
AntoineRR, line 169: Unnecessary clone

src/types.rs
```diff
@@ -78,20 +140,20 @@ pub struct Response {
...
-    pub body: String,
+    pub body: ActixBytesWrapper,
     pub file_path: Option<String>,
 }
 
 #[pymethods]
 impl Response {
     #[new]
-    pub fn new(status_code: u16, headers: HashMap<String, String>, body: String) -> Self {
+    pub fn new(status_code: u16, headers: HashMap<String, String>, body: &PyBytes) -> Self {
```
AntoineRR, line 150: The signature of this method also has to change a little:
```suggestion
    pub fn new(status_code: u16, headers: HashMap<String, String>, body: &PyAny) -> PyResult<Self> {
```

robyn/router.py
```diff
@@ -28,7 +28,7 @@ def _format_response(self, res):
         if type(res) == dict:
             status_code = res.get("status_code", 200)
             headers = res.get("headers", {"Content-Type": "text/plain"})
-            body = res.get("body", "")
+            body = str(res.get("body", "")).encode("utf-8")
```
AntoineRR, line 31: After changing the Rust code to take either `bytes` or an `str`, this line can be reverted.

AntoineRR (changes requested):
Hi @madhavajay, and thanks for this PR!
I found a way to fix the issue you had with breaking compatibility with `str` return type. I left some suggestions that will make your code compatible with both `bytes` and `str`.
The `ActixBytesWrapper` struct is actually very useful and it could probably help with other features in the future too, good job on this!

AntoineRR:
@sansyrox feel free to add your comments too if you have some, I focused mostly on the `str` vs `bytes` issue :wink:

integration_tests/base_routes.py
```diff
@@ -270,12 +270,32 @@ def file_download_sync():
...
+
+
+@app.get("/binary_output_async")
+async def binary_output(request):
+    return b"OK"
+    # response = Response(
+    #     status_code=200,
+    #     headers={"Content-Type": "application/octet-stream"},
+    #     body=b"OK",
+    # )
```
sansyrox, line 296: Can you make two different test cases for this? That way we will be increasing the code coverage as we will still be testing the `Response` constructor type syntax and also the `b"` type syntax

src/types.rs
```diff
@@ -100,10 +162,11 @@ impl Response {
         // we should be handling based on headers but works for now
         self.response_type = "static_file".to_string();
         self.file_path = Some(file_path.to_string());
-        self.body = match read_file(file_path) {
+        let string = match read_file(file_path) {
```
sansyrox, line 165: We can name this thing a bit better? Something like `string_response` or  just `response`?

sansyrox (commented):
Hey @madhavajay 👋 

Thank you for your PR. The `ActixBytesWrapper` is a very helpful struct 😄 and thank you @AntoineRR for the review 😄 

I have a few inline comments in addition to @AntoineRR 's review

src/types.rs
madhavajay: ❤️ Thank you, my knowledge of pyo3 is limited and knowing about this `downcast` is awesome.

madhavajay:
@AntoineRR and @sansyrox thanks for the feedback. I have made the changes.

I thought it might be a good idea to include a bad input test but the async one doesn't await and i'm not sure how to get `pytest` to work with `async`?

src/types.rs
AntoineRR: You're welcome, it's a pleasure to help 😉

AntoineRR:
That's a good idea to include a test for bad input. It should however be done without the `session` fixture, just by creating the `Response` struct with an invalid body and asserting we get a `ValueError` Exception IMO. Doing it this way there is no need to call a route on the server.

madhavajay:
@AntoineRR I tried but it seems like the exceptions aren't catchable right now so I think ill just remove this so we can merge and then this can be addressed separately.

Screen Shot 2023-01-30 at 8 12 29 pm
Screen Shot 2023-01-30 at 8 12 10 pm

src/types.rs
```diff
@@ -78,32 +154,33 @@ pub struct Response {
...
-    pub fn new(status_code: u16, headers: HashMap<String, String>, body: String) -> Self {
-        Self {
+    pub fn new(status_code: u16, headers: HashMap<String, String>, body: &PyAny) -> PyResult<Self> {
+        return Ok(Self {
             status_code,
             // we should be handling based on headers but works for now
             response_type: "text".to_string(),
             headers,
-            body,
+            body: ActixBytesWrapper::new(body).unwrap(),
```
AntoineRR, line 170: @madhavajay this is the line that makes your test fail. You shouldn't unwrap here as it will make the Rust code panic. Instead, you want the `new` function to return the error. It will then be interpreted to raise an `Exception` in Python, and you will be able to catch it in the test.

```suggestion
            body: ActixBytesWrapper::new(body)?,
```

madhavajay: Ah, yes this is much better thank you!
I have added the tests back.

integration_tests/test_unsupported_types.py
```diff
@@ -0,0 +1,41 @@
...
+        True,
+        A,
+        {"body": "OK"},
+        ["OK", b"OK"],
+        Response(
+            status_code=200,
+            headers={},
+            body=b"OK",
+        ),
+    ]
```
sansyrox, line 22: Let's use parameterization here: https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

integration_tests/test_unsupported_types.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_good_body_types():
+    good_bodies = ["OK", b"OK"]
+
+    for body in good_bodies:
+        _ = Response(
+            status_code=200,
+            headers={},
+            body=body,
+        )
```
sansyrox, line 41: same here(https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids)

src/types.rs
```diff
@@ -78,32 +154,33 @@ pub struct Response {
...
     pub file_path: Option<String>,
 }
 
 #[pymethods]
 impl Response {
     #[new]
-    pub fn new(status_code: u16, headers: HashMap<String, String>, body: String) -> Self {
-        Self {
+    pub fn new(status_code: u16, headers: HashMap<String, String>, body: &PyAny) -> PyResult<Self> {
+        return Ok(Self {
```
sansyrox, line 165: I don't think we need this return statement. We can just return 

```
Ok(Self { .... })

```

sansyrox (commented):
Hey @madhavajay 👋 

Thank you for the updates. 😄 

I just have a few final suggestions

src/types.rs
madhavajay: Done

sansyrox (commented):
This looks good to me 😄 

@AntoineRR , do you want to give it a final glance?

AntoineRR:
LGTM :+1:

AntoineRR (approved):

sansyrox:
@madhavajay , there's just one final conflict. Everything else looks good 😄

AntoineRR:
The conflict was with my test refactor, I solved it from Github :slightly_smiling_face:

sansyrox:
Thank you @madhavajay and @AntoineRR . Great work 😄

PR #377, opened 2023-01-28
test: organize and add tests

AntoineRR:
**Description**

I felt like the tests needed to be better organized, so it would be easier to write new tests and to check the unit test coverage. This PR does the following to achieve those goals:
- Add a `utils.py` file that helps with requesting endpoints, and checks the status code and global headers of the response.
- Test everything with sync and async endpoints
- Add tests for adding a global header, startup handler and shutdown handler, and a test for modifying request headers with middlewares (which is skipped for now as this is currently bugged in Robyn)
- Rename endpoints for a better understanding (sync endpoints start with `/sync` and async one with `/async` for example)
- Fix some docstrings

There is still room for improvements, like actually testing if the startup and shutdown handlers are executed, but I didn't figure out a way to test this properly for now.

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -1,34 +1,32 @@
-import asyncio
-import logging
 import os
 import pathlib
 
 from robyn import WS, Robyn, jsonify, serve_file, serve_html
-from robyn.log_colors import Colors
 from robyn.robyn import Response
 from robyn.templating import JinjaTemplate
 
 app = Robyn(__file__)
 websocket = WS(app, "/web_socket")
-i = -1
 
-logger = logging.getLogger(__name__)
 current_file_path = pathlib.Path(__file__).parent.resolve()
 jinja_template = JinjaTemplate(os.path.join(current_file_path, "templates"))
+
+# ===== Websockets =====
+
+websocket_state = 0
 
 
 @websocket.on("message")
 async def connect(websocket_id):
-    print(websocket_id)
-    global i
-    i += 1
-    if i == 0:
-        return "Whaaat??"
-    elif i == 1:
-        return "Whooo??"
-    elif i == 2:
-        i = -1
-        return "*chika* *chika* Slim Shady."
+    global websocket_state
+    if websocket_state == 0:
+        response = "Whaaat??"
+    elif websocket_state == 1:
+        response = "Whooo??"
+    elif websocket_state == 2:
+        response = "*chika* *chika* Slim Shady."
+    websocket_state = (websocket_state + 1) % 3
+    return response
 
 
 @websocket.on("close")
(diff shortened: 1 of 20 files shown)


sansyrox:
Hey @AntoineRR 👋 

Thanks for the PR - the `pre-commit.ci - pr ` is failing. I will do a thorough review in 20 minutes.

AntoineRR:
Oh I didn't notice the precommit failing, I will fix that thanks

robyn/argument_parser.py
```diff
@@ -35,7 +35,7 @@ def __init__(self) -> None:
             help="Set the log level name",
         )
 
-        self.args = self.parser.parse_args()
+        self.args, unknown = self.parser.parse_known_args()
```
sansyrox, line 38: nice!! I wasn't aware of this function 👍

integration_tests/utils.py
```diff
@@ -0,0 +1,68 @@
+from typing import Optional
```
sansyrox, line 1: One nit @AntoineRR ,

I am not a big fan of naming a file as `utils` as I believe that a `utils` file eventually becomes a dump of functions we don't understand where to put.

Can you name we name this file like `http_methods_helpers` or `http_methods_testing_helpers` ?

integration_tests/test_middlewares.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from utils import get
+
+
+@pytest.mark.skip(reason="Fix middleware request headers modification")
```
sansyrox, line 5: This is broken?

integration_tests/test_basic_routes.py
```diff
@@ -0,0 +1,73 @@
+# Test the routes with:
```
sansyrox, line 1: Do you think something like Parameterizing will be helpful here(https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids) ?

integration_tests/test_app.py
```diff
@@ -0,0 +1,33 @@
...
+    assert startup.handler == mock_startup_handler
+    assert startup.is_async is False
+    assert startup.number_of_params == 0
+
+    app.shutdown_handler(mock_shutdown_handler)
+    assert Events.SHUTDOWN in app.event_handlers
+    shutdown = app.event_handlers[Events.SHUTDOWN]
+    assert shutdown.handler == mock_shutdown_handler
+    assert shutdown.is_async is True
+    assert shutdown.number_of_params == 0
```
sansyrox, line 33: Nice!!

integration_tests/base_routes.py
```diff
@@ -41,239 +39,412 @@ def message():
     return "Hello world, from ws"
 
 
-callCount = 0
+# ===== Lifecucle handlers =====
```
sansyrox, line 42: a minor typo here

integration_tests/base_routes.py
```diff
@@ -41,239 +39,412 @@ def message():
...
+
+
+async def startup_handler():
+    print("Starting up")
+
+
+@app.shutdown_handler
+def shutdown_handler():
+    print("Shutting down")
```
sansyrox, line 51: are we using these anymore with the new tests?

integration_tests/base_routes.py
```diff
@@ -41,239 +39,412 @@ def message():
...
 @app.get("/")
-async def hello(request):
-    global callCount
-    callCount += 1
-    message = "Called " + str(callCount) + " times"
-    print(message, request)
-    return {"status_code": 200, "body": "hello", "type": "text"}
+async def middlewares():
+    return ""
+
```
sansyrox, line 72: Same question - Are we using these?

sansyrox (commented):
Amazing work @AntoineRR ! 🔥  This PR must've required a lot of work.

I have a few questions and suggestions. Let me know what you think of them 😄

AntoineRR:
Thanks for the review ! :wink: I will go through it and let you know when I did the changes

robyn/argument_parser.py
AntoineRR: I had some issues while debugging because some unknown flags were passed to the file, this method fixed the issue because it lets you ignore unknown flags :wink:

integration_tests/utils.py
AntoineRR: Sure! I must say I got used to `utils` but I have no issue renaming the file

integration_tests/test_middlewares.py
AntoineRR: Yes, I started investigating it but I'm actually waiting for PR #375 to be merged because I think its new struct will help as we could make `Request` a pyclass and pass its instance to the middlewares (and routes) instead of a hashmap, and that should help fix this. We could also manipulate a true `Request` object in the routes and misslewares instead of a raw dict, which would be much nicer!

integration_tests/test_basic_routes.py
AntoineRR: Much better indeed, I struggled to make it work with fixtures but it is done now, I will push it soon :slightly_smiling_face:

integration_tests/base_routes.py
AntoineRR: It is still registered for the app, but I couldn't find a way to test they are working correctly. I would like them to imply a side effect I could then check in the test. I left it here for now until someone finds a way to actually write a test for it.

integration_tests/base_routes.py
AntoineRR: Yes, the new test in `test_middlewares.py` uses this route! It doesn't work for now but will hopefully soon be fixed

AntoineRR:
PR updated @sansyrox :slightly_smiling_face:

integration_tests/test_basic_routes.py
```diff
@@ -0,0 +1,62 @@
...
+        ("/async/str/const", "async str const get", None, None),
+        ("/async/dict/const", "async dict const get", "async_const", "dict"),
+        (
+            "/async/response/const",
+            "async response const get",
+            "async_const",
+            "response",
+        ),
+    ],
+)
```
sansyrox, line 34: Nice! ❇️

integration_tests/test_get_requests.py
```diff
@@ -1,96 +1,47 @@
...
+    def check_response(r: Response):
+        assert r.text.startswith("\n\n<!DOCTYPE html>")
+        assert "Jinja2" in r.text
+        assert "Robyn" in r.text
 
-def test_jinja_template(session):
-    r = requests.get(f"{BASE_URL}/template_render")
-    assert "Jinja2" in r.text
-    assert "Robyn" in r.text
+    check_response(get("/sync/template"))
```
sansyrox, line 32: One last nit @AntoineRR - let's split the async and the sync tests in two different parts? I know this will make the test suite a bit slower but it will make things easier to test.

sansyrox (commented):
@AntoineRR , all looks good! 😄 

I just have one final comment.

sansyrox (approved):
Perfect. Looks good to me. Great work! 😄

PR #378, opened 2023-01-29
ci: build artifacts on every push and pull

sansyrox:
**Description**

Changing this as I want to build an artifact on every PR. That way it will be easier for us to just download the artifact and test it on our PC.

### DIFF
--- a/.github/workflows/preview-deployments.yml
+++ b/.github/workflows/preview-deployments.yml
@@ -13,14 +13,11 @@ on:
 jobs:
   macos:
     runs-on: macos-latest
-    strategy:
-      matrix:
-        python-version: ["3.7"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.7"
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
@@ -32,11 +29,7 @@ jobs:
         uses: messense/maturin-action@v1
         with:
           target: x86_64
-          args: -i python --release --out dist
-      - name: Install build wheel - x86_64
-        run: |
-          pip install --force-reinstall dist/robyn*.whl
-          cd ~ && python -c 'import robyn'
+          args: -i python --release --out dist --no-sdist
       - name: Build wheels - universal2
         uses: messense/maturin-action@v1
         with:
@@ -51,13 +44,12 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: ["3.7"]
         target: [x64, x86]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.7"
           architecture: ${{ matrix.target }}
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
@@ -69,7 +61,7 @@ jobs:
         uses: messense/maturin-action@v1
         with:
           target: ${{ matrix.target }}
-          args: -i python --release --out dist --no-sdist
+          args: -i python3.7 --release --out dist --no-sdist
       - name: Upload wheels
         uses: actions/upload-artifact@v3
         with:
@@ -80,7 +72,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7"]
         target: [x86_64, i686]
     steps:
       - uses: actions/checkout@v3
@@ -92,13 +83,13 @@ jobs:
           default: true
       - uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.7"
       - name: Build Wheels
         uses: messense/maturin-action@v1
         with:
           target: ${{ matrix.target }}
           manylinux: auto
-          args: -i python${{ matrix.python-version }} --release --out dist --no-sdist
+          args: -i python3.7 --out dist --no-sdist
       - name: Upload wheels
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


sansyrox:
Hey @AntoineRR 👋 

The rationale behind this is that the top-level tag checking doesn't allow the jobs to run on every PR, and hence we won't be generating build artifacts when required.

AntoineRR:
Yes that was the goal of my previous PR actually, but I had no idea you used the artifacts from PR sorry. I found it overkill to build the package for every OS, and every Python version on each PR. I think there are way too many jobs running in PRs (more than 120 in this PR currently!).

Maybe we could only build one package for windows, linux and macos on PRs (based on Python 3.11 for example)? This way we would still have artifacts to test, but only 3 instead of I think 45 currently.

To achieve this, I would create a new workflow (called preview build maybe) that runs on PRs and keep the release workflow as it is.

What do you think @sansyrox ? Let me know if I misunderstood your needs around build artifacts.

sansyrox:
> Maybe we could only build one package for windows, linux and macos on PRs (based on Python 3.11 for example)? This way we would still have artifacts to test, but only 3 instead of I think 45 currently.

This is a great suggestion. Let's build the packages for the latest versions only

AntoineRR:
Thanks! Let me know if you need help for this :smile:

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,128 @@
...
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  macos:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        python-version: ["3.7"]
```
sansyrox, line 18: I am building on the oldest version instead of the latest version to ensure the most backward compatibility in every release

sansyrox:
@AntoineRR , can you have another review whenever you get the time?

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,128 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+  linux-cross:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [{ version: "3.7", abi: "cp37-cp37m" }]
```
AntoineRR, line 112: Remove from matrix

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,128 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build Wheels
+        uses: messense/maturin-action@v1
+        env:
+          PYO3_CROSS_LIB_DIR: /opt/python/${{ matrix.python.abi }}/lib
+        with:
+          target: ${{ matrix.target }}
+          manylinux: auto
+          args: -i python3.9 --release --out dist --no-sdist
```
AntoineRR, line 123:
```suggestion
          args: -i python3.7 --release --out dist --no-sdist
```

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,128 @@
+# CI to release the project for Linux, Windows, and MacOS
+
+name: Upload Preview Packages
+
+on:
+  push:
+    branches:
+      - main
```
AntoineRR, line 8: Is this needed ? If we do PRs each time the action will run in the PR and again when merged to main, do we want this ?

AntoineRR (changes requested):
@sansyrox I left you some comments :slightly_smiling_face: 
Two more things here:
- I think we could merge the last two jobs, `linux` and `linux-cross` together, it should be possible I think
- Is there an issue with the commit history? I think you merged your changes with one of my previous commit based on the commit message. Maybe you should reopen a new PR if it is too hard to fix?

.github/workflows/preview-deployments.yml
sansyrox: @AntoineRR , I have done it for ease of use. I would love to click on the final merge and get the artifacts.

sansyrox:
> I think we could merge the last two jobs, linux and linux-cross together, it should be possible I think

These two environments require different logic. It will make the pipeline unnecessarily complex in my opinion.

> Is there an issue with the commit history? I think you merged your changes with one of my previous commit based on the commit message. Maybe you should reopen a new PR if it is too hard to fix?

All fixed :D

.github/workflows/preview-deployments.yml
AntoineRR: Ok makes sense ^^

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,128 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+  windows:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: ["3.7"]
```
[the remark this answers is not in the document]
AntoineRR: You forgot this one :wink:

.github/workflows/preview-deployments.yml
```diff
@@ -0,0 +1,120 @@
...
+          default: true
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.7"
+      - name: Build Wheels
+        uses: messense/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          manylinux: auto
+          args: -i python${{ matrix.python-version }} --release --out dist --no-sdist
```
AntoineRR, line 93: Replace `{{ matrix...}}` with 3.7

AntoineRR (changes requested):
There are some minor adjustments needed, but this looks great otherwise, good job! :wink:

.github/workflows/preview-deployments.yml
sansyrox: Thanks @AntoineRR 😄

AntoineRR (approved):
Good job @sansyrox, LGTM! :+1:

PR #380, opened 2023-01-29
fix: return 404 status code when route isn't found

### LINKED ISSUE #376: [BUG] Return 404 status code when route isn't set
**Description**

Currently requesting a unknown route will result in getting an empty response with a 200 status code.

**Expected Behavior**

We should probably return a 404 status code automatically when this happens.

AntoineRR:
**Description**

This PR fixes #376


sansyrox (approved):
This is very neat! ✨

sansyrox (approved):
LGTM! 👍

PR #383, opened 2023-01-29
chore(ci): fix robyn installation in test CI

AntoineRR:
**Description**

I noticed with my latest PRs that the CI had some issues after my updates... It was downloading the `robyn` package from Pypi instead of installing it from the `whl` built locally. Also the if condition for installing the macos rust target wasn't working.
This fixes all those issues! (had to fix the `watchdog` version for this as we want to have the same version installed with the requirements and required by `robyn`)


sansyrox (approved):
LGTM! 👍

PR #389, opened 2023-02-03
chore: fix requirements files and update packages

AntoineRR:
**Description**

Some requirements were not in the right requirements file. Also I updated the packages.

### DIFF
--- a/.github/workflows/python-CI.yml
+++ b/.github/workflows/python-CI.yml
@@ -28,7 +28,7 @@ jobs:
         run: rustup target add aarch64-apple-darwin
       - name: Setup Rust part of the project
         run: |
-          maturin build -i python --universal2 --out dist --no-sdist
+          maturin build -i python --universal2 --out dist
           pip install --no-index --find-links=dist/ robyn
       - name: Test with pytest
         run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/python-CI.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: rustup target add aarch64-apple-darwin
       - name: Setup Rust part of the project
         run: |
-          maturin build -i python --universal2 --out dist --no-sdist
```
sansyrox, line 31: Why do we want to upload the `sdist` files?

sansyrox: I think the wheels suffice

AntoineRR: This is because of the new version of maturin, the `no-sdist` flag disappeared in v0.13.
See https://www.maturin.rs/migration.html#from-012-to-013:
> maturin 0.13 replaced --no-sdist with the new --sdist option in maturin build command, source distributions are now only built when --sdist is specified.

sansyrox: Oh, perfect then 😄 Thank you!

sansyrox (approved):
LGTM! 👍 

Thank you! :D

Issue #244, opened 2022-07-21
[Feature Request] Add response body in middlewares before and after request


sansyrox:
**Current Behavior**

Now, there are four issues related to this topic:
- It should be possible to access `response["body"]` in the `after_request` middleware.
- There is no response dictionary accessible in the middleware. We only return the request object.
- If we have a `before_request` middleware before a middleware then we can't access the body in the post route.
- And finally, we can't even access the body in `before_request`

Try adding this in the `base_routes`:
```python3
@app.before_request("/post_with_body")
async def hello_before_request(request):
    global callCount
    callCount += 1
    print(request)
    return ""


@app.after_request("/post_with_body")
async def hello_after_request(request):
    global callCount
    callCount += 1
    print(request)
    return ""
```

**Desired Behavior**

Issues are fixed 😅 

---
**Screenshots / Mockups**

**Alternatives**

---

sansyrox:
This is complete!

PR #397, opened 2023-02-07
feat: get rid of intermediate representations of requests and responses

AntoineRR:
**Description**

This PR is a draft to pass a `Request` and `Response` struct to the Python routes (`Response` is used in the "after" middleware).
Currently, I managed to do it, but unfortunately the body of the `Request` and `Response` objects is wrapped in the `ActixBytesWrapper` struct so it is unusable on Python's side.

This will be a breaking change, as the response and request objects you get from Python's struct won't be dictionaries anymore. This would be used like this:

```
@app.get("/sync/param/:id")
def sync_param(request: Request):
    id = request.params["id"]
    return id
```

instead of this:

```
@app.get("/sync/param/:id")
def sync_param(request: dict):
    id = request["params"]["id"]
    return id
```


sansyrox:
@AntoineRR , I think this PR will conflict with #363 .

AntoineRR:
Yes probably, we should merge the other PR prior to this one and I'll do the corrections after. This PR is not ready anyway for now

AntoineRR:
@sansyrox, I have something that works now.
With this PR, we can fix the middlewares as well as #244.
There are some limitations regarding request and response modification in middlewares however:
- I could not add a header by doing `headers["new"] = "header"` but had to create a new dictionary and set it as the headers with `request.headers = new_dict`
- The `str` of the body is stored in `request.body.content`. I would like to make it accessible directly from `request.body` but couldn't do that for now.

If you have any idea on how to fix this, let me know !

AntoineRR:
Hey @sansyrox, I improved the way the body is handled in this PR, using `as_str`, `as_bytes` and `set_body` methods. Setting the headers in the `Response` still isn't the best (maybe we could use a `set_headers` method as well?) but I think the PR is ready for a first review.

sansyrox (commented):
Thanks for the great work @AntoineRR 😄 

I have a few comments at the PR and I will also try to do another review tomorrow 😅

integration_tests/base_routes.py
```diff
@@ -2,7 +2,7 @@
 import pathlib
 
 from robyn import WS, Robyn, jsonify, serve_file, serve_html
-from robyn.robyn import Response
+from robyn.robyn import Request, Response
```
sansyrox, line 5: I have a suggestion. Can we import the `Request` and the `Response` type in the `robyn/__init__.py` and then re-export them?

```
__all__=[..."Response", "Request"]
```

integration_tests/base_routes.py
```diff
@@ -193,14 +263,14 @@ async def async_json_const_get():
 
 
 @app.get("/sync/param/:id")
-def sync_param(request):
-    id = request["params"]["id"]
+def sync_param(request: Request):
```
sansyrox, line 266: Love these api accessors!

src/server.rs
```diff
@@ -368,57 +341,57 @@ async fn index(
...
+                .unwrap_or_else(|e| {
+                    error!("Error while executing before middleware: {e}");
+                    request
+                })
+        }
+        None => request,
+    };
+
+    // Route execution
+    let mut response = if let Some(r) = const_router.get_route(req.method(), req.uri().path()) {
```
sansyrox, line 361: one small suggestion

can we rename `r` to `res` here? It will make slightly better. This is a personal quirk

src/server.rs
```diff
@@ -368,57 +341,57 @@ async fn index(
...
-        MiddlewareRoute::AfterRequest,
-        &middleware_router,
-        req.uri().path(),
-    )
-    .await;
+    response.headers.extend(
+        global_response_headers
+            .iter()
+            .map(|elt| (elt.key().clone(), elt.value().clone())),
+    );
```
sansyrox, line 380: this is niceee

src/types.rs
```diff
@@ -7,26 +7,55 @@ use std::{
...
+            v.as_bytes().to_vec()
+        } else {
+            return Err(PyValueError::new_err(format!(
+                "Could not convert {} specified body to bytes",
+                type_of(content)
+            )));
+        };
+        self.content = Bytes::from(value);
+        Ok(())
+    }
```
sansyrox, line 57: I have a suggestion here. How about using the `#[getter]` and `#[setter]` here(https://pyo3.rs/main/class#object-properties-using-getter-and-setter)

We can keep the base content as `bytes` but then override the getters and setters to string. 

And then only keep an `as_bytes` method when we want to analyze the byte response.

src/types.rs
```diff
@@ -108,17 +145,26 @@ impl FunctionInfo {
...
+    #[pyo3(get, set)]
     pub queries: HashMap<String, String>,
+    #[pyo3(get, set)]
     pub headers: HashMap<String, String>,
     pub method: Method,
+    #[pyo3(get, set)]
     pub params: HashMap<String, String>,
-    pub body: Bytes,
+    #[pyo3(get, set)]
+    pub body: ActixBytesWrapper,
```
sansyrox, line 159: Same for `headers` and `body` here - you can use the getter and setters.

Additionally, we can create a custom class over hashmap for `Headers` and override `__getattr__` and `__setattr__` methods in pyo3. 

Or extend the `dict` class?

src/types.rs
```diff
@@ -131,37 +177,68 @@ impl Request {
             .headers()
             .iter()
             .map(|(k, v)| (k.to_string(), v.to_str().unwrap().to_string()))
+            .chain(
```
sansyrox, line 180: Oh, this is interesting. I wasn't aware of a `chain` method.

src/types.rs
```diff
@@ -131,37 +177,68 @@ impl Request {
...
-    }
 }
 
-#[derive(Debug, Clone)]
 #[pyclass]
+#[derive(Debug, Clone)]
 pub struct Response {
     pub status_code: u16,
     pub response_type: String,
+    #[pyo3(get, set)]
```
sansyrox, line 202: same suggestions of `getter` and `setter` as the `Request` object.

AntoineRR:
Hey @sansyrox, thanks a lot for the great suggestions. I updated the code.
There is just one more thing I am still struggling to do, and it is modifying the headers directly through `request.headers["some_key"] = "some_value"`. This should be doable by changing the type of `headers` to `Py<Pydict>`, but I would like to avoid this as it would make us use the GIL in more places in the code where I feel it shouldn't be required. LMK if you have a suggestion for this!

sansyrox:
Hey @AntoineRR 👋 

Firstly, thank you for the amazing work! 😄 

Regarding your query -> Actually `Py` containers are GIL independent. (https://docs.rs/pyo3/latest/pyo3/struct.Py.html) So, you won't be making extra GIL calls. Additionally, we will be reducing the copies of structs.

AntoineRR:
@sansyrox I think we have to acquire the GIL to perform any operation on the headers when it's wrapped in a `Py`, be it read or write. For example we'll have to do it when adding global headers to `Request` and `Response`.

sansyrox:
> @sansyrox I think we have to acquire the GIL to perform any operation on the headers when it's wrapped in a Py, be it read or write. For example we'll have to do it when adding global headers to Request and Response.

@AntoineRR , I wouldn't worry too much about it as I was going through the codebase and we can optimize this by GIL batching(something that is highly due).

And I belive this will provide a much better interface to work with. And I developer experience is the top most priority imo

sansyrox:
@AntoineRR ,

> What do you mean by "GIL batching"? Does this mean you want to provide a way to perform several operations at once while holding the GIL?

exactly!

> If we don't mind acquiring the GIL in some more places, I can try to wrap the field in a Py.

I believe this will be a temporary drop in performance and we can fix it in the upcoming releases.

AntoineRR:
Ok I could try it like this.
If it doesn't play well, I thought it is also possible to use different structs for using in Rust and Python, and convert one into the other when required. This may be cheaper than acquiring the GIL and more similar to what we have now, i.e. using a `Request` object in Rust, converting it to a `Hashmap` to use in Python and then back to `Request`, except the `Hashmap` would be a new object, a `PyRequest` containing only the field we want to expose. What do you think about this alternative?

sansyrox:
@AntoineRR , I think this approach will be faster than acquiring and releasing GIL on edits. Do you think we should start from this?

AntoineRR:
I think it should be faster too. Let's try this idea then! I will work on this in the following days and keep you updated here.

sansyrox:
@AntoineRR ,sounds great 😊

sansyrox:
Hey @AntoineRR 👋 

I just added this (https://github.com/sansyrox/robyn/pull/443) . It will now eliminate the guessing of performance 😄

AntoineRR:
Nice, thanks! That's good to know :wink:

sansyrox:
Hey @AntoineRR 👋  I hope all's good.

Is this PR up for review?

AntoineRR:
Hey @sansyrox, no I'm still working on this. I'll assign you as a reviewer once the PR is ready

AntoineRR:
Hey @sansyrox, I finally achieved what I wanted! The PR is ready for a first review, or you can wait until I rebased everything correctly (I see there is a new conflict with the latest commit).
With my latest commit, the `request.headers` field can be used as a true python `dict` without any issue and the `request.body` field is either a `str` or `bytes` depending on what you assign to it. There is no more need for the `ActixBytesWrapper` struct too.
The idea is that we have a `Request` and `Response` struct which are pure Rust, and have `ToPyObject` and `FromPyObject` implementations that convert them into and from `PyRequest` and `PyResponse` that are the objects used (instead of a `dict` currently) on Python's side.

integration_tests/test_middlewares.py
```diff
@@ -4,10 +4,11 @@
...
-@pytest.mark.skip(reason="Fix middleware request headers modification")
-def test_middlewares(session):
-    r = get("/")
-    assert "before" in r.headers
-    assert r.headers["before"] == "before_request"
+@pytest.mark.parametrize("function_type", ["sync", "async"])
+def test_middlewares(function_type: str, session):
+    r = get(f"/{function_type}/middlewares")
+    # We do not want the request headers to be in the response
+    assert "before" not in r.headers
```
sansyrox, line 11: Niceee

integration_tests/views/async_view.py
```diff
@@ -1,9 +1,12 @@
...
+
+
 def AsyncView():
     async def get():
         return "Hello, world!"
 
-    async def post(request):
-        body = bytes(request["body"]).decode("utf-8")
+    async def post(request: Request):
+        body = request.body
```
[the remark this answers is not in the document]
sansyrox: we can return `"body": request.body` ?

src/executors/mod.rs
```diff
@@ -1,56 +1,56 @@
...
-pub async fn execute_middleware_function<'a>(
-    request: &Request,
-    function: FunctionInfo,
-) -> Result<HashMap<String, HashMap<String, String>>> {
-    // TODO:
-    // add body in middlewares too
-
+pub async fn execute_middleware_function<T>(input: &T, function: FunctionInfo) -> Result<T>
+where
+    T: for<'a> FromPyObject<'a> + ToPyObject,
```
sansyrox, line 33: what does `for` do here?

sansyrox (commented):
Thank you for the PR @AntoineRR 😄 

I have some initial questions. I will complete the review soon 😄

src/executors/mod.rs
AntoineRR: This was a tricky bug to solve, I didn't know about this either. I asked about it here on the pyo3 repo -> https://github.com/PyO3/pyo3/discussions/3058, and found more info about the syntax here -> https://doc.rust-lang.org/nomicon/hrtb.html

integration_tests/views/async_view.py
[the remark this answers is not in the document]
AntoineRR: I think so I will update it

AntoineRR:
@sansyrox I rebased the branch and applied your recommendations :slightly_smiling_face:

src/lib.rs
```diff
@@ -11,15 +11,17 @@ use shared_socket::SocketHeld;
...
 #[pymodule]
 pub fn robyn(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     // the pymodule class to make the rustPyFunctions available
     m.add_class::<Server>()?;
     m.add_class::<SocketHeld>()?;
     m.add_class::<FunctionInfo>()?;
-    m.add_class::<Response>()?;
+    m.add_class::<PyRequest>()?;
+    m.add_class::<PyResponse>()?;
+    m.add_class::<Url>()?;
```
sansyrox, line 24: @AntoineRR , why do we need to export this?

src/types/function_info.rs
```diff
@@ -0,0 +1,24 @@
...
+#[pymethods]
+impl FunctionInfo {
+    #[new]
+    pub fn new(handler: Py<PyAny>, is_async: bool, number_of_params: u8) -> Self {
+        Self {
+            handler,
+            is_async,
+            number_of_params,
+        }
+    }
```
sansyrox, line 23: Love this segregation.

sansyrox (commented):
Hey @AntoineRR ,

I have one more question here. I will have give it a final review tomorrow. And then we can merge it 😄

src/lib.rs
AntoineRR: This is because we can then use the `Url` struct in python, and do things like `request.url.host` instead of `request.url["host"]`

sansyrox: @AntoineRR , wouldn't that still work if you just export PyRequest ?

src/types/request.rs
```diff
@@ -0,0 +1,104 @@
+use actix_web::{web::Bytes, HttpRequest};
+use dashmap::DashMap;
+use pyo3::{prelude::*, types::PyDict};
+use std::collections::HashMap;
+
+use super::{check_body_type, get_body_from_pyobject, Url};
```
sansyrox, line 6: one minor nit - can you replace `super` with an absolute path? I am not a big fan of relative imports.

src/types/response.rs
```diff
@@ -0,0 +1,130 @@
...
+use actix_web::{HttpRequest, HttpResponse, HttpResponseBuilder, Responder};
+use pyo3::{
+    exceptions::{PyIOError, PyValueError},
+    prelude::*,
+    types::{PyBytes, PyDict, PyString},
+};
+
+use crate::io_helpers::{apply_hashmap_headers, read_file};
+
+use super::{check_body_type, get_body_from_pyobject};
```
sansyrox, line 13: same here.

src/executors/mod.rs
sansyrox: Oh, wow. This will solve a lot of lifetime issues I faced.

src/types/response.rs
```diff
@@ -0,0 +1,130 @@
...
+
+#[derive(Debug, Clone, FromPyObject)]
+pub struct Response {
+    pub status_code: u16,
+    pub response_type: String,
+    pub headers: HashMap<String, String>,
+    #[pyo3(from_py_with = "get_body_from_pyobject")]
```
sansyrox, line 20: One last question - what does `from_py_with` do?

AntoineRR: I use this to use a method when converting the `body` field of the Python object when it is extracted to a `Response`. This way we can ensure it is either a `str` or `bytes` and convert it to a `Vec<u8>`.
You can read more here -> https://pyo3.rs/v0.18.1/conversions/traits#derivefrompyobject-field-attributes

AntoineRR: It is used with the `#[derive(FromPyObject)]` macro

src/types/request.rs
AntoineRR: Done :)

src/lib.rs
AntoineRR: I tested it and it does indeed work without exporting the `Url` struct. It probably makes no sense to allow the user to create `Url` instances from Python so I will remove this export.

src/types/response.rs
sansyrox: Perfect. Thank you! :D

sansyrox (approved):
Great work @AntoineRR 😄 

This PR was intense 🔥

PR #402, opened 2023-02-11
fix: Fix Windows tests

### LINKED ISSUE #372: [BUG] All tests are not passing on windows
**Description**

Currently, we see that all tests are not passing on windows. 

**Expected Behavior**

Make them pass

**Screenshots**

**Environment:**

- Host OS:
- Browser:

---

---

sansyrox:
get conditional global hosts as windows does not allow 0.0.0.0 host

**Description**

This PR fixes #372 

Fixes test on windows


integration_tests/conftest.py
```diff
@@ -33,6 +34,18 @@ def kill_process(process: subprocess.Popen) -> None:
         pass
 
 
+def get_network_host():
+    hostname = socket.gethostname()
+    ip_address = socket.gethostbyname(hostname)
+    # windows return ip address else return 0.0.0.0
+    # windows doesn't support 0.0.0.0
+
+    if platform.system() == "Windows":
```
AntoineRR, line 43: We used something else in other places in the code to get the OS:
```
if sys.platform.startswith("win32")
```
Can we use the same thing in all places ? Probably replace the occurences of the line I wrote above by `if platform.system() == "Windows"` which feels more understandable.

integration_tests/test_base_url.py
```diff
@@ -1,6 +1,19 @@
...
+def get_network_host():
+    hostname = socket.gethostname()
+    ip_address = socket.gethostbyname(hostname)
+    # windows return ip address else return 0.0.0.0
+    # windows doesn't support 0.0.0.0
+
+    if platform.system() == "Windows":
+        return ip_address
+    else:
+        return "0.0.0.0"
```
AntoineRR, line 16: You should probably import the method from `conftest.py` instead of redefining it here. Or maybe move the method to another file and use it both here and in `conftest.py` if you want to keep only the fixtures and methods used only by fixtures in the `conftest.py` file.

AntoineRR (changes requested):
Thanks for fixing the remaining test for Windows @sansyrox ! I have a few suggestions before we can merge this.

integration_tests/conftest.py
sansyrox: Oh great shout! I will change it there too

sansyrox:
These are great suggestions @AntoineRR 😄 

I will implement them tomorrow haven't slept in two days haha 🤣

sansyrox:
@AntoineRR , all done! 😄

Issue #222, opened 2022-07-03
[Feature Request] Add a way to precompute the route existence


sansyrox:
**Current Behavior**

Currently the the string need to go through a list of all the hashmap keys 

**Desired Behavior**

Add a pre-computed set to make the lookup time faster.
---
**Screenshots / Mockups**

**Alternatives**

---

KaviiSuri:
Hey, I'd like to pick this up, where can I get started?

sansyrox:
Route lookup is now backed by a matchit radix-trie (and const routes by a precomputed fast map) rather than iterating over keys, which provides the precomputed/fast route-existence check requested here. Closing as resolved.

PR #405, opened 2023-02-11
improvment: Add route set  to improve 404 responce time.

mikaeelghr:
**Description**

This PR fixes [#222](https://github.com/sansyrox/robyn/issues/222)

### DIFF
--- a/src/routers/mod.rs
+++ b/src/routers/mod.rs
@@ -5,6 +5,7 @@ use crate::types::FunctionInfo;
 
 pub mod const_router;
 pub mod http_router;
+pub mod route_set;
 pub mod middleware_router;
 pub mod types;
 pub mod web_socket_router;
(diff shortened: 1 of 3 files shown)


AntoineRR:
Hi and thank you for taking the time to open this PR! We appreciate your interest and your willingness to contribute :smile: 
However I took a look at the code and I'm not sure we need this new `RouteSet` struct. This adds a new step before actually routing on the correct function to execute it, and would just make 404 response faster and everything else slower.
This issue may not be relevant anymore, but let's see if @sansyrox agrees with me?

mikaeelghr:
Thanks for your comment @AntoineRR. 

I agree with you and I think this change has increased the complexity and also does not improve performance overall.
So if you agree, we can close the PR.

sansyrox:
@mikaeelghr @AntoineRR , I just saw this PR. I agree with @AntoineRR here.

However, we really appreciate the intent here @mikaeelghr 😄

PR #406, opened 2023-02-11
feat: Convert the returned dictionaries directly to json

### LINKED ISSUE #393: [Feature Request] Convert the returned dictionaries directly to json?
**Current Behavior**

Writing `jsonify` to return the json forms of dictionary is cumbersome.

**Desired Behavior**
 Let's return the json response directly when someone is returning a dictionary?

---

**Screenshots / Mockups**

**Alternatives**

---

mikaeelghr:
**Description**
This would be a [breaking change](https://github.com/sansyrox/robyn/issues/393#issuecomment-1423134680).

This PR fixes #393


sansyrox (changes requested):
@mikaeelghr , I was thinking about this PR for a while and I believe I have an approach that will not break anything.

We can have two ways:

1. First, we can try casting dictionary to a Response. 

2. If that doesn't work, we can return it as a json body.


Or, the second way is that we can directly allow dictionaries in the body along with string and bytes.


What do you think?

cc: @AntoineRR

mikaeelghr:
Thanks for your comment @sansyrox.

In my opinion, your first option is pretty but it's a complex behavior and may confuse the developer. For example consider the situation that the developer needs to send Json response with a similar schema to { "status_code", "body", "type", "headers" }.

Your second option is also a good option.

BTW I think maybe with some startup configuration we can use your first method and deprecate casting dictionary to the Response object after a while.

Something like this (of course with better naming):

app = Robyn(__file__, map_dictionary_to_json=True)

AntoineRR:
I second @mikaeelghr on the first option, it is confusing and would prevent returning json with some special keywords. I would go for the following solution :
- returning a dictionary converts it to json, even if it has a `status_code` or other `body` field.
```
return {
    "field": "This is a wonderful json body",
    "body": "This does not prevent returning json"
}
```
- If some special fields of the response must be specified (for example `headers`), user should return a `Response` struct with the info he wants and a `dict` body:
```
return Response(
    headers = { "whatever": "header" },
    body = {
         "field": "This is a wonderful json body",
         "body": "This does not prevent returning json"
    }
)
```
I think at this stage we can afford breaking changes like this as Robyn hasn't reached a stable API yet. This is what `0.x` versions are for after all! :slightly_smiling_face: 

A few pointers I have in mind so far to implement this:
- Detect if we return a dict body or if the body of a `Response` object is a `dict` and add a the following header to the response in these cases:
```
Content-Type: application/json
```
- Change the `ActixBytesWrapper` struct to accept a `PyDict` too

There are probably things missing but I think this is the general idea. Does it seem ok to you two?

sansyrox:
@mikaeelghr @AntoineRR , thank you for you comments.

I agree with both of you. @mikaeelghr , at this stage we can afford breaking changes and we don't need to worry about slow migrations.

@AntoineRR , I basically agree with all the points you said 😄 

And this way we can finally get rid of the `type=text` in the `Response` class.

sansyrox:
This has been implemented.

PR #407, opened 2023-02-11
feat: Add view controllers

### LINKED ISSUE #221: [Feature Request] Add  views/view controllers
**Current Behavior**

Right now, you can only use per route decorators.

**Desired Behavior**

Add the ability to add views. Views will allow the ability to better structure the code. As you can have all the different kind of routes(GET/POST/PUT....) for the same route under a single class.

**High Level Implementation**
- Create a `add_view` class/protcol
- Handle and call the python interface functions.

---
**Screenshots / Mockups**

**Alternatives**

---

mikaeelghr:
**Description**
Add the ability to add views. 

usage:

```
from robyn import Robyn

app = Robyn(__file__)


@app.view("/")
def view():
    def get():
        return "Hello, world!"
    def post():
        return "Hello, world!"

app.start()

```

This PR fixes #221


sansyrox:
Thanks for the PRs @mikaeelghr . It was amazing waking up to these happy surprises 😄 

I will try to review your PRs over the course of the day.

sansyrox:
Hey @mikaeelghr ,

Thanks for the PR. I have a few suggestions for the PR. Can you write some integration tests for your code?

Second instead of creating an `@app.view` decorator, I wanted to add a function like `app.add_view("/", view)`. The rationale behind that is that if a developer wants to use the decorator syntax, they will use the `@get`, `@post`, etc.

But if they want to organize their code in different files or responsibilities, they should use the view syntax and they can include the function from the module.

sansyrox:
@mikaeelghr , I have added a few tests. Everything else looks great!

mikaeelghr:
Thanks for the follow up @sansyrox, I was very busy this week and I couldn't add the tests.

One more thing, based on your comment, shouldn't we delete the `view` function on the `robyn/__init__.py`?

sansyrox:
> Thanks for the follow up @sansyrox, I was very busy this week and I couldn't add the tests.

No worries @mikaeelghr 😄 

I am actually working a lot with views today. And I realized that having this versatility(the `@app.view` decorator) is a nice touch. Great work! 😄 

I will be pushing an update shortly

robyn/__init__.py
```diff
@@ -165,6 +166,40 @@ def terminating_signal_handler(_sig, _frame):
...
                 observer.join()
 
+    def add_view(self, endpoint: str, view: Callable, const: bool = False):
+        """
+        [This is base handler for the view decorators]
+
+        :param endpoint [str]: [endpoint for the route added]
+        :param handler [function]: [represents the function passed as a parent handler for single route with different route types]
+        """
+        http_methods = {"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD", "OPTIONS"}
```
AntoineRR, line 176: Make this a constant defined outside of the method (with the `Final` typing), like this:
```
HTTP_METHODS: Final[set[str]] = ...
```
This will allocate the set only once.

robyn/__init__.py
```diff
@@ -165,6 +166,40 @@ def terminating_signal_handler(_sig, _frame):
                 observer.stop()
                 observer.join()
 
+    def add_view(self, endpoint: str, view: Callable, const: bool = False):
+        """
+        [This is base handler for the view decorators]
+
+        :param endpoint [str]: [endpoint for the route added]
+        :param handler [function]: [represents the function passed as a parent handler for single route with different route types]
```
AntoineRR, line 174: I'm not very familiar with docstrings but I think there is no need for the `[]`? We don't have them in other parts of the code

robyn/__init__.py
```diff
@@ -165,6 +166,40 @@ def terminating_signal_handler(_sig, _frame):
...
+        """
+        http_methods = {"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD", "OPTIONS"}
+
+        def get_functions(view):
+            functions = get_all_nested(view)
+            output = []
+            for name, handler in functions:
+                route_type = name.upper()
+                if route_type in http_methods:
+                    output.append((route_type.upper(), handler))
```
AntoineRR, line 184: `route_type` is already upper case
```suggestion
                    output.append((route_type, handler))
```

integration_tests/views/async_view.py
```diff
@@ -0,0 +1,11 @@
+def AsyncView():
```
AntoineRR, line 1: Shouldn't this be `snake_case` as it is a function? Same for `SyncView`. Now that I think about it, defining views with a class could be cool too?

integration_tests/test_sync_views.py
```diff
@@ -0,0 +1,25 @@
+from integration_tests.helpers.http_methods_helpers import get, post
+
+
+def test_get_async_view(session):
+    r = get("/sync_view")
+    assert r.status_code == 200
```
AntoineRR, line 6: The status code is already tested in the `get` method and other http methods. You can remove this line in the added tests.

integration_tests/test_base_url.py
```diff
@@ -2,7 +2,7 @@
 
 import requests
 
-from helpers.network_helpers import get_network_host
+from integration_tests.helpers.network_helpers import get_network_host
```
AntoineRR, line 5: This change is making the tests fail AFAIK (happened to me too)

integration_tests/base_routes.py
```diff
@@ -493,4 +518,6 @@ async def async_raise():
         index_file="index.html",
     )
     app.startup_handler(startup_handler)
+    app.add_view("/sync_view", SyncView)
+    app.add_view("/async_view", AsyncView)
```
AntoineRR, line 522: Can we use the same convention as in the other routes for the tests? i.e. `/sync/view` instead of `/sync_view`. Maybe `/sync/view/decorator` for the decorated routes. This helps keep things consistent and organized :slightly_smiling_face:

AntoineRR (changes requested):
Great job! :smile: I added a few small suggestions. One last here: I would put all tests related to views in the same file, be they sync or async. This is what's done for the other features so I think it makes more sense.

robyn/__init__.py
sansyrox: Yep true. :D

robyn/__init__.py
sansyrox: Right, perfect.

integration_tests/views/async_view.py
sansyrox: > Shouldn't this be snake_case as it is a function? 

Actually, no. I wanted the views to be easily identifiable. Having class views is something up for the future, but implementing this as functional components was more difficult and interesting 🤣

integration_tests/test_base_url.py
sansyrox: Thanks @AntoineRR 😄 I was losing my hair over this 🤣

integration_tests/base_routes.py
sansyrox: Great shout. Thank you! 😄

integration_tests/test_sync_views.py
sansyrox: Done!

robyn/__init__.py
sansyrox: @AntoineRR , unfortunately the `final` keyword was introduced in Python 3.8(https://peps.python.org/pep-0591/) and we still support 3.7 and above.

Also, the reallocation is only done during the add time, which will affect the startup but will have no performance impact on serving responses. So, I don't think we need a global variable just for this.


What do you reckon?

AntoineRR: Ok, we can leave it like this, this is not very important :)

sansyrox:
@AntoineRR , everything is resolved now. Thank you for the review 😄 

Thanks for the great work again @mikaeelghr 😄

AntoineRR (changes requested):
Thanks for the fixes! One last thing @sansyrox, can't we merge the `integration_tests/views/async_view.py` and `integration_tests/views/test_view.py` files together? Or at least rename `test_view.py` to `sync_view.py`?

sansyrox:
@AntoineRR , that was already done. Try refreshing maybe? 🤔

sansyrox:
@AntoineRR , 

> can't we merge the integration_tests/views/async_view.py and integration_tests/views/test_view.py files together? 

We can but I wanted to suggest one view per file(like react) for better organization of the code.

>  Or at least rename test_view.py to sync_view.py?

Done! 😄

AntoineRR (approved):

domingues:
@mikaeelghr @sansyrox Genuine question, why not use a simple `class` vs doing code introspection to extract the inner functions? Can't see in each way this is a better approach.

sansyrox:
Hi @domingues 👋 

We don't want to choose function base approach over class based approach. I just like to structure my code in closures sometimes and not having the ability to do that anywhere else was annoying .

We just haven't got around to implementing class-based views as of yet. 😅

PR #409, opened 2023-02-12
chore: use pyproject.toml to generate requirements files

AntoineRR:
**Description**

Currently, the dependencies are specified both in `pyproject.toml` and in the requirements files.
With this PR, I propose to rely on the `pyproject.toml` file to specify the dependencies of the project, and generate the requirements files using [pip-tools](https://pypi.org/project/pip-tools/), via the `pip-compile` command.


sansyrox:
Hey @AntoineRR , 👋 

Thanks for the PR but I have a different suggestion.

While it is true that maintaining `requirements` and `dev-requirements` is a hassle. But I believe that we should be working on this solution (https://github.com/sansyrox/robyn/issues/398) instead of working on an intermediate step.

AntoineRR:
You're right, I never used poetry, I thought it would rely on `requirements.txt` files but it seems it just uses the `pyproject.toml` file? I will take a look at this if I have some time. In the meantime, the `pyproject.toml` file from this PR can probably be used as a starting point for building the one that poetry will rely on, I think it will be similar.

sansyrox:
@AntoineRR , absolutely.

But the pyproject files for poetry are slightly different.

Like this

```
# pyproject.toml

[tool.poetry]
name = "rp-poetry"
version = "0.1.0"
description = ""
authors = ["Philipp <philipp@realpython.com>"]

[tool.poetry.dependencies]
python = "^3.9"

[tool.poetry.dev-dependencies]
pytest = "^5.2"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```


The dependencies are mentioned in `[tool.poetry.dependencies]`

sansyrox:
Closing this as we have now migrated to poetry.

PR #418, opened 2023-02-17
docs: improve readme

AntoineRR:
**Description**

This is a proposal for a new readme with updated information.
Also the readme was  duplicated in the `docs` folder. I kept only the one in the `docs` folder as it should appear on the repo homepage too as stated in [this](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes) documentation. This will make us maintain only one readme instead of trying to keep both in sync.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,16 +3,15 @@
 # Robyn
 
 [![Twitter](https://badgen.net/badge/icon/twitter?icon=twitter&label)](https://twitter.com/robyn_oss)
-[![Gitter](https://badges.gitter.im/robyn_/community.svg)](https://gitter.im/robyn_/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 [![Downloads](https://static.pepy.tech/personalized-badge/robyn?period=total&units=international_system&left_color=grey&right_color=blue&left_text=Downloads)](https://pepy.tech/project/robyn)
-[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![GitHub tag](https://img.shields.io/github/tag/sansyrox/robyn?include_prereleases=&sort=semver&color=black)](https://github.com/sansyrox/robyn/releases/)
 [![License](https://img.shields.io/badge/License-BSD_2.0-black)](#license)
-[![Discord](https://img.shields.io/discord/999782964143603713?label=discord&logo=discord&logoColor=white&style=for-the-badge&color=blue)](https://discord.gg/qKF5sSnC)
+![Python](https://img.shields.io/badge/Support-Version%20%E2%89%A5%203.7-brightgreen)
 
 [![view - Documentation](https://img.shields.io/badge/view-Documentation-blue?style=for-the-badge)](https://sansyrox.github.io/robyn/#/)
+[![Discord](https://img.shields.io/discord/999782964143603713?label=discord&logo=discord&logoColor=white&style=for-the-badge&color=blue)](https://discord.gg/rkERZ5eNU8)
 
-Robyn is an async Python backend server with a runtime written in Rust, btw.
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -3,16 +3,17 @@
...
-[![Discord](https://img.shields.io/discord/999782964143603713?label=discord&logo=discord&logoColor=white&style=for-the-badge&color=blue)](https://discord.gg/qKF5sSnC)
+![Python](https://img.shields.io/badge/Support-Version%20%E2%89%A5%203.7-brightgreen)
 
 [![view - Documentation](https://img.shields.io/badge/view-Documentation-blue?style=for-the-badge)](https://sansyrox.github.io/robyn/#/)
+[![Discord](https://img.shields.io/discord/999782964143603713?label=discord&logo=discord&logoColor=white&style=for-the-badge&color=blue)](https://discord.gg/rkERZ5eNU8)
 
-Robyn is an async Python backend server with a runtime written in Rust, btw.
+Robyn is a fast async Python web framework coupled with a web server written in Rust.
+
+Check out the talk at **PyCon Sweden 2021** about [Robyn: An async python web framework with a Rust runtime](https://www.youtube.com/watch?v=DK9teAs72Do)
```
sansyrox, line 16: I think we should have a separate talk section. But we can address this later. 😄

sansyrox (commented):
All looks great! But, would this not replace the README preview on the main GitHub preview?

docs/README.md
sansyrox: Or, rather , we can just link it here - https://github.com/sansyrox/robyn/blob/main/docs/community-resources.md

AntoineRR: That's a good point, updated it

AntoineRR:
The `docs/README.md` file should indeed appear as a preview on the repo main page.
I've been thinking about this though, and wonder if we shouldn't have a completely different README.md for the docs (that will appear in the home page of the docs)? Not sure it makes sense to have the readme both in the repo and on the docs home page

sansyrox:
> I've been thinking about this though, and wonder if we shouldn't have a completely different README.md for the docs (that will appear in the home page of the docs)? 

We can do that but what will be different in both of them?

AntoineRR:
I updated the `docs/README.md` with a more documentation centered approach, LMK if that suits you.
Also added the Robyn logo as the docs website icon!

docs/README.md
```diff
@@ -1,101 +1,20 @@
...
-
-If you're feeling curious. You can take a look at a more detailed architecture [here](https://github.com/sansyrox/robyn/blob/main/docs/architecture.md).
-
-## ⚙️ To Develop Locally
-
-1. Install the development dependencies: `pip install -r dev-requirements.txt`
-
-1. Install the pre-commit git hooks: `pre-commit install`
-
-1. Add more routes in the `integration_tests/base_routes.py` file(if you like).
```
sansyrox, line 73: @AntoineRR , I think we should not delete this. If people will want to have a quick start documentation summary, they will have to go to the previous page. But if we keep it here, they won't have to move away from the documentation.

I believe that we need a proper custom documentation website revamp but I think we can keep this for now.

What say? @AntoineRR

docs/index.html
```diff
@@ -12,6 +12,7 @@
       name="viewport"
       content="width=device-width, initial-scale=1.0, minimum-scale=1.0"
     />
+    <link rel="icon" href="https://user-images.githubusercontent.com/29942790/140995889-5d91dcff-3aa7-4cfb-8a90-2cddf1337dca.png" />
```
sansyrox, line 15: Niceee

docs/README.md
AntoineRR: Ok that makes sense, let's keep this readme as a copy of the one on the repo for now then. We'll see later how to deal with this properly.

sansyrox (approved):
Great work @AntoineRR . LGTM! 👍

PR #421, opened 2023-02-19
better way to compare type

jmishra01:
Instead of checking type of variable using type method, we can use `isinstance` method to check.

This PR fixes #


sansyrox (approved):
Thank you for the PR @jmishra01 . LGTM 😄 ❇️

PR #431, opened 2023-02-25
feat: implement the segregation of request and response headers

sansyrox:
**Description**

This PR fixes #

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -61,6 +61,7 @@ async def hello_before_request(request, response):
     request["headers"]["before"] = "before_request"
     response["headers"]["before"] = "before_request"
 
+
 @app.after_request("/")
 async def hello_after_request(request, response):
     request["headers"]["after"] = "after_request"
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -60,18 +60,19 @@ def shutdown_handler():
 async def hello_before_request(request, response):
     request["headers"]["before"] = "before_request"
     response["headers"]["before"] = "before_request"
-    return request
+    return ""
 
 
 @app.after_request("/")
-async def hello_after_request(request):
+async def hello_after_request(request, response):
     request["headers"]["after"] = "after_request"
-    return request
+    response["headers"]["after"] = "after_request"
+    return ""
 
 
 @app.get("/")
 async def middlewares(request, response):
-    return ""
+    return "Hello world"
 
 
 # ===== Routes =====
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -60,14 +60,11 @@ def shutdown_handler():
 async def hello_before_request(request, response):
     request["headers"]["before"] = "before_request"
     response["headers"]["before"] = "before_request"
-    return ""
-
 
 @app.after_request("/")
 async def hello_after_request(request, response):
     request["headers"]["after"] = "after_request"
     response["headers"]["after"] = "after_request"
-    return ""
 
 
 @app.get("/")
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -257,7 +257,7 @@ async def file_download_async():
 
 
 @app.get("/sync/queries")
-def sync_queries(request):
+def sync_queries(request, response):
     query_data = request["queries"]
     return jsonify(query_data)
 
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -57,8 +57,9 @@ def shutdown_handler():
 
 
 @app.before_request("/")
-async def hello_before_request(request):
+async def hello_before_request(request, response):
     request["headers"]["before"] = "before_request"
+    response["headers"]["before"] = "before_request"
     return request
 
 
@@ -69,7 +70,7 @@ async def hello_after_request(request):
 
 
 @app.get("/")
-async def middlewares():
+async def middlewares(request, response):
     return ""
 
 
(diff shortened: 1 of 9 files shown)


src/executors/mod.rs
```diff
@@ -14,51 +14,68 @@ use pyo3::prelude::*;
...
-pub async fn execute_http_function(request: &Request, function: FunctionInfo) -> Result<Response> {
+pub async fn execute_http_function(
+    request: &mut Request,
+    response: &mut Response,
+    function: FunctionInfo,
+) -> Result<Response> {
     if function.is_async {
         let output = Python::with_gil(|py| {
-            let function_output = get_function_output(&function, py, request);
+            let function_output = get_function_output(&function, py, request, response);
```
sansyrox, line 78: should not require a response here

src/executors/mod.rs
```diff
@@ -14,51 +14,68 @@ use pyo3::prelude::*;
...
+        Python::with_gil(|py| get_function_output(&function, py, request, response)?.extract())
             .context("Failed to execute handler function")
     }
 }
 
 #[inline]
-pub async fn execute_http_function(request: &Request, function: FunctionInfo) -> Result<Response> {
+pub async fn execute_http_function(
+    request: &mut Request,
+    response: &mut Response,
```
sansyrox, line 73: same here

src/executors/mod.rs
```diff
@@ -68,7 +85,7 @@ pub async fn execute_http_function(request: &Request, function: FunctionInfo) ->
         })
     } else {
         Python::with_gil(|py| -> Result<Response> {
-            let output = get_function_output(&function, py, request)?;
+            let output = get_function_output(&function, py, request, response)?;
```
sansyrox, line 88: same here

src/routers/const_router.rs
```diff
@@ -43,9 +43,10 @@ impl Router<Response, Method> for ConstRouter {
             event_loop.context("Event loop must be provided to add a route to the const router")?;
 
         pyo3_asyncio::tokio::run_until_complete(event_loop, async move {
-            let output = execute_http_function(&Request::default(), function)
-                .await
-                .unwrap();
+            let output =
+                execute_http_function(&mut Request::default(), &mut Response::default(), function)
```
sansyrox, line 47: should maybe pass request object here and no response

src/server.rs
```diff
@@ -334,27 +336,95 @@ impl Default for Server {
...
+    // params
+
+    if modified_response.contains_key("headers") {
+        Python::with_gil(|py| {
+            let headers = modified_response.get("headers").unwrap();
+            let headers = headers.extract::<HashMap<String, String>>(py).unwrap();
+            for (key, value) in headers {
+                response.headers.insert(key, value);
+            }
+        });
```
sansyrox, line 388: dk about these

integration_tests/base_routes.py
```diff
@@ -57,20 +57,22 @@ def shutdown_handler():
 
 
 @app.before_request("/")
-async def hello_before_request(request):
+async def hello_before_request(request, response):
     request["headers"]["before"] = "before_request"
-    return request
+    response["headers"]["before"] = "before_request"
+    return ""
```
sansyrox, line 63: find a way that this is not needed

sansyrox:
Hey @AntoineRR 👋 

I hope all's well. Would it be possible for you to do a strict review of this PR? I was heavily delirious while implementing this 😅

AntoineRR:
Hi @sansyrox, I had less time for Robyn recently but all's good thanks! Sure, I'll take a look at this soon and see if you were that delirious then :stuck_out_tongue:

AntoineRR (commented):
I took a look at your changes, and I think you aim for the same goal as mine from PR #397, although we took very different approach. Both aim at adding body and other things in middlewares and fix middlewares (I think I only miss the request/response headers separation in my PR). I will try to highlight the design choices we made to compare them here :slightly_smiling_face: 

### In my PR
- We create a `Request` object from actix `HttpRequest`
- We pass the `Request` object to the "before" middleware, and get a new `Request` object as a result
- We pass the `Request` object to the route function, and get a `Response` object as a result
- We pass the `Response` object to the "after" middleware, and get a new `Response` object as a result
- This `Response` object is returned as the result of the route call.

The `Request` and `Response` object are used directly from Python, so no need to convert them to `Hashmap`s.

### In this PR
- The `Response` and `Request` objects are both created in the beginning
- We do the following for "before" middlewares, route function, and "after" middleware:
  - Convert the `Request` and `Response` objects to `Hashmap`s
  - Call the Python function with the two `Hashmap`s
  - Convert the `Hashmap`s back to `Request` and `Response`

### My opinion
IMO, we should provide an access for either the `Request` or `Response` object depending on the case, and not both at the same time as it doesn't really make sense to give access to the `Response` before it was generated by the route function.
I think my approach is lighter (it should be faster as we do not convert between `Request`, `Response` and `Hashmap`s) and less confusing (because we give access to only the relevant objects in the route and middlewares). However I still need to work on some aspects for my PR to be ready.
Of course this is just my opinion and do not hesitate to tell me if you disagree! :smile:

sansyrox:
> Hi @sansyrox, I had less time for Robyn recently but all's good thanks! Sure, I'll take a look at this soon and see if you were that delirious then 😛

Haha, I really appreciate it 😄 

> and I think you aim for the same goal as mine from PR 

Firstly, apologies for working on the same thing( as you but there was no other way of fixing the segregation 😅 . I will give attribution to you in one commit so github considers this PR as co-authored.


I just saw your PR, and I agree with your approach. Should I fetch the changes off of you PR and then work on it? I believe it will make things faster in iteration

AntoineRR:
No worries, maybe my PR wasn't very explicit ^^
I think it will be easier if we both work on my PR if we keep its architecture ?
I will rebase it this evening  to solve the merge conflicts

sansyrox:
@AntoineRR , sounds good to me 😃

sansyrox:
Not needed anymore. Stale

PR #441, opened 2023-03-10
feat: expose request/connection info

r3b-fish:
**Description**
Implemented access of request information: method, path, host:port and protocol.
In generally, it needs to implement more complicated logic in middlewares.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -15,8 +15,8 @@ use actix_web::{http::Method, HttpRequest};
 use anyhow::Result;
 use dashmap::DashMap;
 use pyo3::exceptions::PyValueError;
-use pyo3::types::{PyBytes, PyString};
-use pyo3::{exceptions, prelude::*};
+use pyo3::types::{PyBytes, PyDict, PyString};
+use pyo3::{exceptions, intern, prelude::*};
 
 use crate::io_helpers::read_file;
 
@@ -109,13 +109,38 @@ impl FunctionInfo {
 }
 
 #[derive(Default)]
+pub struct Url {
+    pub scheme: String,
+    pub host: String,
+    pub path: String,
+}
+
+impl Url {
+    fn new(scheme: &str, host: &str, path: &str) -> Self {
+        Self {
+            scheme: scheme.to_string(),
+            host: host.to_string(),
+            path: path.to_string(),
+        }
+    }
+
+    pub fn to_object(&self, py: Python<'_>) -> PyResult<PyObject> {
+        let dict = PyDict::new(py);
+        dict.set_item(intern!(py, "scheme"), self.scheme.as_str())?;
+        dict.set_item(intern!(py, "host"), self.host.as_str())?;
+        dict.set_item(intern!(py, "path"), self.path.as_str())?;
+        Ok(dict.into_py(py))
+    }
+}
+
+#[derive(Default)]
 pub struct Request {
     pub queries: HashMap<String, String>,
     pub headers: HashMap<String, String>,
     pub method: Method,
     pub params: HashMap<String, String>,
     pub body: Bytes,
-    pub url: HashMap<String, String>,
+    pub url: Url,
 }
 
 impl Request {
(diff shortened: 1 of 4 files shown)


sansyrox:
Hi @r3b-fish 👋


Thank you for the PR 😊 the pr looks nice from the first glance but I will give it a deeper look tonight 😊

src/types.rs
```diff
@@ -139,15 +142,28 @@ impl Request {
...
         result.insert("body", self.body.to_object(py));
+        let method = PyString::new(py, self.method.as_ref()).to_object(py);
+        result.insert("method", method);
+        let path = PyString::new(py, self.path.as_ref()).to_object(py);
+        result.insert("path", path);
+        let host = PyString::new(py, self.host.as_ref()).to_object(py);
+        result.insert("host", host);
+        let protocol = PyString::new(py, self.protocol.as_ref()).to_object(py);
+        result.insert("protocol", protocol);
+
```
sansyrox, line 166: One suggestion 

We can optimise this dictionary creation using `intern` strings.

Also, can you please rebase with the latest main to get the benchmarks?

r3b-fish: I've updated code. Can you check it, pls.

src/types.rs
```diff
@@ -139,15 +142,27 @@ impl Request {
...
         result.insert("queries", self.queries.to_object(py));
         result.insert("headers", self.headers.to_object(py));
         result.insert("body", self.body.to_object(py));
+
+        let mut url = HashMap::new();
+        url.insert("path", self.path.to_string());
+        url.insert("host", self.host.to_string());
+        url.insert("scheme", self.scheme.to_string());
+        result.insert("url", url.to_object(py));
+
```
sansyrox, line 165: @r3b-fish , all looks good to me. But what is the need to create a new hasmap here? Why not return all the params as is?

r3b-fish: FastAPI uses [the same structure](https://www.starlette.io/requests/#url). I think we may implement it from a popular solution.

r3b-fish: I may rewrite if you want, no problem, it doesn't matter for me

sansyrox: No, exposing a URL object looks like a very nice interface. Thank you for this idea :smile:

src/types.rs
```diff
@@ -139,15 +142,27 @@ impl Request {
             method: req.method().clone(),
             params: HashMap::new(),
             body,
+            path: req.path().to_string(),
+            host: req.connection_info().host().to_string(),
+            scheme: req.connection_info().scheme().to_string(),
```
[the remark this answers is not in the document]
r3b-fish: Thanks for your comment! Now, this looks awesome ;)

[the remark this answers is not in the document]
sansyrox: I am glad you liked the comments @r3b-fish 😄 .

I have one more follow up comment before we can merge this. We can add one more optimization and use the `intern!` macro from py03.

So the code will look something like this:

```rust
struct URL {
    method: String;
    path: String;
...
}

impl URL {
    fn new(method: &str, path: &str, ...) -> Self {
        Self()        
    }

    pub fn to_object(&self, py: Python<'_>) -> PyResult<PyObject> {
        let dict = PyDict::new(py);
        let method = self.method
        dict.set_item(intern!(py, "method"), self.method)?;
        dict.set_item(intern!(py, "path"), self.method)?;
        Ok(dict.into())
            
}
}

```


Using this will help us create this string without reallocations.

You can read more about it here - https://pyo3.rs/main/doc/pyo3/macro.intern.html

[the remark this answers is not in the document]
r3b-fish: Thanks for you help. I'm not sure about the "method". Can you check my code again, please.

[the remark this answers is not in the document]
sansyrox: @r3b-fish , `method` was a typo from my end. My bad!


Great work btw! :smile:

sansyrox (approved):
Great work! :fire: 

LGTM :smile:

PR #444, opened 2023-03-12
chore: update rust dependencies

AntoineRR:
**Description**

Update the Rust dependencies. Two notes:
- There is a new deprecation in pyo3. `Option` args should not appear before non `Option` args in function parameters anymore. This concerns the Directory api.
- The `dashmap` crate has been updated to use `parking_lot` so no need to use a forked repo anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,7 +19,7 @@ dependencies = [
  "futures-util",
  "log",
  "once_cell",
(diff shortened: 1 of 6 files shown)


sansyrox (approved):
Great work! Thank you @AntoineRR 😄

PR #446, opened 2023-03-16
fix(dev): fix hot reloading with dev flag

### LINKED ISSUE #439: Can't access new or updated route while on dev option
### Bug Description

Dev option for hot reloading is not working to recompile python code into rust object

### Steps to Reproduce

1. Start simple app.py,  `python app.py --dev`

```
from robyn import Robyn

app = Robyn(__file__)


@app.get("/")
async def hello(request):
    return "Hello World"


app.start(port=8080)
```
2. Update return value of hello function, or add new route
DEBUG:fsevents:NativeEvent(path="/pp.py", inode=30320742, flags=11400, id=285275140): is_file, is_inode_meta_mod, is_modified
DEBUG:fsevents:NativeEvent(path="web_api/app.py~", inode=30678452, flags=15700, id=285275143): is_coalesced, is_created, is_file, is_inode_meta_mod, is_modified, is_owner_change, is_removed
DEBUG:fsevents:queue_event <FileModifiedEvent: event_type=modified, src_path='web_api/app.py', is_directory=False>
DEBUG:fsevents:queue_event <FileCreatedEvent: event_type=created, src_path='web_api/app.py~', is_directory=False>
DEBUG:fsevents:queue_event <DirModifiedEvent: event_type=modified, src_path='web_api', is_directory=True>
DEBUG:fsevents:queue_event <FileModifiedEvent: event_type=modified, src_path='web_api/app.py~', is_directory=False>
DEBUG:fsevents:queue_event <FileDeletedEvent: event_type=deleted, src_path='web_api/app.py~', is_directory=False>
DEBUG:fsevents:queue_event <DirModifiedEvent: event_type=modified, src_path='web_api', is_directory=True>

3. But when rehit GET root path, still return old value or new route is not available

### Your operating system

on macbook intel, mac os monterey

3.10.9

### Your Robyn version

0.25.0

### Additional Info

_No response_

AntoineRR:
**Description**

This PR fixes #439

The idea behind this PR is, when the dev flag is used, to:
- Use the process run by the user to setup an observer and launch the server as a child process
- When the observer detects a change in the directory, destroy all child processes and restart a new server
- When a SIGINT is received, destroy both every child processes and the observer


robyn/reloader.py
```diff
@@ -0,0 +1,86 @@
...
+    finally:
+        observer.stop()
+        observer.join()
+
+
+class EventHandler(FileSystemEventHandler):
+    def __init__(self, file_path: str) -> None:
+        self.file_path = file_path
+
+        self.processes = []
```
sansyrox, line 49: Thanks for the PR @AntoineRR 😄 

Do we still need this `self.processes` data member?

AntoineRR: Thanks, I forgot to remove it. I fixed it now :)

robyn/reloader.py
```diff
@@ -0,0 +1,84 @@
...
+            process.kill()
+            process.wait()  # Required to avoid zombies
+
+    def reload(self):
+        self.stop_server()
+
+        new_env = os.environ.copy()
+        new_env[
+            "IS_RELOADER_SETUP"
+        ] = "True"  # This is used to check if a reloader is already running
```
sansyrox, line 62: Hey @AntoineRR ,

All looks great. I just have one final question - does `IS_RELOADER_SETUP` env variable indicate that it is a subprocess generated from a change ?

AntoineRR: This env variable is used to check if the reloader is running. You can think of it as a check that the subprocess was launched after a file modification by the reloader yes

sansyrox: @AntoineRR , but why do we need an environment variable for this? Just for sharing data between processes?

We can use `multiprocessing.Event` or `Lock` for the same - https://docs.python.org/3/library/multiprocessing.html?highlight=event#multiprocessing.Event

AntoineRR: I am not sure we can use this to share information between processes, but only between threads from the same process? I based my implementation on an old commit from sanic, but maybe they improved the way it is done since -> https://github.com/sanic-org/sanic/commit/52c2a8484e6aa5fa13aaade49e1f2597dd006e15.
If you have an alternative in mind please let me know, as I don't see how to use a `Lock` or `Event` in this situation.

robyn/__init__.py
```diff
@@ -30,15 +27,23 @@ def __init__(self, file_object: str) -> None:
         self.file_path = file_object
         self.directory_path = directory_path
         self.config = Config()
```
sansyrox, line 29: Hey @AntoineRR ,

I have an unrelated suggestion to the PR. But this will help us with writing integration tests for the dev server.

We can pass the config as a part of the Robyn initializer. 

```
def __init__(self, file_object: str, config: Config = Config()):
...
  self.config = config
```

This way we will be able to pass in a dev flag and custom flag while writing test code.

robyn/__init__.py
```diff
@@ -30,15 +27,23 @@ def __init__(self, file_object: str) -> None:
...
+
+        load_vars(project_root=directory_path)
+        logging.basicConfig(level=self.config.log_level)
+
+        # If we are in dev mode, we need to setup the reloader
+        # This process will be used by the watchdog observer while running the actual server as children processes
+        if self.config.dev and not os.environ.get("IS_RELOADER_SETUP", False):
```
sansyrox, line 36: @AntoineRR , 1) you can initialize the event here and pass is to `setup_reloader`

robyn/reloader.py
```diff
@@ -0,0 +1,84 @@
...
+import psutil
+
+from watchdog.events import FileSystemEventHandler
+from watchdog.observers import Observer
+
+from robyn.logger import Colors, logger
+
+
+def setup_reloader(directory_path: str, file_path: str):
+    event_handler = EventHandler(file_path)
```
sansyrox, line 15: you can then pass the `Event/Lock` to event handler init

robyn/reloader.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def stop_server(self):
+        for process in psutil.Process().children(recursive=True):
+            process.kill()
+            process.wait()  # Required to avoid zombies
+
+    def reload(self):
+        self.stop_server()
+
+        new_env = os.environ.copy()
```
sansyrox, line 59: and then check here.

Lmk, if this makes sense. I will be happy to push a commit to this PR otherwise 😅

sansyrox (commented):
Hey @AntoineRR 👋 

Thanks for the PR 😄 

I have some inline comments

robyn/reloader.py
sansyrox: Also, if `multiprocessing.Event` doesn't work , we can have a look at this (https://docs.python.org/3/library/multiprocessing.html?highlight=event#multiprocessing.BoundedSemaphore) .

I haven't used this personally, but can give a deeper dive if need be.

AntoineRR: @sansyrox I'm really not convinced we can use those, with your method we would just create a new instance of `Event` each time we create a new process after reloading. I think we should merge the PR to have a first working solution, and maybe we could change things later? You could open a new PR if you manage to get something that works without env variables.

robyn/__init__.py
AntoineRR: That's a good idea, I added this :smile:

sansyrox (approved):
Sorry for the delays. I missed some important details here.

Great work as usual 🔥 🔥

PR #451, opened 2023-03-25
docs: Add docs for v0.26.0

### LINKED ISSUE #454: Add documentation for `robyn.env` file
There is no documentation regarding `robyn.env` file.

Need to add it

sansyrox:
**Description**

Fixes #454 

Add documentation for the latest release

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -37,7 +37,7 @@ async def h(request):
 ```python
 @app.post("/post")
 async def postreq(request):
-    return bytearray(request["body"]).decode("utf-8")
+    return request.body
 ```
 
 #### PUT Request
@@ -45,7 +45,7 @@ async def postreq(request):
 ```python
 @app.put("/put")
 async def postreq(request):
-    return bytearray(request["body"]).decode("utf-8")
+    return request.body
 ```
 
 #### PATCH Request
@@ -53,7 +53,7 @@ async def postreq(request):
 ```python
 @app.patch("/patch")
 async def postreq(request):
-    return bytearray(request["body"]).decode("utf-8")
+    return request.body
 ```
 
 #### DELETE Request
@@ -61,7 +61,7 @@ async def postreq(request):
 ```python
 @app.delete("/delete")
 async def postreq(request):
-    return bytearray(request["body"]).decode("utf-8")
+    return request.body
 ```
 
 #### Directory Serving
@@ -131,6 +131,20 @@ from robyn.robyn import Response
 @app.get("/response")
 async def response(request):
     return Response(status_code=200, headers={}, body="OK")
+```
+
+#### Status Codes
+
+Robyn provides `StatusCodes` if you want to return type safe Status Responses.
+
+```python
+
+from robyn import StatusCodes
+
+
+@app.get("/response")
+async def response(request):
+    return Response(status_code=StatusCodes.HTTP_200_OK.value, headers={}, body="OK")
 ```
 
 #### Returning a byte response
@@ -208,6 +222,16 @@ async def response_headers():
     return {
         "headers": {"Header": "header_value"},
     }
+```
+
+
+Additionally, you can access headers for per route.
+
+```python
+@app.get("/test-headers")
+def sync_before_request(request: Request):
+    request.headers["test"] = "we are modifying the request headers in the middle of the request!"
+    print(rquest)
 ```
 
 ## Query Params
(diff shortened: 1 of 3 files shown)


docs/features.md
```diff
@@ -210,6 +224,16 @@ async def response_headers():
...
 
+
+Additionally, you can access headers for per route.
+
+```python
+@app.get("/test-headers")
+def sync_before_request(request: Request):
+    request.headers["test"] = "test-header"
+    return request
+```
```
AntoineRR, line 235: This will not work I think. In a route, we want to return a Response or the body of a response. If we want to change the headers of a request object, it is only relevant in a before middleware.

docs/features.md
```diff
@@ -327,11 +351,13 @@ You can use both sync and async functions for middlewares!
 ```python
 @app.before_request("/")
 async def hello_before_request(request):
```
AntoineRR, line 353:
```suggestion
async def hello_before_request(request: Request):
```

docs/features.md
```diff
@@ -327,11 +351,13 @@ You can use both sync and async functions for middlewares!
...
+    request.headers["before"] = "sync_before_request"
     print(request)
 
 
 @app.after_request("/")
 def hello_after_request(request):
+    request.headers["after"] = "sync_after_request"
     print(request)
```
AntoineRR, line 361:
```suggestion
def hello_after_request(response: Response):
    response.headers["after"] = "sync_after_request"
    print(response)
```

AntoineRR (changes requested):
Hey @sansyrox , thanks for updating the docs. I left a few comments.
Also I think there are other parts in the docs where we access `body`, `params` and `queries`. Those should change as well. I found some others in the `features.md` file, but also `graphql-integration.md`. Can you update them too please?

docs/features.md
```diff
@@ -133,6 +133,20 @@ async def response(request):
...
+Robyn provides `StatusCodes` if you want to return type safe Status Responses.
+
+```python
+
+from robyn import StatusCodes
+
+
+@app.get("/response")
+async def response(request):
+    return Response(status_code=StatusCodes.HTTP_200_OK, headers={}, body="OK")
```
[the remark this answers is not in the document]
sansyrox: Oh, right. My bad. I will update it.

docs/features.md
sansyrox: > This will not work I think. In a route, we want to return a Response or the body of a response. If we want to change the headers of a request object, it is only relevant in a before middleware.


@AntoineRR , what you mean by this will not work? Yep, look at it, it is a bad example. But is working from looking at the code 😅

docs/features.md
sansyrox: Ooh , playing typed vs untyped Python, are we? 🤣 . I will update it.

sansyrox:
Thanks for the review @AntoineRR 😄 

I just have a question.

docs/features.md
sansyrox: Are you suggesting that we should change the example here?

AntoineRR: @sansyrox I just tested this and it does not crash, although the response we get from the server is something like `<builtins.Request object at 0x7f63d56f2330>` (which is the result of `str(request)`.
Also, there really is no point currently in modifying the request headers in a route, because we will not use the request in the rest of the code afterwards. Route methods take `Request` as an input and returns `Response`. The only part were it makes sense to modify request headers is in a before middleware, because the modified `Request` object will then be used inside the route method. This example would be more relevant like this IMO:

```python
@app.before_request("/test-headers") # Note this is a middleware
def test_headers_before_request(request: Request):
    request.headers["test"] = "test-header"
    return request
```

docs/features.md
AntoineRR: Ahah my opinion on the subject is that everything should be typed because it allows to catch more errors and provides better hints when coding ^^

docs/features.md
AntoineRR: Please don't forget to change the param to a `Response` in the after middleware. As we already executed the route method, we got rid of the `Request` and now have a `Response`, which is what we can modify in the after middleware.

sansyrox: done!

docs/features.md
sansyrox: @AntoineRR , the reason this example exists is to showcase the possibility. The ability to access headers in middlewares is shown in a section below. 

I am not concerned if it is a sensible thing to do or not, I just want to document that you can do it in Robyn.

I hope this makes sense 😅

sansyrox:
Merging as want to release v0.26.1

PR #466, opened 2023-04-09
feat: Robyn CLI

### LINKED ISSUE #448: Implement Robyn CLI
I have had an idea of creating a cli tool for Robyn.

Possible invocation could be like - `python -m robyn --args`

Now, the responsibilities I had in mind was just `initialization` - e.g. - `python -m robyn --create/--init` and then it will walk you through a list of commands to start the project.

But after yesterday's community call(suggested by @AntoineRR )

We can also include features like

- `dev` - running the dev server
- `run` - running the server by setting some environment variables 
- and more features in the future.

Oluwaseun241:
**Description**
Implementing Robyn cli

This PR fixes #448


robyn/__main__.py
```diff
@@ -0,0 +1,39 @@
...
+    prompt="Enter project name",
+    default="Sample-project",
+    help="The name of your new Robyn project",
+)
+@click.option(
+    "--project-dir",
+    prompt="Enter project directory",
+    default=os.getcwd(),
+    help="The directory where your new Robyn project is created",
+)
```
sansyrox, line 18: @Oluwaseun241 , thank you for the great work. 😄 

I believe we should replace the `click.option` with an `input` instead.

e.g. 

```
def create():
    project_name = input("Enter the project name: ")
    project_dir = input("Enter the name of the project directory: ")
    
    ....
```


This way we will be able to create a flow of question and answers rather than a cli tool where everyone needs to remember the flags.

robyn/__main__.py
```diff
@@ -0,0 +1,39 @@
+import click
+import os
+from robyn import Robyn
```
sansyrox, line 3: Also, we are not using this statement. We can remove this 😄

Oluwaseun241: @sansyrox 
I tested this and this approach don't seems to fix the issue
so....that means we're not using click again??

Oluwaseun241: @sansyrox All done you can now check the improved version

robyn/__main__.py
```diff
@@ -0,0 +1,28 @@
+import os
+
+
+project_name = input("Enter the project name: ")
+project_dir = input("Enter the name of the project directory: ")
+
+
+def create(project_name, project_dir):
```
sansyrox, line 8: @Oluwaseun241 , I didn't mean this 😅 

The code would look like the following

```python3
import click
import os

@click.command()
def create():
  project_name = input("Enter the project name: ")
  project_dir = input("Enter the name of the project directory: ")
  print(f"Creating a new Robyn project '{project_name}' in '{project_dir}'...")

    # Create a new directory for the project
    project_path = os.path.join(project_dir, project_name)
    os.makedirs(project_path)

    # Create the main application file
    app_file_path = os.path.join(project_path, "app.py")
    with open(app_file_path, "w") as f:
        f.write(
            'from robyn import Robyn\n\napp = Robyn()\n\nif __name__ == "__main__":\n app.run()'
        )

    print(f"New Robyn project '{project_name}' created in '{project_dir}' ")
    
if __name__=="__main__":
   create()
```

sansyrox (commented):
I think there was some miscommunication.

Let me know if it still doesn't make sense. 😄

robyn/__main__.py
Oluwaseun241: I think using click is not that efficient again since we're not using the click.option

robyn/__main__.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # Create a new directory for the project
+    project_path = os.path.join(project_dir, project_name)
+    os.makedirs(project_path)
+
+    # Create the main application file
+    app_file_path = os.path.join(project_path, "app.py")
+    with open(app_file_path, "w") as f:
+        f.write(
+            'from robyn import Robyn\n\napp = Robyn()\n\nif __name__ == "__main__":\n app.run()'
```
sansyrox, line 21: One last change. The syntax should be this 

```
            'from robyn import Robyn\n\napp = Robyn(__file__)\n\nif __name__ == "__main__":\n app.run()'
```

Oluwaseun241: @sansyrox  All done

sansyrox (approved):
LGTM! Great work 🔥

PR #481, opened 2023-04-17
Feat: Cli improvements

### LINKED ISSUE #469: Extend Robyn cli
With the release of v0.28.0 - robyn has a `create` function.

Extend it to add

1. Docker config
2. nginx/caddy config
3. suggest more ideas.

Oluwaseun241:
**Description**
This is an improvement to the CLI adding:
-Docker config
-Nginx config

This PR fixes #469

### DIFF
--- a/robyn/__main__.py
+++ b/robyn/__main__.py
@@ -60,12 +60,12 @@ CMD ["python3.10", "/workspace/foo/
 app.py", "--log-level=DEBUG"]
 
                 """
-        )
+            )
     elif docker == "no":
         print("Docker not included")
     else:
         print("Unknown Command")
-    
+
     # Final
     print(f"New Robyn project created in '{project_dir}' ")
 
--- a/robyn/__main__.py
+++ b/robyn/__main__.py
@@ -6,8 +6,15 @@ from .argument_parser import Config
 
 def create():
     project_dir = input("Enter the name of the project directory: ")
-    docker = input("Do you want Docker? ")
+    docker = input("Need Docker? (Y/N) ")
+    docker = check(value)
     # Initailize a new Robyn project
+    def check(value):
+        while value not in ['Y', 'N']:
+            print("Invalid input. Please enter Y or N")
+            value = input("Need Docker? (Y/N) ")
+        return value
+    docker = check(docker)
 
     print(f"Creating a new Robyn project '{project_dir}'...")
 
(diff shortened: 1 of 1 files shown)


sansyrox (commented):
Hey @Oluwaseun241 👋 

Thank you for the PR . Great work! 😄 I am traveling atm but I have a few comments.

sansyrox (commented):
@Oluwaseun241 , thanks for the PR :D

I have a few suggestions

robyn/__main__.py
```diff
@@ -6,6 +6,7 @@
 
 def create():
     project_dir = input("Enter the name of the project directory: ")
+    docker = input("Do you want Docker? ")
```
sansyrox, line 9: Can we change the input prompt to ask for something like

```
project_dir = input("Need Docker? (Y/N)")
```

robyn/__main__.py
```diff
@@ -33,6 +34,39 @@ def index():
             """
         )
 
+    # DockerFile configuration
+    if docker == "yes":
```
sansyrox, line 38: create a sanitization function here, which will take in the value of the variable and check for the `Y/N` value

robyn/__main__.py
```diff
@@ -33,6 +34,39 @@ def index():
...
+app.py", "--log-level=DEBUG"]
+
+                """
+            )
+    elif docker == "no":
+        print("Docker not included")
+    else:
+        print("Unknown Command")
+
+    # Final
```
sansyrox, line 69: we can get rid of this comment?

robyn/__main__.py
sansyrox, line 9: Can we change the input prompt to ask for something like

```
project_dir = input("Need Docker? (Y/N)")
```

robyn/__main__.py
```diff
@@ -6,7 +6,16 @@
 
 def create():
     project_dir = input("Enter the name of the project directory: ")
+    docker = input("Need Docker? (Y/N) ")
+
     # Initailize a new Robyn project
+    def check(value):
+        while value not in ["Y", "N"]:
```
sansyrox, line 13: @Oluwaseun241 , I don't think I want to a loop here. Can you add an `if` check to see if the value is not equal to `Y` or `y`, we replace it with `N`

Oluwaseun241: The loop was there because I don't want them to restart the CLI when there is a mistakenly or genuinely unknown command

sansyrox (approved):
Thank you @Oluwaseun241 😄 

Great work 🔥

Issue #524, opened 2023-05-30
Add rate limit throttling option


IdoKendo:
This issue proposes to add an optional way to include rate limiting on endpoints.
Example usage:
```py
@app.get("/", rate_limit=(3, 1))
async def hello_world():
    return "Hello world"
```
In this case, the limit will be 3 requests per minute:

```bash
❯ curl http://127.0.0.1:8080/ -s -o /dev/null -w "%{http_code}"
200
❯ curl http://127.0.0.1:8080/ -s -o /dev/null -w "%{http_code}"
200
❯ curl http://127.0.0.1:8080/ -s -o /dev/null -w "%{http_code}"
200
❯ curl http://127.0.0.1:8080/ -s -o /dev/null -w "%{http_code}"
429
```

In order to allow multiple workers to track the same rate limit, there will be a need to use a store such as Redis to keep the information. The usage information should be per endpoint and client.

The Redis client can be stored either as an environment variable (ROBYN_REDIS) or as a startup argument (--redis).

In case there is no Redis client defined and any endpoint has a stated rate_limit, the app would store the information in memory. This should be noted on startup in case any endpoint in the app defines a rate limit.

IdoKendo:
Few questions:
1. Would you then use this limiter as part of the decorator? i. e.
```py
from robyn import RateLimiter
from robyn import Robyn

app = Robyn(__file__)
limiter = RateLimiter(app, "100/minute;10/second")

@app.get("/", limiter=limiter)
async def hello_world():
    return "Hello world"
```
Or did you have some other implementation in mind? In my opinion, this is nicer, since you can have different limits on different routes. Maybe some routes need stricter limits than others etc.

2. On the other hand, do you think that this should also allow for global limiting, e.g:
```py
from robyn import RateLimiter
from robyn import Robyn

app = Robyn(__file__)
limiter = RateLimiter(app, "100/minute;10/second")
app.set_limit(limiter)

@app.get("/")
async def hello_world():
    return "Hello world"
```

3. Lastly, on a different note, I started looking into the implementation and I don't see that we keep any client request data to be able to track limits per client, does it exist somewhere and I just missed it, or should I work on implementing it first?

IdoKendo:
After much deliberation and back-and-forth, I've decided to make this feature an external plugin. 
The repo is located at https://github.com/IdoKendo/robyn_rate_limits and it's publish to PyPI under `robyn-rate-limits`.

PR #530, opened 2023-06-02
feat: add ip_addr field to Request object

IdoKendo:
**Description**

This PR contributes towards #524 by adding the client's IP address to the request object.
By having the client's IP address we can later on implement rate limiting per IP address.

An example assert was added to the middleware base routes.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -380,6 +380,10 @@ async fn index(
     req: HttpRequest,
 ) -> impl Responder {
     let mut request = Request::from_actix_request(&req, body, &global_request_headers);
+    request.ip_addr = req
+        .peer_addr()
+        .map(|val| val.ip().to_string())
+        .unwrap_or_else(String::new);
 
     // Before middleware
     // Global
--- a/src/server.rs
+++ b/src/server.rs
@@ -380,10 +380,6 @@ async fn index(
     req: HttpRequest,
 ) -> impl Responder {
     let mut request = Request::from_actix_request(&req, body, &global_request_headers);
-    request.ip_addr = req
-        .peer_addr()
-        .map(|val| val.ip().to_string())
-        .unwrap_or_else(String::new);
 
     // Before middleware
     // Global
--- a/src/types/request.rs
+++ b/src/types/request.rs
@@ -14,6 +14,7 @@ pub struct Request {
     #[pyo3(from_py_with = "get_body_from_pyobject")]
     pub body: Vec<u8>,
     pub url: Url,
+    pub ip_addr: String,
 }
 
 impl ToPyObject for Request {
@@ -33,6 +34,7 @@ impl ToPyObject for Request {
             body,
             method: self.method.clone(),
             url: self.url.clone(),
+            ip_addr: self.ip_addr.clone(),
         };
         Py::new(py, request).unwrap().as_ref(py).into()
     }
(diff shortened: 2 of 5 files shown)


src/server.rs
```diff
@@ -380,6 +380,10 @@ async fn index(
     req: HttpRequest,
 ) -> impl Responder {
     let mut request = Request::from_actix_request(&req, body, &global_request_headers);
```
JackThomson2, line 382: Should the ip_addr not be set in this function?

IdoKendo: You're right! Was kinda planning on doing it and just missed it in the end, will move it.

IdoKendo:
> @IdoKendo , Thanks for the PR. Great work 😄
> 
> I have a question - should we be rate limiting based on the IP address or per client?

Yeah, so for public-facing APIs it's usually limited by IP. If it is a private API and some authentication is in place it should be blocked by auth. So for the rate limit implementation, we'll need to consider both cases.
Not sure where we're standing on authentication for private APIs, but for now I think adding the IP address in the request object is a necessary first step. Also, it could benefit other features in the future as well.

sansyrox:
> Yeah, so for public-facing APIs it's usually limited by IP. If it is a private API and some authentication is in place it should be blocked by auth. So for the rate limit implementation, we'll need to consider both cases.
Not sure where we're standing on authentication for private APIs, but for now I think adding the IP address in the request object is a necessary first step. Also, it could benefit other features in the future as well.

@IdoKendo , perfect then. The authentication in general is still a WIP. So, this will be a good approach.

src/types/request.rs
```diff
@@ -14,6 +14,7 @@ pub struct Request {
     #[pyo3(from_py_with = "get_body_from_pyobject")]
     pub body: Vec<u8>,
     pub url: Url,
+    pub ip_addr: String,
```
JackThomson2, line 17: Just thinking, would an optional string make more sense here? Rather than giving them an empty string if it's missing? Just a thought. Would also save allocating if we didn't have one

IdoKendo: That's actually a good idea, how would that translate on the Python side? as just a `None`?

sansyrox: @IdoKendo , yes. A `None | str`

src/types/request.rs
```diff
@@ -95,6 +99,8 @@ pub struct PyRequest {
     pub method: String,
     #[pyo3(get)]
     pub url: Url,
+    #[pyo3(get)]
+    pub ip_addr: Option<String>,
```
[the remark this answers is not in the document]
JackThomson2: Actix returns the requester info as optional https://docs.rs/actix-web/latest/actix_web/struct.HttpRequest.html#method.peer_addr not sure in what cases this is empty but I guess it's best to handle it at least.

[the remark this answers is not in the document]
sansyrox: Ah, okay. Fair enough. :D

sansyrox (commented):
One last comment @IdoKendo before we can merge this. Can you please write some docs regarding this in the `docs/examples.md` section?

IdoKendo:
> One last comment @IdoKendo before we can merge this. Can you please write some docs regarding this in the `docs/examples.md` section?

No problem, I will add the docs! 😎

sansyrox (approved):
Great work! LGTM 😄

PR #532, opened 2023-06-02
Test: Using snmalloc as the global allocator

JackThomson2:
Just a test for now, OSX is failing. Just want to see if there's much perf difference

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,6 +412,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cmake"
+version = "0.1.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


sansyrox:
Hey @JackThomson2 👋 

Welcome back. How's everything ? 😄 

Can you please tell me more about `snmalloc`? What is it? What does it do? and so on.

JackThomson2:
Hey good thanks. 

Snmalloc replaces the system memory allocator which can improve the performance fairly significantly especially in memory intensive apps such as this one

sansyrox:
@JackThomson2 , I was reading about allocators today. What do you think about something like `je-malloc` ?

sansyrox (commented):
Thanks for the experiment, @JackThomson2 — allocator tuning is genuinely interesting for us. This one was explicitly a "just a test, OSX is failing" spike from 2023, and it's now well out of date (the lockfile predates our Cargo.lock v4 / pyo3 0.27 toolchain). It also overlaps with the mimalloc approach in #593, and main still ships no custom global allocator, so this is really an open design question rather than a mergeable PR. I'd recommend closing this; if we decide to pursue a custom allocator, we should benchmark mimalloc vs snmalloc fresh against current main. Really appreciate you kicking off the investigation.

PR #533, opened 2023-06-04
feat: Migrated Enum style error codes to constants style

### LINKED ISSUE #435: Convert the integer status code to the status code enums
We recently created a `status_code.py` file (https://github.com/sansyrox/robyn/blob/main/robyn/status_codes.py) , which contains an enum for all the status codes.

However, we are still using integer status codes in our codebase(not the tests but the main code). We would like to replace it with the enum values.

Mr-Sunglasses:
**Description**
We are using Enums for the status code it is syntactically not good as we need to call `status.HTTPSTATUS.value` which does not look right sometimes or gets wrong sometimes by the user, instead now we are just using `status.HTTPSTATUS` which is a common practice.

This PR fixes #435


sansyrox:
Hey @Mr-Sunglasses 👋 

Great work. Thank you for the PR :D

We have a reference of the `StatusCodes` enum in the `docs` folder as well. Can you make an edit there too?

Mr-Sunglasses:
@sansyrox Sure

Mr-Sunglasses:
Hey @sansyrox 👋 I have done the necessary changes 😄 .

docs/features.md
```diff
@@ -139,12 +139,12 @@ Robyn provides `StatusCodes` if you want to return type safe Status Responses.
...
 ```python
 
-from robyn import StatusCodes
+from robyn import status_codes
 
 
 @app.get("/response")
 async def response(request):
-    return Response(status_code=StatusCodes.HTTP_200_OK.value, headers={}, body="OK")
+    return Response(status_code=status_codes.HTTP_200_OK.value, headers={}, body="OK")
```
sansyrox, line 147:
```suggestion
    return Response(status_code=status_codes.HTTP_200_OK, headers={}, body="OK")
```

sansyrox (approved):
Great work. LGTM! 🔥

PR #534, opened 2023-06-04
feat: allow before middlewares to return Response objects

### LINKED ISSUE #500: Return a `Response` in before middleware
When adding a before middleware, we can currently only return a `Request` object.
This would be useful to allow returning `Response` objects too, so that when the before middleware returns a `Response`, we do not even run the route method.

AntoineRR:
**Description**

This PR fixes #500

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -375,6 +375,27 @@ async def route_before_request(request: Request):
     request.headers["before"] = "route_before_request"
 ```
 
+In the before middleware, you can choose to directly return a `Response` object. When doing so, the route method and the after middlewares will not be executed.
+
+```python
+def is_user_logged(request: Request):
+    # Check the validity of a JWT cookie for example
+    ...
+
+@app.before_request("/")
+async def hello_before_request(request: Request):
+    if is_user_logged(request):
+        return Response(401, {}, "User isn't logged in!")
+
+@app.get("/")
+async def route(request: Request):
+    print("This won't be executed if user isn't logged")
+
+@app.after_request("/")
+async def hello_after_request(response: Response):
+    print("This won't be executed if user isn't logged")
+```
+
 ## MultiCore Scaling
 
 To run Robyn across multiple cores, you can use the following command:
--- a/docs/features.md
+++ b/docs/features.md
@@ -384,16 +384,17 @@ def is_user_logged(request: Request):
 
 @app.before_request("/")
 async def hello_before_request(request: Request):
-    if is_user_logged(request):
+    if not is_user_logged(request):
+        # The request is aborted, we are returning an error before reaching the route method
         return Response(401, {}, "User isn't logged in!")
 
 @app.get("/")
 async def route(request: Request):
-    print("This won't be executed if user isn't logged")
+    print("This won't be executed if user isn't logged in")
 
 @app.after_request("/")
 async def hello_after_request(response: Response):
-    print("This won't be executed if user isn't logged")
+    print("This won't be executed if user isn't logged in")
 ```
 
 ## MultiCore Scaling
--- a/integration_tests/helpers/http_methods_helpers.py
+++ b/integration_tests/helpers/http_methods_helpers.py
@@ -23,6 +23,14 @@ def get(
     headers: dict = {},
     should_check_response: bool = True,
 ) -> requests.Response:
+    """
+    Makes a GET request to the given endpoint and checks the response.
+
+    endpoint str: The endpoint to make the request to.
+    expected_status_code int: The expected status code of the response.
+    headers dict: The headers to send with the request.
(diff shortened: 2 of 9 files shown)


robyn/router.py
```diff
@@ -136,13 +136,11 @@ def add_middleware(
...
             async def async_inner_handler(*args):
-                await handler(*args)
-                return args
+                return await handler(*args)
 
             @wraps(handler)
             def inner_handler(*args):
-                handler(*args)
-                return args
+                return handler(*args)
```
AntoineRR, line 143: This fixes a bug we never noticed in our tests, were a middleware would not return what it specifies after the `return` keyword, but its own arguments. As we always only modified the arguments in the middleware and returned them it was left unnoticed. It took me sometime to figure it out ^^

sansyrox: Wow! Thank you for catching this. ❤️ 

I can imagine how annoying this would've been.

sansyrox (approved):
Hey @AntoineRR 👋 

Great work as usual 😄 

All looks great. I will just do a final check by tonight and then we will be able to merge it :D

robyn/router.py
AntoineRR: Thanks that wasn't easy to find ^^

integration_tests/helpers/http_methods_helpers.py
```diff
@@ -14,11 +14,15 @@ def check_response(response: requests.Response, expected_status_code: int):
 
 
 def get(
```
sansyrox, line 16: @AntoineRR , one question. What does `should_check_response` do?

Can we add a docstring here?

Also, do you think we should keep the `http_method_helpers` homogeneous in terms of the function definition?

AntoineRR:
I added some documentation for this new feature as well :slightly_smiling_face:

src/server.rs
```diff
@@ -383,29 +383,32 @@ async fn index(
...
+        middleware_router.get_route(&MiddlewareType::BeforeRequest, req.uri().path())
+    {
+        before_middlewares.push(function);
+        request.path_params = route_params;
+    }
+    for before_middleware in before_middlewares {
+        request = match execute_middleware_function(&request, &before_middleware).await {
+            Ok(MiddlewareReturn::Request(r)) => r,
+            Ok(MiddlewareReturn::Response(r)) => {
+                // If a before middleware returns a response, we abort the request and return the response
```
sansyrox, line 399: I think we should ideally be propagating the response to the next middleware, no?

integration_tests/helpers/http_methods_helpers.py
sansyrox: Thank you :D :D

src/server.rs
AntoineRR: The input of a before middleware is a `Request` object, not a `Response`. I don't think we can send it to the next middleware. I have some ideas to implement middleware chaining in the future but I think if one of the before middleware returns a `Response` on the way, we should abort the request ?
We could only pass the `Response` to the after middlewares because that is what they use as an input, but I am not sure we should do this as we may want to consider the user has be authenticated (which would be done in a before middleware) when the request reaches the after middleware if that makes sense.

sansyrox: I was thinking of implementing something like this (https://emmett.sh/docs/2.2.x/pipeline#pipeline ) when I was talking about middleware chaining. We can have a `send_to_next` and `recieved_from_prev` arguments or similar?

docs/features.md
```diff
@@ -375,6 +375,27 @@ async def route_before_request(request: Request):
...
 
+In the before middleware, you can choose to directly return a `Response` object. When doing so, the route method and the after middlewares will not be executed.
+
+```python
+def is_user_logged(request: Request):
+    # Check the validity of a JWT cookie for example
+    ...
+
+@app.before_request("/")
+async def hello_before_request(request: Request):
```
sansyrox, line 386: @AntoineRR , can we just mention here as a comment or something that the request will be aborted here?

src/server.rs
AntoineRR: Thank you for sharing this it is very interesting! I think what he calls pipe is the equivalent of our before middleware. We could have something similar but that would require reworking the whole middleware architecture. I think it is outside the scope of this PR but this article can be very valuable to implement a miiddleware chaining solution later :)

docs/features.md
AntoineRR: Sure, I just updated it :)

sansyrox (approved):
LGTM :D Great work! 🔥

src/server.rs
sansyrox: >  I think it is outside the scope of this PR but this article can be very valuable to implement a miiddleware chaining solution later :)

Yes, absolutely. I agree 😄

PR #536, opened 2023-06-08
Don't respond with pong when a pong message is received.

urjitbhatia:
Makes the server aligned with the WS RFC.

From: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets

See: "You might also get a pong without ever sending a ping; ignore this if it happens."

Fette & Melnikov             Standards Track                   [Page 37]
RFC 6455                 The WebSocket Protocol            December 2011

   A Pong frame MAY be sent unsolicited.  This serves as a
   unidirectional heartbeat.  A response to an unsolicited Pong frame is
   not expected.

Ref: https://datatracker.ietf.org/doc/rfc6455

### DIFF
--- a/src/web_socket_connection.rs
+++ b/src/web_socket_connection.rs
@@ -104,7 +104,6 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for MyWs {
             }
             Ok(ws::Message::Pong(msg)) => {
                 debug!("Pong message {:?}", msg);
-                ctx.pong(&msg)
             }
             Ok(ws::Message::Text(_text)) => {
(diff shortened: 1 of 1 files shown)


sansyrox:
Hey @urjitbhatia 👋 

Firstly, apologies for the late PR review. 

Secondly, this is great. I was not aware of this. Thank you 😄

urjitbhatia:
thanks @sansyrox :)

sansyrox (approved):
LGTM! 😄

PR #537, opened 2023-06-08
Adds ability to share messages from client with py

urjitbhatia:
This PR fixes: https://github.com/sansyrox/robyn/issues/440

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -1,4 +1,3 @@
-import json
 import os
 
 import pathlib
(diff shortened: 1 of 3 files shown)


urjitbhatia:
Note: This isn't a real performance regression since it changes the what the test is doing a bit.

integration_tests/base_routes.py
```diff
@@ -17,20 +17,23 @@
...
-    return response
+    state = websocket_state[websocket_id]
+    if state == 0:
+        response["resp"] = "Whaaat??"
+    elif state == 1:
+        response["resp"] = "Whooo??"
+    elif state == 2:
+        response["resp"] = "*chika* *chika* Slim Shady."
+    websocket_state[websocket_id] = (state + 1) % 3
+    return json.dumps(response)
```
sansyrox, line 36: we can use `jsonify` method from robyn here

urjitbhatia:
@sansyrox switched to `jsonify` and also restored the original WS `benchmark` test since including the json parsing in it pollutes what is actually getting measured.

sansyrox (approved):
LGTM! Great work 🔥

PR #539, opened 2023-06-12
feat: add rate limit throttling

### LINKED ISSUE #524: Add rate limit throttling option

IdoKendo:
**Description**

This PR fixes #524 

There are two approaches to storing the usage of an API endpoint:
1. Using Redis
2. Local in-memory store

While using Redis is the preferred method, I did not want to impose setting up Redis for Robyn users, especially if they might not even want to rate limit their APIs.

The implementation is not done, there are a few more things I would like to add, thus, this PR is still a WIP.

I am having issues with the in-memory store due to a `RuntimeError: already borrowed` error when I try to `curl` requests that have a rate limiter attached to them. Using Redis the rate limiter is working as expected.

@sansyrox If you could take a look and advise it would be very appreciated since I am a bit stuck here. Thanks!

```sh
python integration_tests/base_routes.py
curl HTTP://localhost:8080/sync/rate/get -v
```


sansyrox:
Hey @IdoKendo 👋 

Thank you for the great work 😄 I will take a deeper look by this weekend for sure. :D

JackThomson2:
This is very cool, I just wonder if it makes sense to be in the core Robyn implementation? It'd be adding an additional Redis dependency to all users, which wouldn't be ideal. I imagine a lot of people run robyn behind a load balancer anyway which would do the rate limiting. Not a blocker at all, just asking the question.

Could this not be an external plugin / repository that hooks into the `before_request` handlers?

IdoKendo:
> This is very cool, I just wonder if it makes sense to be in the core Robyn implementation? It'd be adding an additional Redis dependency to all users, which wouldn't be ideal. I imagine a lot of people run robyn behind a load balancer / proxy anyway which would do the rate limiting. Not a blocker at all, just asking the question.
> 
> Could this not be an external plugin / repository that hooks into the `before_request` handlers?

I think making the `Redis` part an optional dependency would solve the issue of adding it to all users. That's also my plan once I get over the hurdle of the in-memory cache. I think it can be even more generic - Some users might want to use `memcached` instead of `Redis`, and it can also be supported.

Surely it can be developed externally, as a plugin, but I think there is an advantage in having it be included in the standard implementation.

sansyrox:
Hey @IdoKendo @JackThomson2 👋 

Thank you for your comments here 😄  I partly agree with both of you here.

We should add rate limiting as a core feature for one but should keep redis/memcached as an optional dependency like we have for templates.

And as for the issue with the implementation, @IdoKendo give me a few more hours. I will run it locally and get back to you 😄

sansyrox:
@IdoKendo , I think I may have found a part reason. This looks like a circular dependency here.

Screen Shot 2023-06-14 at 21 09 22

sansyrox:
And the second error is due to this
Screen Shot 2023-06-14 at 21 20 08

sansyrox:
We get a better error message 

```

21:19:18.717385 line        42         current_timestamp = int(time.time())
New var:....... current_timestamp = 1686773958
21:19:18.717564 line        43         if pool:
21:19:18.717739 line        49         elif self.app:
21:19:18.717911 line        50             calls_count = self.app.server.get_calls_count(
21:19:18.718084 line        51                 limit_key, self.limit_ttl, current_timestamp
21:19:18.718255 line        50             calls_count = self.app.server.get_calls_count(
thread 'actix-rt|system:0|arbiter:0' panicked at 'Already mutably borrowed: PyBorrowError', /Users/sanskar/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.18.1/src/pycell.rs:298:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
21:19:18.718632 exception   50             calls_count = self.app.server.get_calls_count(
Exception:..... pyo3_runtime.PanicException: Already mutably borrowed: PyBorrowError
Call ended by exception
Elapsed time: 00:00:00.002641
--- PyO3 is resuming a panic after fetching a PanicException from Python. ---
Python stack trace below:
  File "/Users/sanskar/repos/robyn/robyn/router.py", line 113, in inner_handler
    if client and rate_limiter and rate_limiter.limit_exceeded(
  File "/Users/sanskar/repos/robyn/venv/lib/python3.10/site-packages/pysnooper/tracer.py", line 305, in simple_wrapper
    return function(*args, **kwargs)
  File "/Users/sanskar/repos/robyn/robyn/throttling.py", line 50, in limit_exceeded
    calls_count = self.app.server.get_calls_count(
pyo3_runtime.PanicException: Already mutably borrowed: PyBorrowError
```

IdoKendo:
> @IdoKendo , I think I may have found a part reason. This looks like a circular dependency here.
> 
> Screen Shot 2023-06-14 at 21 09 22

Thanks for helping to look into it!
Maybe I'm missing something, but why is this a circular dependency?

sansyrox:
Hey @IdoKendo , I made a typo sorry. I meant a circular reference and not a dependency but that should not cause any issues.

I found a fix to our problem.

See this
Screen Shot 2023-06-15 at 12 25 55

IdoKendo:
> Hey @IdoKendo , I made a typo sorry. I meant a circular reference and not a dependency but that should not cause any issues.
> 
> I found a fix to our problem.
> 
> See this Screen Shot 2023-06-15 at 12 25 55

Ohh, great! I think this unblocks me! Thank you! I will continue working on this PR and update you when it's ready.

sansyrox:
Not a problem 😄 I am really excited for this feature

src/server.rs
```diff
@@ -360,6 +363,39 @@ impl Server {
...
+        limit_ttl: u32,
+        current_timestamp: u32,
+    ) -> Vec<u32> {
+        let server = self_.try_borrow_mut();
+        match server {
+            Ok(server) => {
+                let mut cache = server.memory_store.lock().unwrap();
+                let ttl = current_timestamp - limit_ttl;
+                if let Some(timestamps) = cache.get_mut(&limit_key) {
+                    let mut valid_timestamps: Vec<u32> = timestamps
```
JackThomson2, line 380: Can't this be done with 
```Rust
timestamps.retain(|timestamp| timestamp >= &ttl);
timestamps.push(current_timestamp);
```

Also won't this list always be ordered? Couldn't we just binary search for the current time stamp and drop all to the left? A linked list may be more efficient for this as well.

src/server.rs
```diff
@@ -360,6 +363,39 @@ impl Server {
...
         debug!("Added shutdown handler {:?}", self.shutdown_handler);
     }
+
+    /// Get and set call list from memory store
+    pub fn get_calls_list(
+        self_: &PyCell<Self>,
+        limit_key: String,
+        limit_ttl: u32,
+        current_timestamp: u32,
+    ) -> Vec<u32> {
```
JackThomson2, line 373: Can this just not return the length? We're doing expensive clones here, and only ever using the length python side

src/server.rs
```diff
@@ -56,6 +57,7 @@ pub struct Server {
     directories: Arc<RwLock<Vec<Directory>>>,
     startup_handler: Option<Arc<FunctionInfo>>,
     shutdown_handler: Option<Arc<FunctionInfo>>,
+    memory_store: Arc<Mutex<HashMap<String, Vec<u32>>>>,
```
JackThomson2, line 60: Do we need this to be wrapped in an Arc Mutex? This comes with some overhead, since this method always come from python, we can assume only one thread will be reading / writing from it at a time. I'm not sure if we can get a `&mut self` in the `get_calls_list` method

JackThomson2: If you think we will be accessing this concurrently or can't get a `&mut self` reference we should use `Arc<Dashmap<...>>` this is already imported and would do better if we do hit this at the same time

JackThomson2: Also another query, will we need a background thread to clear these out? Current if a client never requests again the entry will stay indefinitely, I could see memory growing if used by a lot of people

IdoKendo:
Hey @sansyrox and @JackThomson2 👋 
Thanks for all the comments and help! I've been away for a bit, but getting back to it!

For now, I've decided to remove the Redis support and add it later - to reduce the scale of this PR to a more bite-sized change.

I've changed the implementation quite a bit and still a WIP but seems to be in a much more positive direction.

If you have any helpful guidance and comments regarding the new direction, I would greatly appreciate it!

clippy.toml
sansyrox, line 1: nicee

sansyrox:
Hey @IdoKendo 👋 

Welcome back 😄 

I will start taking a look soon!

sansyrox (commented):
Hey @IdoKendo 👋 

I had an initial look on the PR and looks good 😄 

I will have a thorough look till Thursday 😄

robyn/throttling.py
```diff
@@ -0,0 +1,28 @@
+import time
+from typing import Dict
+
+
+class RateLimiter:
+    def __init__(
+        self,
+        *,
+        calls_limit: int = 0,
+        limit_ttl: int = 0,
```
sansyrox, line 10: One question - Is there a reason to allow only `kwargs`?

robyn/throttling.py
```diff
@@ -0,0 +1,28 @@
...
+    ) -> None:
```
sansyrox, line 11: Can you please add a docstring for this class?

src/server.rs
```diff
@@ -220,6 +227,27 @@ impl Server {
...
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap()
+                            .as_secs();
+                        for item in calls_count.iter() {
+                            let mut valid_timestamps = item.value().clone();
+                            valid_timestamps.retain(|timestamp| timestamp >= &ttl);
+                            calls_count.insert(item.key().clone(), valid_timestamps);
+                            calls_count.remove_if(item.key(), |_, t| t.is_empty());
+                        }
+                        thread::sleep(Duration::from_secs(60));
```
sansyrox, line 245: One question - why add a sleep for just `60` seconds? Can we add some comment for this magic number please?

src/server.rs
```diff
@@ -376,11 +404,41 @@ async fn index(
...
+    let calls_count_header = calls_count
+        .get(&limit_key)
+        .unwrap()
+        .iter()
+        .map(ToString::to_string)
+        .collect::<Vec<String>>()
+        .join(",");
+    request
+        .headers
+        .insert(String::from("x-robyncc"), calls_count_header);
```
sansyrox, line 440: Can you please explain the logic here?

integration_tests/base_routes.py
sansyrox, line 1: Nicee

sansyrox (commented):
Hey @IdoKendo 👋 

Great work! 🔥 

I have left an initial review. Do let me know if you require any more information from me 😄

robyn/throttling.py
IdoKendo: I think having something like `RateLimiter(60, 120)` is very not intuitive and forces whoever reads the code to check it every time or to add some comment e.g. `# 60 calls per two minutes` - Forcing it to be `RateLimiter(calls_limit=60, limit_ttl=120)` is much more readable IMO.

src/server.rs
IdoKendo: I was actually contemplating this - How often do we want to clear the cache of calls - I'm thinking if there were no calls for a minute it should cover most use cases, but maybe it should be configurable in env? WDYT?

src/server.rs
IdoKendo: Sure. The rate limiting is done via a sliding window, however, in this part of the program, we don't know where the window starts. In order to pass it forward into the python part I'm embedding it in a custom header that I can later read and check how many calls fall in the allowed window.

robyn/throttling.py
sansyrox: Wow! This makes sense. Perfect.

src/server.rs
sansyrox: > but maybe it should be configurable in env? WDYT?

This approach sounds good. Would it be possible to add this config in the `robyn.env` file?

robyn/__init__.py
```diff
@@ -56,6 +58,7 @@ def __init__(self, file_object: str, config: Config = Config()) -> None:
         self.event_handlers = {}
         self.exception_handler: Optional[Callable] = None
         self.authentication_handler: Optional[AuthenticationHandler] = None
+        self.rate_limiter = RateLimiter()
```
sansyrox, line 61: One question - do we always want to have a ratelimiter instance? We can make it optional and handle the cases in the code? Or make as constant argument in the `__init__` function like config

src/server.rs
```diff
@@ -220,6 +227,25 @@ impl Server {
...
+                    let ttl = SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .unwrap()
+                        .as_secs();
+                    for item in calls_count.iter() {
+                        let mut valid_timestamps = item.value().clone();
+                        valid_timestamps.retain(|timestamp| timestamp >= &ttl);
+                        calls_count.insert(item.key().clone(), valid_timestamps);
+                        calls_count.remove_if(item.key(), |_, t| t.is_empty());
+                    }
```
sansyrox, line 243: One suggestion, we can reduce the code here

```python
                    for mut item in calls_count.iter_mut() {
                        let valid_timestamps = item.value_mut();
                        valid_timestamps.retain_mut(|timestamp| timestamp >= &mut ttl);
                        calls_count.remove_if(item.key(), |_, t| t.is_empty());
                    }

```

src/server.rs
```diff
@@ -220,6 +227,25 @@ impl Server {
...
         });
 
+        let calls_count = self.calls_count.clone();
+        Python::with_gil(|py| {
+            py.allow_threads(|| {
+                thread::spawn(move || loop {
+                    let ttl = SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .unwrap()
+                        .as_secs();
```
sansyrox, line 237: Correct me if I am wrong but the `ttl` variable here is the seconds elapsed from `1970-1-1`?

sansyrox: Then after 60 seconds(thread::sleep) , this variable will be t+60, and remove all the previous timestamps, right?

sansyrox: I am wondering if the `ttl` specified in the RateLimit constructor and the `ttl` variable here should be connected 🤔

src/server.rs
sansyrox: Thank you, makes sense 😄

sansyrox:
Hey @IdoKendo 👋 

Thank you for making the latest changes. I have left a new review 😄

robyn/__init__.py
IdoKendo: I prefer having the RateLimiter instance handle the details of whether to check the limit etc, keep all the responsibilities there, than to handle the cases in the Robyn class.

src/server.rs
IdoKendo: Yes, this TTL is for retention of old calls, the idea for this thread is to clean up calls of old users, to avoid having the calls_count inflate endlessly. Not sure it should be related to the TTL for the RateLimiter, especially since different endpoints (RateLimiter instances) can have different TTLs.

IdoKendo:
> Hey @IdoKendo 👋
> 
> Thank you for making the latest changes. I have left a new review 😄

Thanks for the review! Helping us nail this implementation down 😎 

I do wonder, I see that CI is always failing in Ubuntu yet passes in other platforms - Any thought what could be the root cause? Can't really see an issue there.

src/server.rs
sansyrox: @IdoKendo , I see, this makes sense. Maybe in the future we could also add the ability to clear the cache based on the size of the cache.

But one thing - let us have an environment variable to make it configurable.

robyn/__init__.py
sansyrox: @IdoKendo , alright makes sense. However, can we make it a default argument in the constructor? This will also allow the users to extend the RateLimiter with a custom implementation if they desire to.

sansyrox:
Hey @IdoKendo 👋 

Thank you for the update. I have just left one final comment 😄 

> I do wonder, I see that CI is always failing in Ubuntu yet passes in other platforms - Any thought what could be the root cause? Can't really see an issue there.

The error seems to be here
```
r = method(route, expected_status_code=429) 
```

This is resulting in a 200 status code. Looks like the rate limit is not being applied on Ubuntu. I can try testing it on my real linux machine tomorrow 😅

IdoKendo:
> 
> The error seems to be here
> 
> ```
> r = method(route, expected_status_code=429) 
> ```
> 
> This is resulting in a 200 status code. Looks like the rate limit is not being applied on Ubuntu. I can try testing it on my real linux machine tomorrow 😅

Thanks! Let me know if I can help - can't seem to reproduce it on my end.

sansyrox:
Hey @IdoKendo :wave: 

I did run the tests. However, I found an unusual observation

The test run perfectly if we run just them , i.e. 

```
venv) [batman@batmachine robyn]$ pytest -vv -k rate_limiting
============================================================ test session starts ============================================================
platform linux -- Python 3.11.3, pytest-7.2.1, pluggy-1.2.0 -- /home/batman/repos/robyn/venv/bin/python3
codspeed: 2.0.1 (callgraph: not supported)
cachedir: .pytest_cache
rootdir: /home/batman/repos/robyn
plugins: codspeed-2.0.1
collected 116 items / 110 deselected / 6 selected                                                                                           

integration_tests/test_rate_limiting.py::test_throttling[/sync/rate/get-get] PASSED                                                   [ 16%]
integration_tests/test_rate_limiting.py::test_throttling[/async/rate/get-get] PASSED                                                  [ 33%]
integration_tests/test_rate_limiting.py::test_throttling[/sync/rate/put-put] PASSED                                                   [ 50%]
integration_tests/test_rate_limiting.py::test_throttling[/async/rate/put-put] PASSED                                                  [ 66%]
integration_tests/test_rate_limiting.py::test_throttling[/sync/rate/post-post] PASSED                                                 [ 83%]
integration_tests/test_rate_limiting.py::test_throttling[/async/rate/post-post] PASSED                                                [100%]

===================================================== 6 passed, 110 deselected in 0.22s =====================================================

```


However, they fail when run with everyone else

```
venv) [batman@batmachine robyn]$ pytest 
============================================================ test session starts ============================================================
platform linux -- Python 3.11.3, pytest-7.2.1, pluggy-1.2.0
codspeed: 2.0.1 (callgraph: not supported)
rootdir: /home/batman/repos/robyn
plugins: codspeed-2.0.1
collected 116 items                                                                                                                         

integration_tests/test_app.py ....                                                                                                    [  3%]
integration_tests/test_authentication.py ........                                                                                     [ 10%]
integration_tests/test_base_url.py ....                                                                                               [ 13%]
integration_tests/test_basic_routes.py ..................                                                                             [ 29%]
integration_tests/test_binary_output.py ....                                                                                          [ 32%]
integration_tests/test_delete_requests.py ....                                                                                        [ 36%]
integration_tests/test_env_populator.py .                                                                                             [ 37%]
integration_tests/test_exception_handling.py ......                                                                                   [ 42%]
integration_tests/test_file_download.py ..                                                                                            [ 43%]
integration_tests/test_get_requests.py ..........                                                                                     [ 52%]
integration_tests/test_middlewares.py ....                                                                                            [ 56%]
integration_tests/test_patch_requests.py ....                                                                                         [ 59%]
integration_tests/test_post_requests.py ....                                                                                          [ 62%]
integration_tests/test_put_requests.py ....                                                                                           [ 66%]
integration_tests/test_rate_limiting.py FF.FFF                                                                                        [ 71%]
integration_tests/test_status_code.py .....                                                                                           [ 75%]
integration_tests/test_subrouter.py .........                                                                                         [ 83%]
integration_tests/test_unsupported_types.py .........                                                                                 [ 91%]
integration_tests/test_views.py ........                                                                                              [ 98%]
integration_tests/test_web_sockets.py ..                                                 
```

sansyrox:
This implies an issue with the fixture being called from these throttling tests.

IdoKendo:
> This implies an issue with the fixture being called from these throttling tests.

What fixture? the paramterize?

sansyrox:
Hey @IdoKendo 👋 

I have pushed a change. Let's see if that will work.

sansyrox:
Hey @IdoKendo 👋 

All looks good to me 😄 Can you please update some documentation for the same?

IdoKendo:
> Hey @IdoKendo 👋
> 
> All looks good to me 😄 Can you please update some documentation for the same?

Done 😃

sansyrox (approved):
LGTM! 🔥 👍 Great work!

JackThomson2:
This adds a non-trivial amount of code to the hotpath even if it isn't enabled. Isn't this going to slow down robyn significantly?

sansyrox:
@JackThomson2 , thanks for pointing this out. I just ran the benchmarks and the regressions are too much. 

This is the after
```
 ~/repos/robyn  main *47  oha  --insecure -c 100 -n 100000 http://localhost:8080                                                                              ✔  1m 32s  robyn Py
Summary:
  Success rate: NaN
  Total:        5.3133 secs
  Slowest:      -inf secs
  Fastest:      inf secs
  Average:      0.0000 secs
  Requests/sec: 0.0000

  Total data:   0 B
  Size/request: NaN
  Size/sec:     0 B

Response time histogram:

Latency distribution:
  10% in NaN secs
  25% in NaN secs
  50% in NaN secs
  75% in NaN secs
  90% in NaN secs
  95% in NaN secs
  99% in NaN secs

Details (average, fastest, slowest):
  DNS+dialup:   0.0000 secs, inf secs, -inf secs
  DNS-lookup:   0.0000 secs, inf secs, -inf secs

Status code distribution:

 ~/repos/robyn  main *47  oha  --insecure -c 100 -n 100000 http://localhost:8080                                                                                  ✔  5s  robyn Py
Summary:
  Success rate: 0.1492
  Total:        26.7531 secs
  Slowest:      0.4223 secs
  Fastest:      0.0145 secs
  Average:      0.1818 secs
  Requests/sec: 2948.1840

  Total data:   126.38 KiB
  Size/request: 11 B
  Size/sec:     4.72 KiB

Response time histogram:
  0.015 [1]    |
  0.055 [1569] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.096 [1456] |■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.137 [1765] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.178 [1623] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.218 [1452] |■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.259 [945]  |■■■■■■■■■■■■■■■■■
  0.300 [663]  |■■■■■■■■■■■■
  0.341 [803]  |■■■■■■■■■■■■■■
  0.381 [939]  |■■■■■■■■■■■■■■■■■
  0.422 [549]  |■■■■■■■■■

```

and this was the before. 

```
Summary:
  Success rate: 1.0000
  Total:        1.5308 secs
  Slowest:      0.0316 secs
  Fastest:      0.0047 secs
  Average:      0.0153 secs
  Requests/sec: 6532.5208

  Total data:   107.42 KiB
  Size/request: 11 B
  Size/sec:     70.17 KiB

Response time histogram:
  0.005 [1]    |
  0.007 [10]   |
  0.010 [2]    |
  0.013 [3273] |■■■■■■■■■■■■■■■■■■■■
  0.015 [489]  |■■
  0.018 [5224] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.021 [924]  |■■■■■
  0.024 [40]   |
  0.026 [1]    |
  0.029 [23]   |
  0.032 [13]   |

Latency distribution:
  10% in 0.0111 secs
  25% in 0.0118 secs
  50% in 0.0167 secs
  75% in 0.0174 secs
  90% in 0.0181 secs
  95% in 0.0186 secs
  99% in 0.0193 secs

Details (average, fastest, slowest):
  DNS+dialup:   0.0024 secs, 0.0007 secs, 0.0033 secs
  DNS-lookup:   0.0000 secs, 0.0000 secs, 0.0000 secs

Status code distribution:
```

@IdoKendo , I will be reverting this PR for now. @JackThomson2 , do you have any suggestions on how we can make it faster?

I also wonder why codspeed didn't show us any regressions.

sansyrox:
@IdoKendo , can you please remove this PR?

JackThomson2:
You could try putting this code [here](https://github.com/sparckles/robyn/pull/539/files#diff-48d27698cc708c7efe770fd897e4885efaa0a15cae30d54936068603ba03bbd9R423-R449) behind a conditional if rate limiting is enabled on the specified endpoint, it may help but only if it isn't enabled. As for fixing it when rate limiting *is* enabled. 

The code [here](https://github.com/sparckles/robyn/pull/539/files#diff-48d27698cc708c7efe770fd897e4885efaa0a15cae30d54936068603ba03bbd9R433-R438) does a lot of allocating which I understand why this approach was taken as you'd need a mutable reference. You could try putting the array in a mutex and acquiring a lock to append too it (this should make the rate limiting more accurate as currently if two requests are made at once there's a chance one is missed.) It's a complicated problem and would require some thought into how to make it performant, just some ideas from the top of my head

IdoKendo:
I've reopened the PR at #572

PR #545, opened 2023-06-22
make returning text from websocket messages optional

urjitbhatia:
`websocket.on("message")` methods shouldn't be required to return a string value. This patch makes it optional to send a response.
The RFC does not mandate sending a text response when an application level data frame is received.


sansyrox (approved):
Thanks @urjitbhatia 😄 

LGTM 🔥

sansyrox:
@urjitbhatia , rustfmt is failing. Can you please address that?

sansyrox:
Perfect, now. Great work @urjitbhatia ✨

PR #546, opened 2023-06-23
Feat/handle unwraps safely

urjitbhatia:
**Description**

This PR fixes #

### DIFF
--- a/src/web_socket_connection.rs
+++ b/src/web_socket_connection.rs
@@ -1,4 +1,4 @@
-use crate::types::function_info::FunctionInfo;
+use std::collections::HashMap;
 
 use actix::prelude::*;
 use actix::{Actor, AsyncContext, StreamHandler};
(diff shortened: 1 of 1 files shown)


sansyrox:
Hey @urjitbhatia 👋 

I hope all's well. Just to confirm is this still supposed to be in draft ?😅

urjitbhatia:
Hello @sansyrox - I am still working on this in the background. I am currently focused on a different project at work but will soon come back to this.
The idea is to slowly replace the `unwraps` with safer code handling with detailed error logs etc.

Hopefully, I'll be sending an update soon! 😄

sansyrox:
Thanks for the update @urjitbhatia 😄 

No rush at all :D

sansyrox (commented):
Thanks for this, @urjitbhatia — swapping that `.unwrap()` chain for real error handling was the right instinct. Since 2023 the WebSocket layer has been refactored away from `src/web_socket_connection.rs` entirely, so this patch no longer applies, and the good news is the concern is already resolved on main: `src/executors/web_socket_executors.rs` now does a proper `match result.extract::<Option<String>>()` instead of unwrapping. Given that, I think we can safely close this one as addressed. Thanks for helping push the codebase toward safer error handling!

PR #564, opened 2023-07-21
feat: replace default jsonify with an orjson powered jsonify

### LINKED ISSUE #525: Jsonify from rust
Use the library called serde and `pythonize`

sansyrox:
Robyn exposes a function named `jsonify`, which used the default json library. This commit replaces it with the faster `orjson` library.

**Description**

This PR fixes #525


sansyrox:
Screen Shot 2023-07-22 at 17 12 49
Surprisingly orjson causes a regression

PR #572, opened 2023-07-31
feat: add rate limit throttling

### LINKED ISSUE #524: Add rate limit throttling option

IdoKendo:
**Description**

This PR fixes #524 

Reopening the PR after it was reverted in #539 - See earlier discussion there.


IdoKendo:
After much deliberation and back-and-forth, I've decided to make this feature an external plugin. 
The repo is located at https://github.com/IdoKendo/robyn_rate_limits and it's publish to PyPI under `robyn-rate-limits`.

PR #575, opened 2023-08-03
Gave user option to enable and disable Robyn generated logs

### LINKED ISSUE #482: Disable robyn logs in dev(default) mode
Issue:

Currently, while running our application in development mode, we are seeing logs from the Robyn framework mixed with our application logs. It's making it challenging to sift through the logs and focus on application-specific debugging information. To have a clear distinction and improve the debugging experience, an environment variable should be introduced to control the visibility of Robyn framework logs.

Expected Behavior:

By default, only application logs should be visible in development mode. However, if the developer needs to debug something related to the Robyn framework, they should be able to enable Robyn logs by setting a specific environment variable.

Steps to Reproduce:

Run the application in development mode.
Observe that the console shows both application and Robyn framework logs.
Suggested Solution:

Introduce a new environment variable, say ENABLE_ROBYN_LOGS, to control the logging of the Robyn framework. By default, this should be false, hiding Robyn logs in dev mode. If a developer sets ENABLE_ROBYN_LOGS=true, Robyn logs should then be visible.

mayachidambaram:
**Description**

This PR fixes #482 

(Not completely finished yet, gave user the option to enable and disable the Robyn generated Python logs while in dev mode but not the Robyn generated Rust logs yet).


robyn/logger.py
sansyrox, line 1: Hey @mayachidambaram :wave: 

Just to nudge you in the right direction. The `Colors` enum is only an enum and changing the logging mechanism here will only disable the colors and the entire logs.

mayachidambaram: Hi @sansyrox
I'm a bit confused by this suggestion because when I set up an environment variable called ENABLE_ROBYN_LOGS=true in the terminal, all the python logs appear and when ENABLE_ROBYN_LOGS=false, all of the python logs disappear. 

For example if I run the sample app.py file using "python3 app.py --dev" this is what appears when it's automatically set to false:

DEBUG:robyn.server:The number of workers is 1
INFO:actix_server.builder:starting 1 workers
INFO:actix_server.server:Actix runtime found; starting in Actix runtime
DEBUG:robyn.server:Max payload size is 1000000

**When we open up port 8080**

INFO:root:This is my dev log **<---- This is a log I created in app.py, acting as an application log**
DEBUG:robyn.server:Response: Response { status_code: 200, response_type: "text", headers: {"Content-Type": "text/plain"}, body: [67, 72, 69, 82, 82, 89, 32, 66, 79, 77, 66, 33], file_path: None }

**When we stop running the application**

DEBUG:robyn.server:Ctrl c handler
DEBUG:watchdog.observers.inotify_buffer:in-event <InotifyEvent: src_path=b'/workspaces/robyn', wd=1, mask=IN_IGNORED, cookie=0, name=>

and when set to true, we get these Robyn logs:

INFO:robyn.logger:SERVER IS RUNNING IN VERBOSE/DEBUG MODE. Set --log-level to WARN to run in production mode. **in blue**
INFO:robyn.logger:Dev server initialized with the directory_path : /workspaces/robyn **in blue**
INFO:robyn.logger:SERVER IS RUNNING IN VERBOSE/DEBUG MODE. Set --log-level to WARN to run in production mode. **in blue**
INFO:robyn.logger:Robyn version: 0.37.0 **in green**
INFO:robyn.logger:Starting server at 127.0.0.1:8080 **in green**
INFO:robyn.logger:Press Ctrl + C to stop **green**

DEBUG:robyn.server:The number of workers is 1
INFO:actix_server.builder:starting 1 workers
INFO:actix_server.server:Actix runtime found; starting in Actix runtime
DEBUG:robyn.server:Max payload size is 1000000

**When we open up port 8080**

INFO:root:This is my dev log
DEBUG:robyn.server:Response: Response { status_code: 200, response_type: "text", headers: {"Content-Type": "text/plain"}, body: [67, 72, 69, 82, 82, 89, 32, 66, 79, 77, 66, 33], file_path: None }

**When we stop running the application**

INFO:actix_server.server:SIGINT received; starting forced shutdown
DEBUG:actix_server.accept:paused accepting connections on 127.0.0.1:8080
INFO:actix_server.accept:accept thread stopped
INFO:actix_server.worker:shutting down idle worker
INFO:robyn.logger:Terminating reloader **in green**
DEBUG:robyn.server:Ctrl c handler
INFO:robyn.logger:Terminating server!! **in green**
INFO:robyn.logger:Terminating server!! **in green**

The logs that are currently showing (with the exception of "INFO:root:This is my dev log") were logs that my team and I assumed were Rust generated logs since we found those log messages behind "debug!" functions in .rs files. 

We assumed that the issue for use to solve was to completely hide the Robyn logs if ENABLE_ROBYN_LOGS=false and show the Robyn logs is ENABLE_ROBYN_LOGS=true. Would it be okay if you elaborated a little bit, because we believe that we did that on the Python side of things and some clarification would be greatly appreciated if we're misinterpreting something.

mayachidambaram:
I actually think I may know what you are talking about now. 

This line in particular:  self.logger.info(self._format_msg(msg, color, bold, underline)) 

This is being called in an additional file called events.py four times. Are you suggesting that we have to turn off those instances as well?

sansyrox:
@mayachidambaram :wave: 

> This line in particular: self.logger.info(self._format_msg(msg, color, bold, underline))

This is being called in an additional file called events.py four times. Are you suggesting that we have to turn off those instances as well?

Nope. We will need to change the configuration of `self.logger`

sansyrox:
Hey @mayachidambaram 👋 

I think this pr is moving in wrong direction 😅 The solution that I had in mind was to have two loggers - 
`logger = logging.getLogger('internal_logger')` and `logger = logging.getLogger('application_logger')`. Use one of them for robyn internals and expose the other one outside for application logging. And when we talk about disabling the logger, we should provide two options of disabling the internal logger or disabling the application logger.

I hope that made sense. Do let me know if you require any further information.

sansyrox (commented):
Thank you for taking a swing at this, @mayachidambaram — making Robyn's logs opt-in/opt-out (#482) is a reasonable goal. As you noted, this was left unfinished, and it's drifted significantly: it conflicts with main and touches a lot of unrelated files (`Cargo.lock`, README, `.python-version`, and the integration test handlers, which would break the test suite). There are also some correctness issues I'd want fixed in any revival — on the Python side `debug()` no longer logs even when enabled, and on the Rust side `env_logger::init()` is called per-message (it panics on the second call) while real `error!` logs get suppressed behind the env var. Since main now offers `--log-level` (and `--fast` defaults to `WARNING`) which covers much of this, I think the cleanest path is to close this draft; if you'd like to pursue a focused, smaller version layered on the current `--log-level` system, I'd be glad to help scope it. Really appreciate the contribution either way.

PR #578, opened 2023-08-08
feat(lib.rs): add mimalloc as the global allocator

sansyrox:
**Description**

Add a new allocator, this will result in a performance boost

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -31,10 +31,10 @@ body:
     attributes:
       label: Your operating system
       options:
-      - Windows
-      - MacOS
-      - Linux
-      - Other (specify below)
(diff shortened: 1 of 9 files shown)


sansyrox:
This issue will be fixed when we retire python 3.7

banipreetr:
Interesting @sansyrox ! I am learning more about allocators and I'd be interested to know what motivated you to choose mimalloc over default system allocator? I see there are some other allocators as well (for example: tcmalloc, etc.). Why do we choose mimalloc here specifically?

Thanks!

sansyrox:
Hey @banipreetr 👋 

Mimalloc offers a better performance in general cases. See here https://www.microsoft.com/en-us/research/uploads/prod/2019/06/mimalloc-tr-v1.pdf

We will have to properly integrate it here and benchmark the benefits it brings for us 😄

PR #601, opened 2023-09-07
ci: drop sun and ibm support

sansyrox:
**Description**

This PR drops sun and ibm support


sansyrox:
not even sure why we kept them for so long 🤷‍♂️

PR #603, opened 2023-09-12
chore: decrease .whl filesize

IdoKendo:
**Description**

This PR reduces the .whl file size to 2.2MB on release mode. This is about 30% smaller than current build size which is 3.4MB - I have not seen any performance degradation or issues arise from these changes so it's a net gain by tweaking these values.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "robyn"
 version = "0.40.0"
 authors = ["Sanskar Jethi <sansyrox@gmail.com>"]
-edition = "2018"
+edition = "2021"
 description = "A web server that is fast!"
 license = "BSD License (BSD)"
 homepage = "https://github.com/sansyrox/robyn"
@@ -38,6 +38,11 @@ serde_json = "1.0.104"
 [features]
 io-uring = ["actix-web/experimental-io-uring"]
 
+[profile.release]
+codegen-units = 1
+lto = true
+panic = "abort"
+strip = true
 
 [package.metadata.maturin]
 name = "robyn"


Cargo.toml
```diff
@@ -38,6 +38,11 @@ serde_json = "1.0.104"
 [features]
 io-uring = ["actix-web/experimental-io-uring"]
 
+[profile.release]
+codegen-units = 1
+lto = true
+panic = "abort"
```
sansyrox, line 44: Hey @IdoKendo :wave: 


Thank you for the great work :smile: Can you please share some documentation of these three flags? Or explain what is the purpose of these three flags?

sansyrox: codegen-units - https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

sansyrox: @IdoKendo , what are the other two here?

IdoKendo: 1. By default, Cargo instructs compilation units to be compiled and optimized in isolation. LTO instructs the linker to optimize at the link stage. This can, for example, remove dead code and often times reduces binary size.
2. By default, when Rust code encounters a situation when it must call panic!(), it unwinds the stack and produces a helpful backtrace. The unwinding code, however, does require extra binary size. rustc can be instructed to abort immediately rather than unwind, which removes the need for this extra unwinding code.
ref: https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles

sansyrox (approved):
Amazing. Great work @IdoKendo :fire:

PR #609, opened 2023-09-18
feat: log routes when the router is populated

### LINKED ISSUE #599: Log the endpoints when app starts
I find it quite helpful to get a quick overview of what all endpoints are present in the application by just starting the application. For instance, if I can get the list of endpoints when I run the `test_server`. Also for projects with handlers defined in multiple files, it is very convenient to see all the endpoints in one place without going over the code. 

[NestJs](https://github.com/nestjs/nest) has this and it is quite helpful. 

Untitled

It seems fairly straightforward to implement this.

carlosm27:
**Description**
This PR is addressing the feature request of adding logs to the endpoints when the app starts.

image

This PR fixes #599


robyn/__init__.py
```diff
@@ -173,6 +173,16 @@ def start(self, url: str = "127.0.0.1", port: int = 8080):
         logger.info(f"Robyn version: {__version__}")
         logger.info(f"Starting server at {url}:{port}")
 
+        endpoint = {}
+
+        routes = self.router.get_routes()
```
IdoKendo, line 178: I think we should write the logs in the add_route method, instead of doing the get_routes here - it's minor, but it slows down the startup time

IdoKendo (changes requested):

IdoKendo:
Hey @carlosm27 👋 Great work! Added a few comments, let me know if something is unclear!

robyn/__init__.py
```diff
@@ -173,6 +173,16 @@ def start(self, url: str = "127.0.0.1", port: int = 8080):
...
+
+        for route in routes:
```
sansyrox, line 180: Hey @carlosm27 :wave: 

Thank you for the PR :smile: 
We should be using something like this

```suggestion
        for method, route in routes:
                logger.info("Logging endpoint: {method: %s, route: %s}", method, route)
```

As the logging library is optimized for deferred string evaluation https://docs.python.org/3/howto/logging.html#optimization

sansyrox: Thank you for the reviews @IdoKendo :smile:

carlosm27:
Thank you both @IdoKendo  and @sansyrox  for their reviews and suggestions. I'm going to apply them.

carlosm27:
I tried this approach:
```
logger.info("Logging endpoint: {method: %s, route: %s}", method, route)
```
But I receive the following error message:

image

I used f-string instead for the logging:

```
logger.info(f"Logging endpoint: method: {route[0]}, route: {route[1]}")
```

robyn/__init__.py
```diff
@@ -99,6 +99,11 @@ def add_route(
             }
             route_type = http_methods[route_type]
 
+        routes = self.router.get_routes()
+
+        for route in routes:
```
IdoKendo, line 104: You don't need to do the for loop here, this function runs for every route separately and has the endpoint and method as input parameters.

IdoKendo (changes requested):

robyn/__init__.py
carlosm27: Sorry, I misunderstand when you said to write the logs in `add_route` method. I tried this without using a for loop, but it shows one route when running the `base_routes.py` file. What I'm trying to do is that it shows all the routes in the list of routes.
```
if len(routes) > 2:
            logger.info(f"Logging endpoint: method={routes[0][0]}, route={routes[0][1]}")
```
But it just shows the logging of the one route when a I run the `base_routes.py` file
image

carlosm27: My bad. Sorry, yes I misunderstand.

```
logger.info(f"Logging endpoint: method={route_type}, route={endpoint}")

```
image

sansyrox:
> logger.info("Logging endpoint: {method: %s, route: %s}", method, route)

@carlosm27 maybe try this `Logging endpoint: method: %s, route: %s`

carlosm27:
> > logger.info("Logging endpoint: {method: %s, route: %s}", method, route)
> 
> @carlosm27 maybe try this `Logging endpoint: method: %s, route: %s`

Hey @sansyrox  I tried this: `logger.info("Logging endpoint: method: %s, route: %s", route_type, endpoint)`. But I got the same error message.

image

sansyrox:
@carlosm27 , the issue is not with your code but adding that is breaking something else. I can have a look tomorrow 😄

sansyrox:
@carlosm27 , I understand what you mean. Our abstraction is not the best right now. We can merge your PR and address that later.

Thank you for the great work! 😄

carlosm27:
> @carlosm27 , I understand what you mean. Our abstraction is not the best right now. We can merge your PR and address that later.
> 
> Thank you for the great work! 😄

Hey @sansyrox thank you for your review and feedback.

Issue #515, opened 2023-05-19
Add a way to trigger `send` in web sockets.


sansyrox:
We need to be able to expose the WebSocket object in every route.

Issue #181, opened 2022-04-03
[Feature Request] Add a way to "send" in a websocket 


sansyrox:
**Current Behavior**
No way to explicitly send something on a  websocket uniquely.

Either use a router with an id or find a way to send a socket "self" directly in the function.

And impl a method for that function.

**Desired Behavior**
This needs to be done.

---
**Screenshots / Mockups**

**Alternatives**

---

Issue #175, opened 2022-03-24
Add an ability to notify websockets through other websockets/functions.


sansyrox:
**Current Behavior**
Right now, we are returning an id. This helps us uniquely identify a websocket connection. But I would like to associate an object with the uuid, which would allow you notify the ws reciever.

**Desired Behavior**

I want to create and make the websocket object accessible like the request object and have a notify callback present in that. 

This means that the next websocket update will be a breaking change.

The current websocket events will be changed to lifecycle events. Analogous to global middlewares in requests.

And each one of them will have a websocket object that will be used send and receive events.

### With this issue and the previous one, I am manifesting a refactor in robyn codebase. A task to do this weekend.

PR #620, opened 2023-09-30
fix: websockets

### LINKED ISSUE #87: Add web sockets
https://actix.rs/docs/websockets/

Read the docs here.

sansyrox:
**Description**

This PR fixes #87 #135 #175 #181 #515


sansyrox:
We are sticking with the callback model as of now, as it is something I enjoy. Will observe for a few months to see if we need to consider the `app.route` model like other frameworks.

sansyrox:
make the rust python functions start with `_`. 

verify how the broadcast deals with different endpoints.

PR #625, opened 2023-10-03
fix: pass response headers from app to router

### LINKED ISSUE #605: add_response_header may fail sometimes
### Bug Description

```python
from robyn import Robyn, jsonify

app = Robyn(__file__)


@app.get("/")
async def h(request):
    return jsonify({"chinese": "中文"})


app.add_response_header("content-type", "application/json; charset=UTF-8")
app.start()
```

when refresh chrome, the output display two results:

`{"chinese":"中文"}`
or
`{"chinese":"涓枃"}`

so, I check it in DevTools,  the `Content-Type` change sometimes

`application/json; charset=UTF-8`
or
`text/plain`

### Steps to Reproduce

_No response_

### Your operating system

Windows

3.8

### Your Robyn version

None

### Additional Info

INFO:robyn.logger:Robyn version: 0.37.0

Python 3.8.3

IdoKendo:
**Description**

This PR fixes #605

There was a conflict of Content-Type between what is defined in `app.response_headers` and having it being overridden in the `Router` class. I am unsure why this caused them to sometimes return `text/plain` and sometimes return the value defined in `app.response_headers`, but this fix guarantees that `app.response_headers` is the header that will be set in the `Response`.

### DIFF
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -57,7 +57,7 @@ class Router(BaseRouter):
             headers = res.get("headers", headers)
             description = res.get("description", "")
 
-            if type(status_code) != int:
+            if not isinstance(status_code, int):
                 status_code = int(status_code)  # status_code can potentially be string
 
             response = Response(
@@ -91,8 +91,7 @@ class Router(BaseRouter):
         exception_handler: Optional[Callable],
         default_response_headers: List[Header],
     ) -> Union[Callable, CoroutineType]:
-        headers_list = (h.as_list() for h in default_response_headers)
-        response_headers = {d[0]: d[1] for d in headers_list}
+        response_headers = {d.key: d.val for d in default_response_headers}
 
         @wraps(handler)
         async def async_inner_handler(*args):
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -91,9 +91,8 @@ class Router(BaseRouter):
         exception_handler: Optional[Callable],
         default_response_headers: List[Header],
     ) -> Union[Callable, CoroutineType]:
-        response_headers = {
-            d[0]: d[1] for d in (h.as_list() for h in default_response_headers)
-        }
+        headers_list = (h.as_list() for h in default_response_headers)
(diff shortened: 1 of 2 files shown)


robyn/router.py
```diff
@@ -79,25 +89,42 @@ def add_route(
         handler: Callable,
         is_const: bool,
         exception_handler: Optional[Callable],
+        default_response_headers: List[Header],
     ) -> Union[Callable, CoroutineType]:
+        response_headers = {
+            d[0]: d[1] for d in (h.as_list() for h in default_response_headers)
+        }
```
sansyrox, line 96: Hey @IdoKendo 👋 

Great work as usual 😄  However, can we do something else instead of nested interpolation?

IdoKendo: Hey @sansyrox 👋 
Haha, yeah I was also not very satisfied with this line, but honestly the `Header` API is kinda awkward, so it's a bit tricky to do it any other way without introducing tons of boilerplate. I can try to fiddle with it so it's a bit nicer, or we can open a separate issue for it. WDYT?

sansyrox: @IdoKendo , let us just use two for loops for now. 

> I can try to fiddle with it so it's a bit nicer, or we can open a separate issue for it. WDYT?

I think we should definitely investigate 😄 But I will not consider it a blocker for this PR

robyn/router.py
```diff
@@ -79,25 +89,41 @@ def add_route(
         handler: Callable,
         is_const: bool,
         exception_handler: Optional[Callable],
+        default_response_headers: List[Header],
     ) -> Union[Callable, CoroutineType]:
+        headers_list = (h.as_list() for h in default_response_headers)
+        response_headers = {d[0]: d[1] for d in headers_list}
```
sansyrox, line 95:
```suggestion
        response_headers = {h.key: h.val for h in default_response_headers}
```

Hey @IdoKendo 👋 

Can we do this instead?

sansyrox (approved):
LGTM! 🔥 Great work @IdoKendo ✨

PR #629, opened 2023-10-04
feat: add request.json() method

### LINKED ISSUE #612: add a request.json field
having to use json.dumps all the time is not very friendly. add a way to access the request body in a more human friendly manner.

Maybe
Make this the default behavior and the string and the byte string the fall back behaviour

IdoKendo:
**Description**

This PR fixes #612 by adding a `request.json()` method, for example:
```py
@app.post("/hello")
def hello(request: Request):
    j = request.json()
    assert isinstance(j, dict)
```
Passing `'{"hello": "world"}'` would be a valid result, note that passing a non-json value e.g. `'Hello, World!'` would result in a ValueError and would require handling.

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -527,6 +527,25 @@ async def async_body_post(request: Request):
     return request.body
 
 
+# JSON Request
+
+
+@app.post("/sync/request_json")
+def sync_json_post(request: Request):
+    try:
+        return type(request.json())
+    except ValueError:
+        return None
+
+
+@app.post("/async/request_json")
+async def async_json_post(request: Request):
+    try:
+        return type(request.json())
+    except ValueError:
+        return None
+
+
 # --- PUT ---
 
 # dict
--- a/robyn/robyn.pyi
+++ b/robyn/robyn.pyi
@@ -87,7 +87,7 @@ class Request:
     def json(self) -> dict:
         """
         If the body is a valid JSON this will return the parsed JSON data.
-        Otherwise, will throw a ValueError.
+        Otherwise, this will throw a ValueError.
         """
         pass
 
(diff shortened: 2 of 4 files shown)


IdoKendo:
> oh, wow! This is great! I will have a proper look tomorrow morning. But great job @IdoKendo 😄 

Thanks @sansyrox 😃 Hacktoberfest has got me really motivated 🚀

integration_tests/base_routes.py
```diff
@@ -527,6 +527,25 @@ async def async_body_post(request: Request):
     return request.body
 
 
+# JSON Request
+
+
+@app.post("/sync/request_json")
+def sync_json_post(request: Request):
+    try:
+        return type(request.json())
```
sansyrox, line 536: Hey @IdoKendo 👋 

Thank you for the PR 😄 

But can we do `request.body.json()` instead of `request.json()`? In my mind `request.json()` returns a json version of the request object and not request.body

IdoKendo: I don't think we can do it request.body.json() since body is a string/bytes and not an object we instantiate - Anyway I don't think request.json() implies the request object. That's how it works in both Flask's request object as well.

sansyrox: That does make sense. Sounds good. Just one last suggestion for the docstring and then we can merge it :D

robyn/robyn.pyi
```diff
@@ -84,6 +84,9 @@ class Request:
     ip_addr: Optional[str]
     identity: Optional[Identity]
 
+    def json(self) -> dict:
```
sansyrox, line 87: Can we add a docstring here please?

IdoKendo: Added 🔥

sansyrox (approved):
LGTM 🔥

PR #630, opened 2023-10-06
feat: add Request and Url objects constructors to enable easier testing

### LINKED ISSUE #574: How can i import the Url from robyn?
### Bug Description

I got an exception message if I want to import the Url: "AttributeError: module 'robyn.robyn' has no attribute 'Url'."

### Steps to Reproduce

```python
from robyn.robyn import Url
```

I try to use dir() to see what properties this class has.

```python
dir(robyn.robyn )
```

```python
['FunctionInfo', 'HttpMethod', 'Identity', 'MiddlewareType', 'Request', 'Response', 'Server', 'SocketHeld', '__all__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'get_version']
```

So how do I import or create a new Url object?

### Your operating system

MacOS

3.10

### Your Robyn version

latest

### Additional Info

_No response_

IdoKendo:
**Description**

This PR fixes #574 by exposing Url and Request's `new` methods from rust to the python API, thus allowing developers to create them in tests in order to build unit tests that don't require starting a server.


integration_tests/test_request_object.py
sansyrox, line 1: Hey @IdoKendo 👋 

Thank you for the great work 😄 

I think this test should be a part of the `unit_tests` folder. This doesn't look like an integration test to me.

src/types/request.rs
sansyrox, line 1: Nicee

integration_tests/test_request_object.py
IdoKendo: You're right 😄

sansyrox:
I love how we are improving testability. GG @IdoKendo 🔥

IdoKendo:
> I love how we are improving testability. GG @IdoKendo 🔥

Yeah, it's really important IMO! By the way, I'm noticing the macOS integration tests are a bit flaky... Any idea what could be the cause?

sansyrox:
@IdoKendo , a good solution would be not to depend on network calls this much and to have a proper testing client 😅 

But one reason could be that we are spawning and killing multiple processes very frequently.

IdoKendo:
> @IdoKendo , a good solution would be not to depend on network calls this much and to have a proper testing client 😅
> 
> But one reason could be that we are spawning and killing multiple processes very frequently.

Yeah, I guess once the testing client PR will be merged we can create a more resilient testing suite.

sansyrox (approved):
LGTM 😄

PR #638, opened 2023-10-08
feat: Add `--verbose` flag

### LINKED ISSUE #611: fix log level!
current log level is too noisy. this should be a part of --verbose flag and the global log level should control the logging of robyn

carlosm27:
**Description**
This PR address the addition of a `verbose` flag to control the logs in Robyn.

As default the console looks like this when running Robyn:
image

When the `--verbose` flag is passed:
image

This PR fixes #611

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 23.10.0
     hooks:
       - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.292
+    rev: v0.1.1
     hooks:
       - id: ruff
 


robyn/__init__.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, file_object: str, config: Config = Config()) -> None:
         load_vars(project_root=directory_path)
         logging.basicConfig(level=self.config.log_level)
 
-        if self.config.log_level.lower() != "warn":
+        if self.config.verbose is True:
             logger.info(
                 "SERVER IS RUNNING IN VERBOSE/DEBUG MODE. Set --log-level to WARN to run in production mode.",
```
IdoKendo, line 49: I think this message is a bit out of place after changing the if statement.

robyn/__init__.py
```diff
@@ -99,7 +99,8 @@ def add_route(
             }
             route_type = http_methods[route_type]
 
-        logger.info("Logging endpoint: method=%s, route=%s", route_type, endpoint)
+        if self.config.verbose is True:
+            logger.info(f"Logging endpoint: method={route_type}, route={endpoint}")
```
IdoKendo, line 103: Please use lazy evaluation as it appeared before the change.

robyn/__init__.py
```diff
@@ -148,7 +149,9 @@ def add_web_socket(self, endpoint: str, ws: WS) -> None:
         self.web_socket_router.add_route(endpoint, ws)
 
     def _add_event_handler(self, event_type: Events, handler: Callable) -> None:
-        logger.info("Add event %s handler", event_type)
+        if self.config.verbose is True:
+            logger.info(f"Add event {event_type} handler")
```
IdoKendo, line 153: Please use lazy evaluation here too.

IdoKendo (changes requested):
Hey @carlosm27 👋 Great work, left a few comments for minor fixes.

carlosm27:
Hey @IdoKendo thank you for reviewing my PR.

robyn/__init__.py
carlosm27: I don't understand what do you mean here. Should I remove the message, the if statement?

IdoKendo: Sorry, I probably wasn't clear! Previously the message would appear when the log_level is not `WARN`, and the message says to set the log-level to `WARN` on production, so on production it wouldn't appear. After the change, even if the log-level is `WARN` - but verbose is truthy, then the message would appear. Which then doesn't make sense.

Since `--verbose` is now very... well... verbose 😅  I feel that the message is redundant. If someone runs on production with `--verbose` flag, they probably do it purposefully. WDYT?

sansyrox:
Hey @carlosm27 👋 

Apologies for the delayed review and thank you @IdoKendo  for the reviews 😄  The idea for this issue was to have two loggers with different names.

One will be for robyn users and the other one will be for internal. The idea of (-vv or --very-verbose) logs was to enable all the logs and so (-v or --verbose) should only enable the internal logger.

the --log-level flag should in turn be used for the logger for robyn users. I hope this makes sense

carlosm27:
> Hey @carlosm27 👋
> 
> Apologies for the delayed review and thank you @IdoKendo for the reviews 😄 The idea for this issue was to have two loggers with different names.
> 
> One will be for robyn users and the other one will be for internal. The idea of (-vv or --very-verbose) logs was to enable all the logs and so (-v or --verbose) should only enable the internal logger.
> 
> the --log-level flag should in turn be used for the logger for robyn users. I hope this makes sense

Hey @sansyrox,. Yes, it makes sense. Can you tell me which logs should be internal?

sansyrox:
Hey @carlosm27 👋 

Let us make all the rust logs internal to start.

carlosm27:
Dear God, I'm sorry.  I was syncing my branch. And don't know why the PR closed. I apologize for this inconvenient. I hope this mistake doesn't cause any trouble. Please let me know what I can do to amend this mistake.

PR #650, opened 2023-10-16
docs: Docs suggestions

elliotwaite:
Here are some suggestions for the docs that I thought of while reading over them. I'm not sure if you'll want to use all of them, so feel cherry-pick the ones you want.

Also, some of the markdown code sections are displaying strangely on these pages:
* http://www.sparckles.cloud/Robyn/#/features (the part after "You can serve views in two ways:")
* http://www.sparckles.cloud/Robyn/#/plugins

And this page has a line "export const description = ..." at the top that I'm not sure if it's supposed to be there:
* http://www.sparckles.cloud/Robyn/#/comparison

Also, on the docs site, when you click a link in the left nav, it doesn’t scroll to the top of the new page, which is strange when you click a link after having scrolled to the bottom of the previous page.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -2,10 +2,10 @@
 
 Robyn is a Python web server that uses the tokio runtime.
 
-First of all, we have a worker event cycle that basically does all the dirty work. This part manages the runtime and passes all instructions to the rust code.
+First of all, we have a worker event cycle that basically does all the dirty work. This part manages the runtime and passes all instructions to the Rust code.
 This spawns the threading pool
 
-Then when we type the command `python3 app.py` the python code is converted to rust objects and then the router is populated. The incoming requests hit the router and then the rust objects are dispatched in the thread pool and executed depending on their types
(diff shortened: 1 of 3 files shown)


sansyrox:
Thank you for the pr @elliotwaite 😊
Your changes look great from the first glance. I will give a thorough review after waking up tomorrow 

We are migrating to a new docs website - https://beta.robyn.tech

You can refer to the new docs website till then 😊

elliotwaite:
Sounds good. Oh, wow. Those new docs look nice. I'll check them out.

Another thing I was going to suggest, although I'm less certain about this one, is that to me,  all the `h()` functions in the docs seemed odd. Does the "h" stand for "home" or "hello" or something else? Maybe changing them to `home()` or `hello()` would make them more self-documenting. But that might just be me. Feel free to leave them as the more concise `h()` if that's your style.

sansyrox:
All of these will be fixed in the new docs! I just need to finish this https://github.com/sparckles/Robyn/issues/646 before I complete migration 😅

sansyrox (approved):
LGTM! Thank you 😄

PR #651, opened 2023-10-17
docs: New docs fixes

elliotwaite:
Some small fixes for the new docs. Let me know if any of them are incorrect.

Also, on mobile I am unable to click the last link in the left nav panel for "Upcoming Features". This is because the bottom "X" button is in a div that spans the full width of that nav area and covers that bottom link:
image
 
It seems like a good fix would be to make it so the div containing the "X" button was only as wide as the "X" button so that users could still click links in that bottom area to the right of that button. It might also be good to add some extra bottom padding to that nav area so that when users scrolled to the bottom of that nav list, the last item would be shown above that "X" button.

### DIFF
--- a/new_docs/src/pages/documentation/api_reference/getting_started.mdx
+++ b/new_docs/src/pages/documentation/api_reference/getting_started.mdx
@@ -96,7 +96,7 @@ For example, Batman learned to create a POST request like this:
     from robyn import Request
 
     @app.post("/")
-    async def h(request):
+    async def h(request: Request):
         return "Hello World"
 
     ```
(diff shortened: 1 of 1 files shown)


sansyrox:
Thanks for the docs update @elliotwaite 😄 They LGTM 😄 

> It seems like a good fix would be to make it so the div containing the "X" button was only as wide as the "X" button so that users could still click links in that bottom area to the right of that button. It might also be good to add some extra bottom padding to that nav area so that when users scrolled to the bottom of that nav list, the last item would be shown above that "X" button.

Yes, these are good solutions! I will try implementing them tomorrow :D

elliotwaite:
Sounds good.

sansyrox (approved):

PR #677, opened 2023-11-01
Changing from bash to python for authentication script

jeasoft:
**Description**
The script seems to be like a Bash script but should be seem as a Python Script. 

This PR fixes #


sansyrox (approved):
Lgtm! Thank you @jeasoft 😊😊

PR #691, opened 2023-11-11
feat: Add `--verbose` flag 

### LINKED ISSUE #611: fix log level!

carlosm27:
**Description**
This PR add the `--verbose` flag and add the log level `Info APP` to hide the logging messages from Actix.
image

With the `--verbose` flag:
image

This PR fixes #611


robyn/__init__.py
```diff
@@ -32,6 +33,21 @@
...
+
+def info_app(self, message, *args, **kwargs):
+    self._log(INFO_APP, message, args, kwargs)
+
+
+logging.Logger.info_app = info_app  # Add method to Logger class
+
+logging.INFO_APP = INFO_APP  # Optional - for convenience
+
+logger = logging.getLogger(__name__)
```
sansyrox, line 48: Hey @carlosm27 👋 

Thank you for the PR 😄 Can you please explain what the `INFO_APP` parameter does?

carlosm27: Hey @sansyrox it creates a new log level because with the `info` level shows also the logs from Actix too. But, I just notice in the Logger documentation that it doesn't advice creating new logs levels because it can create conflicts with other libraries. So, I will continue working on this and making updates to receive their review.

shivaylamba: Hey @carlosm27 any further updates to this?

sansyrox (commented):
Thanks for this, @carlosm27 — quieting the Actix noise (#611) was a real pain point. Unfortunately this one has drifted a long way from current main: it patches the old pre-`BaseRobyn` `__init__.py` and an `argument_parser.py` that has since been rewritten, so it no longer applies. There are also a couple of bugs to flag — swapping our colored `logger` for a stdlib `logging.getLogger` while still passing `color=Colors.BLUE` will raise a `TypeError`, and the "VERBOSE MODE" message ends up printing unconditionally. Since main now handles this through `--log-level` and `--fast` (which sets `WARNING`), I'd lean toward closing this in favor of a fresh, smaller patch if you still want a dedicated `--verbose` toggle. I'm grateful for the effort and happy to guide a rebased version.

PR #701, opened 2023-11-25
Fixed #605 custom content-type setting not effected sometimes

### LINKED ISSUE #605: add_response_header may fail sometimes

18o:
**Description**

This PR fixes #605 

`global_response_headers` in `Server` is a DashMap, which is case sensitive. 
The default key is `Content-Type` can be stay together with `content-type` in `global_response_headers`, so when it's extended , both `Content-Type` and `content-type` are written to http header without fixed sequence, but the browser is case insensitive, so it will take the last value in `Content-Type` and `content-type`, that's why this bug happened.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -262,8 +262,16 @@ impl Server {
     /// Adds a new response header to our concurrent hashmap
     /// this can be called after the server has started.
     pub fn add_response_header(&self, key: &str, value: &str) {
-        self.global_response_headers
-            .insert(key.to_string(), value.to_string());
+        match key.to_lowercase().as_str() {
+            "content-type" => {
+                self.global_response_headers
+                    .insert("Content-Type".to_str ing(), value.to_string());
+            }
+            _ => {
+                self.global_response_headers
+                    .insert(key.to_string(), value.to_string());
+            }
+        }
     }
 
     /// Removes a new request header to our concurrent hashmap
--- a/src/server.rs
+++ b/src/server.rs
@@ -262,16 +262,8 @@ impl Server {
     /// Adds a new response header to our concurrent hashmap
     /// this can be called after the server has started.
     pub fn add_response_header(&self, key: &str, value: &str) {
-        match key.to_lowercase().as_str() {
-            "content-type" => {
-                self.global_response_headers
-                    .insert("Content-Type".to_string(), value.to_string());
-            }
-            _ => {
-                self.global_response_headers
-                    .insert(key.to_string(), value.to_string());
-            }
-        }
+        self.global_response_headers
+            .insert(key.to_lowercase(), value.to_string());
     }
 
     /// Removes a new request header to our concurrent hashmap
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -262,8 +262,16 @@ impl Server {
     /// Adds a new response header to our concurrent hashmap
     /// this can be called after the server has started.
     pub fn add_response_header(&self, key: &str, value: &str) {
-        self.global_response_headers
-            .insert(key.to_string(), value.to_string());
+        match key.to_lowercase().as_str() {
```
sansyrox, line 265: Hey @18o 👋 

Thank you for the PR 😄 TIL about the web headers fact. Thank you ✨ 

However, I have one question. Why not make everything lowercase?

18o: It's a good question, I was only thought about the `content-type` thing.
Do you mean all keys in the `global_response_headers` make them into lowercase?

sansyrox: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

> An HTTP header consists of its case-insensitive name followed by a colon (:), then by its value. [Whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) before the value is ignored.

I read this at MDN docs today.

sansyrox: @18o , 
> Do you mean all keys in the global_response_headers make them into lowercase?

18o: Yeah, we can make the keys in `global_response_headers ` all uppercase or lowercase, in order to avoid issues like this.

sansyrox: @18o , sounds good. Let us make everything `lower_case` . I haven't seen someone mistype all upper case headers.

18o: ok, it's implemented in the latest commit

sansyrox:
hey @18o 👋 

Apologies for the delayed review here. I am working on a PR here - https://github.com/sparckles/Robyn/pull/704 , which might solve the issue at a more fundamental level.  I will merge it and hopefully, the issue should be fixed.

sansyrox:
Hey @18o 👋 

Thank you for the PR. However, there was a more fundamental flaw in the code. It has been fixed in v0.47.0 😄

PR #727, opened 2023-12-29
Fix cli script

patrick91:
This pr fixes the pyproject.toml to allow running `robyn` as 
a cli tool.

The configuration was almost correct, but poetry doesn't support
the standard declaration.

To test this you can do:

```bash
poetry run robyn
```

and it should also work when installing robyn (or via pipx run robyn)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,9 +33,6 @@ dependencies = [
   "uvloop~=0.19.0; sys_platform != 'win32' and platform_python_implementation == 'CPython' and platform_machine != 'armv7l'"
   ]
 
-[project.scripts]
-robyn = "robyn.cli:run"
-
 [project.optional-dependencies]
 "templating" = ["jinja2 == 3.0.1"]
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -33,9 +33,6 @@ dependencies = [
   "uvloop~=0.19.0; sys_platform != 'win32' and platform_python_implementation == 'CPython' and platform_machine != 'armv7l'"
   ]
 
-[project.scripts]
```
sansyrox, line 36: Hey @patrick91 👋 

Thank you for the PR 😄 . But turns out `[project.scripts]` key is the one required for `maturin` based projects. 

`[tool.poetry.scripts]` will only work for poetry-based publishing system.

patrick91: @sansyrox so then I'm not sure why it doesn't work, how do you build and publish the project?

sansyrox: @patrick91 , oh, it works now 😅 See this - https://github.com/sparckles/Robyn/releases/tag/v0.49.3

> how do you build and publish the project?

we use `maturin build` command with different flags based on the architecture and python version.

patrick91: what about publishing?

sansyrox: @patrick91 , we use `twine` https://github.com/sparckles/Robyn/blob/04291e9f558c2d2d995f98bab1a9e1440331150b/.github/workflows/release-CI.yml#L177

PR #753, opened 2024-02-20
Distutils is removed in Python 3.12 so switch to setuptools

cclauss:
**Description**

This PR should fix the failing GitHub Actions `linux-cross` on Python 3.12.

https://docs.python.org/3/whatsnew/3.12.html#distutils

@messense @konstin Do you have any hints on the [`linux-cross` failures on Py3.12](https://github.com/sparckles/Robyn/actions/runs/7969971919/job/21756682947?pr=753)?
Can `maturin` error logs be written to stderr instead of the tempfile?


sansyrox (approved):
LGTM!

PR #780, opened 2024-04-02
Add default None to QueryParams get fn

### LINKED ISSUE #763: query_params.get("param_name") requires a default value
### Bug Description

query_params.get("param_name") requires a default value

this default value should be optional and not manadataed

### Steps to Reproduce

_No response_

### Your operating system

None

None

### Your Robyn version

None

### Additional Info

_No response_

StandinKP:
**Description**

This PR fixes #763

### DIFF
--- a/src/types/multimap.rs
+++ b/src/types/multimap.rs
@@ -28,7 +28,10 @@ impl QueryParams {
     pub fn get(&self, key: String, default: Option<String>) -> Option<String> {
         match self.queries.get(&key) {
             Some(values) => values.last().cloned(),
-            None => default,
+            None => match default {
+                Some(val) => Some(val),
+                None => None,
+            },
         }
     }
 


StandinKP:
I am now matching the default as well to return either None or the default value. So if you don't pass the default param it will return the None. I can try to add a test case if we have some for testing the query params. I can see test_get_request and other similar tests. Do you want me to add it there?

sansyrox:
> I am now matching the default as well to return either None or the default value. So if you don't pass the default param it will return the None. I can try to add a test case if we have some for testing the query params. I can see test_get_request and other similar tests. Do you want me to add it there?

Yes, please do try it with test cases. As both of the diffs are doing the same things. Do give it a thought, if it still doesn't make a lot of sense then I will be happy to explain :D

StandinKP:
> Yes, please do try it with test cases. As both of the diffs are doing the same things. Do give it a thought, if it still doesn't make a lot of sense then I will be happy to explain :D

I just checked and you are right both of them do the same output. Both of them return None if we don't pass in the default value. So I have a question what exactly is wrong with the current setup?

sansyrox:
See this - https://github.com/sparckles/Robyn/pull/787

sansyrox:
Closing this, as it has been fixed.

PR #819, opened 2024-05-14
fix: trailing slash issue

### LINKED ISSUE #818: Trailing slashes leads to 404
### Bug Description

say, I have a route:

```
@app.get("/sync/str/")
def sync_str_get():
    return "sync str get"
```

now, when I do `curl http://localhost:8080/sync/str/`, everything works fine, but if I do a `curl http://localhost:8080/sync/str`, I get a "Not found".

this is the same case if it is vice versa too (slash not being present in the route & slash being present in the URL).

edit: I tried the same thing with flask, and that worked fine.

### Steps to Reproduce

_No response_

### Your operating system

Linux

3.12

### Your Robyn version

main branch

### Additional Info

ofc, I would like to work on this issue, and I have been debugging it, that led me to `const_router.Router#get_route`.

```
fn get_route(&self, route_method: &HttpMethod, route: &str) -> Option<Response> {
        let table = self.routes.get(route_method)?;
        let route_map = table.read().ok()?;

        match route_map.at(route) {
            Ok(res) => Some(res.value.clone()),
            Err(_) => None,
        }
    }
```

I presume I should add/remove a trailing slash here after checks. @sansyrox please CMIIW. :)

edit: or, we can add both the routes (with & without the slash) in `Server#add_route`, since doing the above approach would introduce an overhead on every request, but this approach only once. what say?

VishnuSanal:
**Description**

This PR fixes #818

@sansyrox PTAL :)

### DIFF
--- a/integration_tests/test_get_requests.py
+++ b/integration_tests/test_get_requests.py
@@ -1,4 +1,5 @@
 import pytest
+import requests
 from requests import Response
 
 from integration_tests.helpers.http_methods_helpers import get
@@ -51,3 +52,12 @@ def test_queries(function_type: str, session):
 
     r = get(f"/{function_type}/queries")
     assert r.json() == {}
+
+
+@pytest.mark.benchmark
+def test_trailing_slash(session):
+    r = requests.get("http://localhost:8080/trailing")
+    assert r.text == "Trailing slash test successful!"
+
+    r = requests.get("http://localhost:8080/trailing/")
+    assert r.text == "Trailing slash test successful!"
--- a/integration_tests/test_get_requests.py
+++ b/integration_tests/test_get_requests.py
@@ -56,7 +56,7 @@ def test_queries(function_type: str, session):
 
 @pytest.mark.benchmark
 def test_trailing_slash(session):
-    r = requests.get("http://localhost:8080/trailing")
+    r = requests.get("http://localhost:8080/trailing")  # `integration_tests#get` strips the trailing slash, tests always pass!`
     assert r.text == "Trailing slash test successful!"
 
     r = requests.get("http://localhost:8080/trailing/")
--- a/src/server.rs
+++ b/src/server.rs
@@ -281,7 +281,7 @@ impl Server {
         is_const: bool,
     ) {
 
-        let second_route: String = if route.ends_with("/") {
+        let second_route: String = if route.ends_with('/') {
              route[0..route.len() - 1].to_string()
         } else {
              format!("{}/", route)
--- a/src/server.rs
+++ b/src/server.rs
@@ -280,11 +280,10 @@ impl Server {
         function: FunctionInfo,
         is_const: bool,
     ) {
-
         let second_route: String = if route.ends_with('/') {
-             route[0..route.len() - 1].to_string()
+            route[0..route.len() - 1].to_string()
         } else {
-             format!("{}/", route)
+            format!("{}/", route)
         };
 
         self._add_route(py, route_type, route, function.clone(), is_const);
--- a/src/server.rs
+++ b/src/server.rs
@@ -289,8 +289,8 @@ impl Server {
              second_route = format!("{}/", route);
         };
 
-        self._add_route(py, route_type, route, &function, is_const);
-        self._add_route(py, route_type, &second_route, &function, is_const);
+        self._add_route(py, route_type, route, function.clone(), is_const);
+        self._add_route(py, route_type, &second_route, function, is_const);
     }
 
     fn _add_route(
@@ -298,7 +298,7 @@ impl Server {
         py: Python,
         route_type: &HttpMethod,
         route: &str,
-        function: &FunctionInfo,
+        function: FunctionInfo,
         is_const: bool,
     ) {
         debug!("Route added for {:?} {} ", route_type, route);
@@ -308,7 +308,7 @@ impl Server {
         if is_const {
             match self
                 .const_router
-                .add_route(route_type, route, function.clone(), Some(event_loop))
+                .add_route(route_type, route, function, Some(event_loop))
             {
                 Ok(_) => (),
                 Err(e) => {
@@ -316,7 +316,7 @@ impl Server {
                 }
             }
         } else {
-            match self.router.add_route(route_type, route, function.clone(), None) {
+            match self.router.add_route(route_type, route, function, None) {
                 Ok(_) => (),
                 Err(e) => {
                     debug!("Error adding route {}", e);
--- a/src/server.rs
+++ b/src/server.rs
@@ -281,16 +281,17 @@ impl Server {
         is_const: bool,
     ) {
 
-        self.save_route(py, route_type, route, &function, is_const);
-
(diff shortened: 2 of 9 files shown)


sansyrox:
Hey @VishnuSanal 👋 

Thanks for the PR, but I see no changes in the PR. Could have a look again?

VishnuSanal:
> Hey @VishnuSanal 👋
> 
> Thanks for the PR, but I see no changes in the PR. Could you have a look again?

ouch!! that lost somewhere whilst fixing the CI. :D PTAL now :)

src/server.rs
```diff
@@ -280,22 +280,40 @@ impl Server {
...
     ) {
+
+        self.save_route(py, route_type, route, &function, is_const);
+
+        if &route[route.len() - 1..] == "/" {
+            self.save_route(py, route_type, &route[0..route.len() - 1], &function, is_const);
+        } else {
+            self.save_route(py, route_type, &format!("{}/", route), &function, is_const);
+        }
+    }
```
sansyrox, line 291: @VishnuSanal , two suggestions
- https://doc.rust-lang.org/std/primitive.str.html#method.ends_with use this instead of route.len() - 1 and
- We should do this if check before the `self.save_route`

src/server.rs
```diff
@@ -280,22 +280,40 @@ impl Server {
...
+
+    pub fn save_route(
```
sansyrox, line 293: also, why rename this function?

src/server.rs
VishnuSanal: > We should do this if check before the self.save_route

why? 🤔

src/server.rs
VishnuSanal: > um... that's a new function. we're calling `#save_route` (the renamed function with the old code) twice from the new function with the old name (`#add_route`). I did it this way by renaming to avoid changing it's implementation from everywhere else.

src/server.rs
```diff
@@ -280,22 +280,41 @@ impl Server {
...
+    ) {
         debug!("Route added for {:?} {} ", route_type, route);
         let asyncio = py.import("asyncio").unwrap();
         let event_loop = asyncio.call_method0("get_event_loop").unwrap();
 
         if is_const {
             match self
                 .const_router
-                .add_route(route_type, route, function, Some(event_loop))
+                .add_route(route_type, route, function.clone(), Some(event_loop))
```
sansyrox, line 309: Why add clones here?

src/server.rs
```diff
@@ -280,22 +280,41 @@ impl Server {
...
+        self._add_route(py, route_type, route, &function, is_const);
+        self._add_route(py, route_type, second_route, &function, is_const);
+    }
+
+    fn _add_route(
+        &self,
+        py: Python,
+        route_type: &HttpMethod,
+        route: &str,
+        function: &FunctionInfo,
```
sansyrox, line 299:
```suggestion
        function: FunctionInfo,
```

Try this and remove the clones?

sansyrox:
Hey @VishnuSanal 

I have made some changes and have added a few suggestions. Do let me know what you think 😄

src/server.rs
sansyrox: nvm. I have phrased it in a better way through code 😅

src/server.rs
VishnuSanal: you'd get a compiler error:

```
error[E0382]: use of moved value: `function`
   --> src/server.rs:289:69
    |
280 |         function: FunctionInfo,
    |         -------- move occurs because `function` has type `FunctionInfo`, which does not implement the `Copy` trait
...
284 |         self.save_route(py, route_type, route, function, is_const);
    |                                                -------- value moved here
...
289 |             self.save_route(py, route_type, &format!("{}/", route), function, is_const);
    |                                                                     ^^^^^^^^ value used here after move
    |
note: consider changing this parameter type in method `save_route` to borrow instead if owning the value isn't necessary
   --> src/server.rs:298:19
    |
293 |     pub fn save_route(
    |            ---------- in this method
...
298 |         function: FunctionInfo,
    |                   ^^^^^^^^^^^^ this parameter takes ownership of the value
help: consider cloning the value if the performance cost is acceptable
    |
284 |         self.save_route(py, route_type, route, function.clone(), is_const);
    |                                                        ++++++++

For more information about this error, try `rustc --explain E0382`.
```

src/server.rs
VishnuSanal: to pass `FunctionInfo` from `&FunctionInfo`, either I should clone here, or before (in `#add_route`). I checked whether we can pass it directly, but we can't we would want to do `#clone` on `Router#add_route`.

sansyrox:
Also, @VishnuSanal , please add an integration test. Add a route in `integration_tests/base_routes.py` and then make a curl request on both the routes like this - https://github.com/sparckles/Robyn/blob/main/integration_tests/test_get_requests.py

VishnuSanal:
> please add an integration test. Add a route in `integration_tests/base_routes.py` and then make a curl request on both the routes like this - https://github.com/sparckles/Robyn/blob/main/integration_tests/test_get_requests.py

@sansyrox why add a new route? also, why `integration_tests/test_get_requests.py`? wouldn't modifying `integration_tests/test_basic_routes.py` like [this](https://github.com/sparckles/Robyn/compare/main...VishnuSanal:Robyn:test_sample#diff-71cc145cea26f53de5d4cfb3ce3d969dfb42d32ed15fc47e4cd643f5f526dcfa) suffice? 🤔

sansyrox:
@VishnuSanal , we tend to add one route for its function. If you explore the file, you will find out that we have one dedicated route for get requests, one for posts, etc.

> why add a new route? also, why integration_tests/test_get_requests.py? wouldn't modifying integration_tests/test_basic_routes.py like

integration_tests/test_get_requests.py
```diff
@@ -51,3 +52,12 @@ def test_queries(function_type: str, session):
 
     r = get(f"/{function_type}/queries")
     assert r.json() == {}
+
+
+@pytest.mark.benchmark
+def test_trailing_slash(session):
+    r = requests.get("http://localhost:8080/trailing")
```
sansyrox, line 59: Hey @VishnuSanal , instead of `requests.get`, you can use the `get` method we have imported from `from integration_tests.helpers.http_methods_helpers import get`

VishnuSanal: @sansyrox That strips the trailing slash. Tests always pass!

sansyrox: Ah, okay. That makes sense. Could you please comment at line 59? And then I will do a final review before merge 😄

sansyrox (approved):
LGTM! Great work @VishnuSanal ✨

sansyrox:
@VishnuSanal , the ci is failing on formatting. Can you please fix that?

sansyrox:
All looks good to merge otherwise 😄

sansyrox:
Great work @VishnuSanal 😄

Issue #835, opened 2024-06-03
Error in Crime Data Endpoints Documentation Code


Amaljyothi44:
### Bug Description

There is an issue with the get_crimes endpoint in the crime data documentation. The code provided is throwing a TypeError when attempting to retrieve query parameters.
```
@app.get("/crimes")
async def get_crimes(request):
    with SessionLocal() as db:
        skip = request.query_params.get("skip", 0)
        limit = request.query_params.get("limit", 100)
        crimes = crud.get_crimes(db, skip=skip, limit=limit)
    return crimes

**Error:**
skip = request.query_params.get("skip", 0)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument 'default': 'int' object cannot be converted to 'PyString'
```

### Your operating system

None

### Your Python version (`python --version`)

Python 3.11.3

### Your Robyn version

robyn==0.54.5

### Additional Info

**Corrected Code:**
```
@app.get("/crimes")
async def get_crimes(request):
    with SessionLocal() as db:  # Use SessionLocal here
        skip = request.query_params.get("skip", "0") 
        limit = request.query_params.get("limit", "100")
        skip = int(skip)
        limit = int(limit)
        crimes = crud.get_crimes(db, skip=skip, limit=limit)
    return crimes
```    
@sansyrox please CMIIW

sansyrox:
Hey @Amaljyothi44 👋 

Thanks for raising the issue. Could you confirm your robyn version?

Amaljyothi44:
@sansyrox I have updated the version

sansyrox:
Hey @Amaljyothi44 👋 

One suggestion - please format code snippets using backticks(\`) . But apart from that yes, this is definitely an issue. We could fix it like your suggestion, but maybe we can add a `cast_to` in the request.query_params, which takes a python function callable and tries to cast it to data type.
( https://github.com/sparckles/Robyn/issues/836 )

e.g.

```
@app.get("/crimes")
async def get_crimes(request):
    with SessionLocal() as db:  # Use SessionLocal here
        skip = request.query_params.get("skip",cast_to=int, default=0) 
        limit = request.query_params.get("limit", cast_to=int, default=100)
        skip = int(skip)
        limit = int(limit)
        crimes = crud.get_crimes(db, skip=skip, limit=limit)
    return crimes
```

What do you think of this approach?

Amaljyothi44:
[...]

PR #842, opened 2024-06-07
cast_to argument in query_params.get

Amaljyothi44:
Add a `cast_to` argument to the query_params.get method for bug #835


sansyrox:
Hey @Amaljyothi44 👋 

Thanks for the PR. You have changed the `pyi` files, which is used just for typing and not adding the business logic. You need to make changes here (https://github.com/sparckles/Robyn/blob/3644f21ce9984758fdf9b0dc9e03ea7094118fe1/src/types/multimap.rs#L9)

PR #847, opened 2024-06-10
feat: Add custom exceptions

### LINKED ISSUE #550: Add custom exceptions in Robyn
There are many exceptions in Web that are recurring. We should be exposing them so that the developers don't have redo the task again and again

Mr-Sunglasses:
**Description**

This PR fixes #550


Mr-Sunglasses:
@sansyrox I've implemented  custom `HTTPException`, take a look and please tell me if it is what we want to achieve in #550  😄

robyn/__init__.py
```diff
@@ -472,6 +473,7 @@ def ALLOW_CORS(app: Robyn, origins: List[str]):
     "SubRouter",
     "AuthenticationHandler",
     "Headers",
+    "HTTPException",
```
sansyrox, line 476: Hey @Mr-Sunglasses , could you add “WebSocket” exception too?

robyn/exceptions.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+import http
```
sansyrox, line 3: Are we using this anywhere?

robyn/reloader.py
sansyrox, line 1: This file is not needed

robyn/robyn.pyi
```diff
@@ -107,7 +107,6 @@ class QueryParams:
             value (str): The value of the query parameter
```
sansyrox, line 107: Same for this

robyn/ws.py
```diff
@@ -20,7 +20,13 @@ class WebSocket:
     # should this be websocket router?
```
sansyrox, line 20: Same for this

sansyrox (commented):
Hey @Mr-Sunglasses ,

Thanks for the PR. I have some suggestions :D

Mr-Sunglasses:
> Hey @Mr-Sunglasses ,
> 
> Thanks for the PR. I have some suggestions :D

Thanks, @sansyrox 🙏🏻, for the suggestions, I've fixed then 😄

sansyrox:
Hey @Mr-Sunglasses 👋 

Could you please add some integration tests? And the CI is also failing for some reason.

sansyrox (commented):
Thanks for this, @Mr-Sunglasses, and apologies it sat so long. The landscape has shifted since you opened it: `robyn/exceptions.py` now already lives in main (it came in with the CORS PR #1019), which is why this is conflicting. That said, you've spotted a real gap — those exceptions still aren't exported from the top-level package, so `from robyn import HTTPException` doesn't actually work today. The most useful path forward would be to rebase and trim this down to just the export wiring in `__init__.py` (adding the import + the `__all__` entries) plus your nice docs example, dropping the duplicate `exceptions.py`. Heads up that this also overlaps with #951, so let's coordinate to avoid double work — and a small test in `test_exception_handling.py` would seal it.

PR #849, opened 2024-06-12
feat: prompt for a new port when a port is occupied

### LINKED ISSUE #557: [bug]: Port Already in Use Error does not Produced while two same apps running on a same port.
### Bug Description

While running two robyn apps on the same port, they do not throw any error, that the port is already in use, instead of this, the app which is running will get the running port.

https://github.com/sparckles/robyn/assets/81439109/6976d7ce-b316-4a92-95fc-926f0d25e0ce

This bug can be fixed alongside the implementation of #508, both the patches of these issues can be combined in the same PR.Robyn

### Steps to Reproduce

https://github.com/sparckles/robyn/assets/81439109/2436f167-397f-449e-9de1-e66e95d85121

### Your operating system

MacOS

3.11

### Your Robyn version

main branch

### Additional Info

_No response_

sansyrox:
**Description**

This PR fixes #557


sansyrox:
@VishnuSanal , would you want to have a look at this PR ? :D

PR #857, opened 2024-06-17
fix: handle the reserved port issue gracefully

### LINKED ISSUE #856: Crashes when a reserved port is used
### Bug Description

```(.venv) RobynApp $ robyn app.py --dev
Starting dev server...
Reloading /home/vishnu/Miscellaneous/fun-projects/RobynApp/app.py...
INFO:robyn.logger:SERVER IS RUNNING IN VERBOSE/DEBUG MODE. Set --log-level to WARN to run in production mode.
INFO:robyn.logger:Added route HttpMethod.GET /
INFO:robyn.logger:Robyn version: 0.55.0
INFO:robyn.logger:Starting server at http://0.0.0.0:22
  File "/home/vishnu/Miscellaneous/fun-projects/RobynApp/app.py", line 12, in <module>
    app.start(host="0.0.0.0", port=22)
  File "/home/vishnu/Miscellaneous/fun-projects/RobynApp/.venv/lib/python3.12/site-packages/robyn/__init__.py", line 228, in start
    run_processes(
  File "/home/vishnu/Miscellaneous/fun-projects/RobynApp/.venv/lib/python3.12/site-packages/robyn/processpool.py", line 31, in run_processes
    socket = SocketHeld(url, port)
             ^^^^^^^^^^^^^^^^^^^^^
PermissionError: Permission denied (os error 13)```

### Steps to Reproduce

```
from robyn import Robyn

app = Robyn(__file__)

@app.get("/")
def index():
    return "hi"

if __name__ == "__main__":
    app.start(host="0.0.0.0", port=22)
```

### Your operating system

Linux

3.12

### Your Robyn version

main branch

### Additional Info

PS: I am working on this.

### PR DESCRIPTION
**Pull request by VishnuSanal**
**Description**

This PR fixes #856

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -191,16 +191,18 @@ class Robyn:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def get_socket(self, url: str, port: int) -> (bool, SocketHeld):
+    def get_socket(self, host: str, port: int) -> (bool, SocketHeld):
         """
-        @param url: the host URL
+        @param host: the host URL
         @param port: the port number
-        @return: a tuple containing a bool that denotes whether
-        the socket was acquired or not followed by the actual socket, if successful
+        @return: this function returns a tuple of whether the socket generation was successful and the actual
+        socket, if successful. the result can have the following 2 possibilities: <br>
+            - true, SocketHeld (if successful) <br>
+            - false, None (if unsuccesssful)
         """
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-                return s.connect_ex(("localhost", port)) != 0, SocketHeld(url, port)
+                return s.connect_ex((host, port)) != 0, SocketHeld(host, port)
         except Exception:
             logger.error(f"Invalid port number: {port}")
             return False, None
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -202,7 +202,9 @@ class Robyn:
         """
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-                return s.connect_ex((host, port)) != 0, SocketHeld(host, port)
+                is_valid = s.connect_ex((host, port))
+                acquired_socket = SocketHeld(host, port)
+                return is_valid, acquired_socket
         except Exception:
             logger.error(f"Invalid port number: {port}")
             return False, None
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -191,7 +191,7 @@ class Robyn:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def get_socket(self, host: str, port: int) -> (bool, SocketHeld):
+    def check_socket_connection(self, host: str, port: int) -> (bool, SocketHeld):
         """
         @param host: the host URL
         @param port: the port number
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -196,7 +196,7 @@ class Robyn:
         @param host: the host URL
         @param port: the port number
         @return: this function returns a tuple of whether the socket generation was successful and the actual
-        socket, if successful. the result can have the following 2 possibilities: <br>
+        socket, if successful. the result can have the following 2 possibilities: 
             - true, SocketHeld (if successful) <br>
             - false, None (if unsuccesssful)
         """
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -202,7 +202,7 @@ class Robyn:
         """
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-                is_valid = s.connect_ex((host, port))
+                is_connection_valid = s.connect_ex((host, port))
                 acquired_socket = SocketHeld(host, port)
                 return is_valid, acquired_socket
         except Exception:
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -196,7 +196,7 @@ class Robyn:
         @param host: the host URL
         @param port: the port number
         @return: this function returns a tuple of whether the socket generation was successful and the actual
-        socket, if successful. the result can have the following 2 possibilities: 
+        socket, if successful. the result can have the following 2 possibilities:
             - true, SocketHeld (if successful) <br>
             - false, None (if unsuccesssful)
         """
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -191,7 +191,7 @@ class Robyn:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def check_socket_connection(self, host: str, port: int) -> (bool, SocketHeld):
+    def check_socket_connection(self, host: str, port: int) -> (bool, Optional[SocketHeld]):
         """
         @param host: the host URL
         @param port: the port number
@@ -204,7 +204,7 @@ class Robyn:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                 is_connection_valid = s.connect_ex((host, port))
                 acquired_socket = SocketHeld(host, port)
-                return is_valid, acquired_socket
+                return is_connection_valid, acquired_socket
         except Exception:
             logger.error(f"Invalid port number: {port}")
             return False, None
@@ -222,14 +222,14 @@ class Robyn:
         port = int(os.getenv("ROBYN_PORT", port))
         open_browser = bool(os.getenv("ROBYN_BROWSER_OPEN", self.config.open_browser))
 
-        valid_socket, _socket = self.get_socket(host, port)
+        is_connection_valid, socket = self.check_socket_connection(host, port)
 
         if _check_port:
-            while not valid_socket:
+            while not is_connection_valid:
                 try:
                     port = int(input("Enter a different port: "))
 
-                    valid_socket, _socket = self.get_socket(host, port)
+                    is_connection_valid, socket = self.check_socket_connection(host, port)
                 except Exception:
                     logger.error("Invalid port number. Please enter a valid port number.")
                     continue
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: robyn/__init__.py [hunk 1]
```diff
@@ -211,14 +212,16 @@ def start(self, host: str = "127.0.0.1", port: int = 8080, _check_port: bool = T
         port = int(os.getenv("ROBYN_PORT", port))
         open_browser = bool(os.getenv("ROBYN_BROWSER_OPEN", self.config.open_browser))
 
-        if _check_port:
```
sansyrox, line 214: this is important, for the tests to pass. Otherwise all the integration tests will fail due to timeouts.

sansyrox:
I have closed the issue. As it is how Unix/(maybe even windows) ports work. Do let me know if you have any more questions 😄

robyn/__init__.py
```diff
@@ -191,12 +191,13 @@ def startup_handler(self, handler: Callable) -> None:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def is_port_in_use(self, port: int) -> bool:
+    def get_socket(self, url: str, port: int) -> (bool, SocketHeld):
```
sansyrox, line 194: Hey @VishnuSanal 👋 

This solution looks interesting. Could you explain what it is doing?

And also add a docstring to the function signature?

sansyrox, line 194: Also could you verify how the test performs with multiple --workers and --processes?

VishnuSanal: how exactly? I tried running `robyn app.py --processes 2 --workers 4` and that seemed to work fine.

sansyrox: Hey @VishnuSanal 

Could you please explain what the function is doing? Reply as a comment here.

VishnuSanal: @sansyrox hi. so, this function returns a tuple of whether the socket generation was successful and the actual socket, if successful. in the try block, we are checking whether the socket is already occupied (as before -- which gives us a true/false). we are also trying to create the socket using `SocketHeld`. which, upon failure goes to the except: block where we are returning false for the status & socket none.

our result can have 2 possibilities:

- true, <valid socket>
- false, none

robyn/__init__.py
```diff
@@ -192,6 +192,12 @@ def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
     def get_socket(self, url: str, port: int) -> (bool, SocketHeld):
+        """
+        @param url: the host URL
+        @param port: the port number
+        @return: a tuple containing a bool that denotes whether
+        the socket was acquired or not followed by the actual socket, if successful
```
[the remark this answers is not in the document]
VishnuSanal: explained the logic in a comment above, please check. :)

robyn/__init__.py
```diff
@@ -191,12 +191,19 @@ def startup_handler(self, handler: Callable) -> None:
...
+        """
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-                return s.connect_ex(("localhost", port)) == 0
+                return s.connect_ex(("localhost", port)) != 0, SocketHeld(url, port)
```
sansyrox, line 203: @VishnuSanal , one final suggestion. Split this into two lines

i.e.

```
acquired_socket = SocketHeld(url, port)
is_valid = s.connect_ex()
```

Also, now that I think of it. we are only checking "localhost" here. But we also deploy to "0.0.0.0" . could you use a variable `host` here instead?

sansyrox: Hey @VishnuSanal , just some final suggestions 👋

robyn/__init__.py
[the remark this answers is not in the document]
sansyrox: Thanks!

robyn/__init__.py
sansyrox: Alright, that sounds good to me.

robyn/__init__.py
```diff
@@ -191,12 +191,23 @@ def startup_handler(self, handler: Callable) -> None:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def is_port_in_use(self, port: int) -> bool:
+    def get_socket(self, host: str, port: int) -> (bool, SocketHeld):
```
sansyrox, line 194:
```suggestion
    def get_socket(self, host: str, port: int) -> (bool, Optional[SocketHeld]):
```

robyn/__init__.py
```diff
@@ -211,11 +222,14 @@ def start(self, host: str = "127.0.0.1", port: int = 8080, _check_port: bool = T
...
+        valid_socket, _socket = self.get_socket(host, port)
+
         if _check_port:
-            while self.is_port_in_use(port):
-                logger.error("Port %s is already in use. Please use a different port.", port)
+            while not valid_socket:
                 try:
                     port = int(input("Enter a different port: "))
+
+                    valid_socket, _socket = self.get_socket(host, port)
```
sansyrox, line 232: @VishnuSanal , let's not use `_socket` since it is being referenced. We pre prend `_` to show something not being used. Let's call it `socket`

robyn/__init__.py
sansyrox, line 194:
```suggestion
    def check_socket_connection(self, host: str, port: int) -> (bool, SocketHeld):
```

robyn/__init__.py
```diff
@@ -191,12 +191,23 @@ def startup_handler(self, handler: Callable) -> None:
...
+        @param port: the port number
+        @return: this function returns a tuple of whether the socket generation was successful and the actual
+        socket, if successful. the result can have the following 2 possibilities: <br>
+            - true, SocketHeld (if successful) <br>
+            - false, None (if unsuccesssful)
+        """
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-                return s.connect_ex(("localhost", port)) == 0
+                is_valid = s.connect_ex((host, port))
```
sansyrox, line 205:
```suggestion
                is_connection_valid = s.connect_ex((host, port))
```

robyn/__init__.py
VishnuSanal: IIRC, I got a name shadowing warning.

robyn/__init__.py
```diff
@@ -211,11 +222,14 @@ def start(self, host: str = "127.0.0.1", port: int = 8080, _check_port: bool = T
         port = int(os.getenv("ROBYN_PORT", port))
         open_browser = bool(os.getenv("ROBYN_BROWSER_OPEN", self.config.open_browser))
 
+        valid_socket, _socket = self.get_socket(host, port)
```
sansyrox, line 225: `is_connection_valid, socket = self.check_socket_connection`

sansyrox: So, a rule of thumb for naming conventions that I follow - the line should read like a sentence.

sansyrox:
Hey @VishnuSanal 👋 

Good job. I have some comments

robyn/__init__.py
sansyrox: @VishnuSanal , but you are using the variable in the lines below

VishnuSanal: my bad then, done! :)

VishnuSanal:
heads up: #868 is not fixed with this PR.

robyn/processpool.py
```diff
@@ -16,6 +16,7 @@
 def run_processes(
     url: str,
     port: int,
+    socket: SocketHeld,
```
sansyrox, line 19: not needed

robyn/__init__.py
```diff
@@ -228,6 +242,7 @@ def start(self, host: str = "127.0.0.1", port: int = 8080, _check_port: bool = T
         run_processes(
             host,
             port,
+            socket,
```
sansyrox, line 245: not needed

robyn/__init__.py
```diff
@@ -191,12 +191,23 @@ def startup_handler(self, handler: Callable) -> None:
     def shutdown_handler(self, handler: Callable) -> None:
         self._add_event_handler(Events.SHUTDOWN, handler)
 
-    def is_port_in_use(self, port: int) -> bool:
+    def check_socket_connection(self, host: str, port: int) -> (bool, Optional[SocketHeld]):
+        """
+        @param host: the host URL
+        @param port: the port number
+        @return: this function returns a tuple of whether the socket generation was successful and the actual
+        socket, if successful. the result can have the following 2 possibilities: <br>
```
sansyrox, line 199:
```suggestion
        socket, if successful. the result can have the following 2 possibilities: 
```

sansyrox:
@VishnuSanal , this looks good! I will try to run it locally tomorrow. And will merge after that :D

sansyrox:
@VishnuSanal , still doesn't work :/

image

VishnuSanal:
the issue is that the default value in the function header (or the one from the user) takes over. I even tried adding `os.environ["ROBYN_PORT"] = str(port)`. the current env actually gets copied on `EventHandler#reload`. but, that doesn't work either.

the flow of control is like: `cli#run` => `cli#start_dev_server` => `reloader#setup_reloader` => `EventHandler#reload` => `Robyn#start` => `processpool#run_processes`

and we are updating the port on `Robyn#start` here. since this comes at a later point than the reloading takes place is why the reloader loads the default value.

the solution I can think of is to save the port as a config variable. what say? @sansyrox

sansyrox:
@VishnuSanal ,

one question - fix: handle the reserved port issue gracefully < are you still trying to resolve this issue? If yes, then we shouldn't worry about it. This an be resolved by using sudo.

VishnuSanal:
> one question - fix: handle the reserved port issue gracefully < are you still trying to resolve this issue? If yes, then we shouldn't worry about it.

alright, sorry for the trouble. I created the issue only because the following was not a good user experience

```
  File "RobynApp/app.py", line 12, in <module>
    app.start(host="0.0.0.0", port=1000)
  File "RobynApp/.venv/lib/python3.12/site-packages/robyn/__init__.py", line 228, in start
    run_processes(
  File "RobynApp/.venv/lib/python3.12/site-packages/robyn/processpool.py", line 31, in run_processes
    socket = SocketHeld(url, port)
             ^^^^^^^^^^^^^^^^^^^^^
PermissionError: Permission denied (os error 13)
```

> This an be resolved by using sudo.

then, probably printing this & exiting will be a better.

PR #858, opened 2024-06-17
feat: live reloading when running as a module

### LINKED ISSUE #654: Live reloading doesn't work when running as a module
### Bug Description

try running `python3 -m robyn_app`

### Steps to Reproduce

_No response_

### Your operating system

None

None

### Your Robyn version

None

### Additional Info

_No response_

VishnuSanal:
**Description**

This PR fixes #654

how to run: `python3 -m app --dev`

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -7,10 +7,11 @@ from typing import Callable, List, Optional, Tuple, Union
 import multiprocess as mp
 from nestd import get_all_nested
 
-from robyn import status_codes, dev_server
+from robyn import status_codes
 from robyn.argument_parser import Config
 from robyn.authentication import AuthenticationHandler
 from robyn.dependency_injection import DependencyMap
+from robyn.dev_server import start_dev_server
 from robyn.env_populator import load_vars
 from robyn.events import Events
 from robyn.jsonify import jsonify
@@ -224,7 +225,7 @@ class Robyn:
         mp.allow_connection_pickling()
 
         if self.config.dev:
-            dev_server.start_dev_server(self.config, self.config.file_path)
+            start_dev_server(self.config, self.config.file_path)
         else:
             run_processes(
                 host,
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -10,6 +10,7 @@ from nestd import get_all_nested
 from robyn import status_codes
 from robyn.argument_parser import Config
 from robyn.authentication import AuthenticationHandler
+from robyn.cli import start_dev_server
 from robyn.dependency_injection import DependencyMap
 from robyn.env_populator import load_vars
 from robyn.events import Events
@@ -56,8 +57,6 @@ class Robyn:
                 "SERVER IS RUNNING IN VERBOSE/DEBUG MODE. Set --log-level to WARN to run in production mode.",
                 color=Colors.BLUE,
             )
-        if self.config.dev:
-            exit("Dev mode is not supported in the python wrapper. Please use the CLI. e.g. python3 -m robyn app.py --dev ")
 
         self.router = Router()
         self.middleware_router = MiddlewareRouter()
@@ -225,21 +224,24 @@ class Robyn:
 
         mp.allow_connection_pickling()
 
-        run_processes(
-            host,
-            port,
-            self.directories,
-            self.request_headers,
-            self.router.get_routes(),
-            self.middleware_router.get_global_middlewares(),
-            self.middleware_router.get_route_middlewares(),
-            self.web_socket_router.get_routes(),
-            self.event_handlers,
-            self.config.workers,
-            self.config.processes,
-            self.response_headers,
-            open_browser,
-        )
+        if self.config.dev:
+            start_dev_server(self.config, self.config.file_path)
+        else:
+            run_processes(
+                host,
+                port,
+                self.directories,
+                self.request_headers,
+                self.router.get_routes(),
+                self.middleware_router.get_global_middlewares(),
+                self.middleware_router.get_route_middlewares(),
+                self.web_socket_router.get_routes(),
+                self.event_handlers,
+                self.config.workers,
+                self.config.processes,
+                self.response_headers,
+                open_browser,
+            )
 
     def exception(self, exception_handler: Callable):
         self.exception_handler = exception_handler
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -7,10 +7,9 @@ from typing import Callable, List, Optional, Tuple, Union
 import multiprocess as mp
 from nestd import get_all_nested
 
-from robyn import status_codes
+from robyn import status_codes, dev_server
 from robyn.argument_parser import Config
 from robyn.authentication import AuthenticationHandler
-from robyn.cli import start_dev_server
 from robyn.dependency_injection import DependencyMap
 from robyn.env_populator import load_vars
 from robyn.events import Events
@@ -225,7 +224,7 @@ class Robyn:
         mp.allow_connection_pickling()
 
         if self.config.dev:
-            start_dev_server(self.config, self.config.file_path)
+            dev_server.start_dev_server(self.config, self.config.file_path)
         else:
             run_processes(
                 host,
--- a/robyn/argument_parser.py
+++ b/robyn/argument_parser.py
@@ -82,6 +82,7 @@ class Config:
         self.version = args.version
         self.compile_rust_path = args.compile_rust_path
         self.create_rust_file = args.create_rust_file
+        self.running_as_module = False
 
         # find something that ends with .py in unknown_args
         for arg in unknown_args:
@@ -90,6 +91,7 @@ class Config:
                 break
         else:
             self.file_path = sys.argv[0]
+            self.running_as_module = True
 
         if self.dev and (self.processes != 1 or self.workers != 1):
             raise Exception("--processes and --workers shouldn't be used with --dev")
--- a/robyn/argument_parser.py
+++ b/robyn/argument_parser.py
@@ -1,4 +1,5 @@
 import argparse
+import sys
 
 
 class Config:
@@ -87,6 +88,8 @@ class Config:
             if arg.endswith(".py"):
                 self.file_path = arg
                 break
+        else:
+            self.file_path = sys.argv[0]
 
         if self.dev and (self.processes != 1 or self.workers != 1):
             raise Exception("--processes and --workers shouldn't be used with --dev")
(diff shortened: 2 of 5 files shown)


sansyrox:
Hey @VishnuSanal 👋 

I am unable to understand the changes. Could you add an example on why we need the new parameters?

robyn/argument_parser.py
```diff
@@ -26,6 +26,12 @@ def __init__(self) -> None:
             default=False,
             help="Development mode. It restarts the server based on file changes.",
         )
+        parser.add_argument(
+            "--file",
+            dest="file_path",
+            default=None,
+            help="The file associated.",
+        )
```
sansyrox, line 34: What does this do?

VishnuSanal: @sansyrox hi, I have changed the implementation. PTAL :)

sansyrox:
> Hey @VishnuSanal 👋 
> 
> 
> 
> I am unable to understand the changes. Could you add an example on why we need the new parameters?
> 
> 
> 
> And how you'd invoke the changes?

Hey @VishnuSanal ,
Could you please update the PR description with the above??

VishnuSanal:
uh, oh. my previous comment never got sent.

> Could you add an example on why we need the new parameters?

I have changed the implementation, PTAL. removed the need for new params.

> And how you'd invoke the changes?

running as a module usually, like `python3 -m app --dev` would do.

please CMIIW.

sansyrox:
@VishnuSanal , I just saw your description. That is not what we want to solve for 😄 

We want to allow the dev mode to work like `robyn -m src --dev`

where src

```
src/
  - __init__.py
  - __main__.py
```

The --dev flag should rerender based on app changes.

VishnuSanal:
thank you for the review, Sanskar! this currently works as expected in my testing, PTAL & LMK what you think. :)

robyn/argument_parser.py
```diff
@@ -87,6 +88,8 @@ def __init__(self) -> None:
             if arg.endswith(".py"):
                 self.file_path = arg
                 break
+        else:
+            self.file_path = sys.argv[0]
```
sansyrox, line 92: this is clever. but where are you using this?

robyn/reloader.py
```diff
@@ -115,11 +117,19 @@ def reload(self):
...
-            env=new_env,
-            start_new_session=False,
-        )
+        if self.config.dev and self.config.file_path is not None:
+            subprocess.call(f'cd {"/".join(self.config.file_path.split("/")[-2:-1])}', shell=True)
+            self.process = subprocess.Popen(
+                [sys.executable, "-m", self.config.file_path.split("/")[-2], *arguments],
+                env=new_env,
+                start_new_session=False,
+            )
```
sansyrox, line 126: too many things are happening here. Could you split these into different lines?

robyn/reloader.py
```diff
@@ -115,11 +117,19 @@ def reload(self):
...
 
-        self.process = subprocess.Popen(
-            [sys.executable, *arguments],
-            env=new_env,
-            start_new_session=False,
-        )
+        if self.config.dev and self.config.file_path is not None:
+            subprocess.call(f'cd {"/".join(self.config.file_path.split("/")[-2:-1])}', shell=True)
+            self.process = subprocess.Popen(
+                [sys.executable, "-m", self.config.file_path.split("/")[-2], *arguments],
```
sansyrox, line 123: why do we need a `-m` here?

robyn/argument_parser.py
```diff
@@ -87,6 +88,8 @@ def __init__(self) -> None:
             if arg.endswith(".py"):
```
sansyrox, line 88: @VishnuSanal , this is what stops the app from running in module mode

e.g.

`robyn -m src --dev` will not work if all the code is present in the `src/__main__.py` file

robyn/argument_parser.py
VishnuSanal: in `EventHandler#reload` (inside the new `if` block I added)

robyn/reloader.py
VishnuSanal: to run again as a module whilst reloading...? since, we have run the app as a module initially like `python3 -m src --dev`

robyn/argument_parser.py
VishnuSanal: > this is what stops the app from running in module mode

yep, the missing parameter is `file_path`. this is why I have assigneed it in the else block & did worksrounds to make it work on other parts of the code.

robyn/reloader.py
```diff
@@ -115,11 +117,24 @@ def reload(self):
...
 
-        self.process = subprocess.Popen(
-            [sys.executable, *arguments],
-            env=new_env,
-            start_new_session=False,
-        )
+        if self.config.dev and self.config.file_path is not None:
+            split = self.config.file_path.split("/")
+
+            working_directory = "/".join(split[-2:-1])
```
sansyrox, line 123: what are we splitting here? could you please write more descriptive names?

robyn/reloader.py
```diff
@@ -115,11 +117,24 @@ def reload(self):
...
+
+            subprocess.call(f"cd {working_directory}", shell=True)
```
sansyrox, line 125: you can use `os.chdir` here instead.

Also, why do we need this again?

robyn/reloader.py
```diff
@@ -115,11 +117,24 @@ def reload(self):
...
+
+            self.process = subprocess.Popen(
+                [sys.executable, "-m", split[-2], *arguments],
```
sansyrox, line 128: can you assign `split[-2]` to a variable?

robyn/__init__.py
```diff
@@ -10,6 +10,7 @@
 from robyn import status_codes
 from robyn.argument_parser import Config
 from robyn.authentication import AuthenticationHandler
+from robyn.cli import start_dev_server
```
sansyrox, line 13: could you move `start_dev_server` to a new file maybe `robyn/dev_server.py`?

sansyrox:
Hey @VishnuSanal 👋 

Thanks for the PR, I have some comments 😄

robyn/__init__.py
```diff
@@ -7,10 +7,9 @@
 import multiprocess as mp
 from nestd import get_all_nested
 
-from robyn import status_codes
+from robyn import status_codes, dev_server
```
sansyrox, line 10: @VishnuSanal , please import from `robyn.dev_server` directly here

robyn/dev_server.py
```diff
@@ -0,0 +1,18 @@
+import os
+from pathlib import Path
+from typing import Optional
+
+from robyn import reloader
+from robyn.argument_parser import Config
+
+
+def start_dev_server(config: Config, file_path: Optional[str] = None):
```
sansyrox, line 9: Could you please add a docstring here?

robyn/reloader.py
```diff
@@ -118,14 +118,10 @@ def reload(self):
         self.built_rust_binaries = compile_rust_files(self.directory_path)
 
         if self.config.dev and self.config.file_path is not None:
-            split = self.config.file_path.split("/")
-
-            working_directory = "/".join(split[-2:-1])
-
-            subprocess.call(f"cd {working_directory}", shell=True)
+            module_name = self.config.file_path.split("/")[-2]
```
sansyrox, line 121: are you sure this will always be the case?

i.e. the module name will always be the second to last name?

robyn/cli.py
```diff
@@ -1,11 +1,12 @@
 import os
 import sys
-from typing import Optional
 import webbrowser
 from InquirerPy.resolver import prompt
 from InquirerPy.base.control import Choice
+
+from . import dev_server
```
sansyrox, line 7: please use absolute paths instead of relative paths.

e.g. `from robyn.dev_server import start_dev_server`

sansyrox:
Hey @VishnuSanal 👋 

Thanks for the update. I have some follow up comments

robyn/reloader.py
VishnuSanal: @sansyrox oh, okay, that makes sense. do you have any suggestions on how to do this?!

here are some solutions from the top of my head:

- can we reload it by running it normally (not a module). but, this is undesirable.

VishnuSanal: hey, my thoughts were, `file_path` contains `sys.argv[0]`. which will be the path to the `__main__.py` file. hence, the second to last of it will be it's directory name. hence the module name. won't this always be the case?

sansyrox: @VishnuSanal , I believe that we can make use of `__file__` that is being passed in the Robyn() class

robyn/cli.py
```diff
@@ -1,11 +1,12 @@
...
 import sys
-from typing import Optional
 import webbrowser
 from InquirerPy.resolver import prompt
 from InquirerPy.base.control import Choice
+
+from robyn.dev_server import start_dev_server
 from .argument_parser import Config
-from .reloader import create_rust_file, setup_reloader
+from .reloader import create_rust_file
```
sansyrox, line 9: @VishnuSanal , could we please use absolute paths here as well?

robyn.argument_parser
robyn.reloader

robyn/reloader.py
```diff
@@ -115,11 +117,20 @@ def reload(self):
         clean_rust_binaries(self.built_rust_binaries)
         self.built_rust_binaries = compile_rust_files(self.directory_path)
 
-        self.process = subprocess.Popen(
-            [sys.executable, *arguments],
-            env=new_env,
-            start_new_session=False,
-        )
+        if self.config.dev and self.config.file_path is not None:
```
sansyrox, line 120: Can we not use `self.file_path`? 

Then we don't need to pass in `config`.

VishnuSanal: we also need `config.dev`. should I pass it as a variable?

VishnuSanal:
current issues:

- doesn't work when using `robyn -m src --dev`. `sys.argv[0]` returns the path to the executable. but, from [here](https://docs.python.org/3/using/cmdline.html#cmdoption-m), it seems like we shouldn't run it like it, but like `python3 -m src --dev`.

(will edit & update this comment...)

VishnuSanal:
note to self: 

the flow of control is like: cli#run => cli#start_dev_server => reloader#setup_reloader => EventHandler#reload => Robyn#start and, we are getting the __file__ object at the last stage. my current implementation starts a new dev server from there.

we need a way to bypass it the first time (inorder to reach `robyn#start`) and, then kill the previous reloader & start new one from there.

PR #870, opened 2024-06-24
feat: implement `set_cookie` function

### LINKED ISSUE #854: add a set cookie interface in the response object
Something like this

```python
@app.get("/cookie/")
def create_cookie():
    response = Response(....)
    response.set_cookie(key="fakesession", value="fake-cookie-session-value")
    return response
```

VishnuSanal:
**Description**

This PR fixes #854

### DIFF
--- a/integration_tests/test_get_requests.py
+++ b/integration_tests/test_get_requests.py
@@ -3,6 +3,8 @@ import requests
 from requests import Response
 
 from integration_tests.helpers.http_methods_helpers import get
+from robyn import Headers
+from robyn import Response as RobynResponse
 
 
 @pytest.mark.benchmark
@@ -61,3 +63,12 @@ def test_trailing_slash(session):
 
     r = requests.get("http://localhost:8080/trailing/")
     assert r.text == "Trailing slash test successful!"
+
+
+@pytest.mark.benchmark
+@pytest.mark.parametrize("key, value", [("fakesession", "fake-cookie-session-value")])
+def test_cookies(session, key, value):
+    response = RobynResponse(status_code=200, headers=Headers({}), description="test cookies")
+    response.set_cookie(key=key, value=value)
+
+    assert response.headers[key] == value
--- a/integration_tests/test_get_requests.py
+++ b/integration_tests/test_get_requests.py
@@ -3,8 +3,6 @@ import requests
 from requests import Response
 
 from integration_tests.helpers.http_methods_helpers import get
-from robyn import Headers
-from robyn import Response as RobynResponse
 
 
 @pytest.mark.benchmark
@@ -68,7 +66,6 @@ def test_trailing_slash(session):
 @pytest.mark.benchmark
 @pytest.mark.parametrize("key, value", [("fakesession", "fake-cookie-session-value")])
 def test_cookies(session, key, value):
-    response = RobynResponse(status_code=200, headers=Headers({}), description="test cookies")
-    response.set_cookie(key=key, value=value)
+    response = get("/cookie", 200)
 
     assert response.headers[key] == value
--- a/robyn/robyn.pyi
+++ b/robyn/robyn.pyi
@@ -285,8 +285,6 @@ class Response:
         headers (Union[Headers, dict]): The headers of the response or Headers directly. e.g. {"Content-Type": "application/json"}
         description (Union[str, bytes]): The body of the response. If the response is a JSON, it will be a dict.
         file_path (Optional[str]): The file path of the response. e.g. /home/user/file.txt
-
-        set_cookie (Callable): The function to set a cookie (onto headers).
     """
 
     status_code: int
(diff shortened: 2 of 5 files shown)


sansyrox:
@VishnuSanal , could you add some integration tests?

integration_tests/test_get_requests.py
```diff
@@ -61,3 +63,12 @@ def test_trailing_slash(session):
 
     r = requests.get("http://localhost:8080/trailing/")
     assert r.text == "Trailing slash test successful!"
+
+
+@pytest.mark.benchmark
+@pytest.mark.parametrize("key, value", [("fakesession", "fake-cookie-session-value")])
```
sansyrox, line 69: @VishnuSanal , could you please do an e2e test, where you create a new endpoint and check a response?

src/types/response.rs
```diff
@@ -156,4 +156,14 @@ impl PyResponse {
             .into_py(py);
         Ok(())
     }
+
+    pub fn set_cookie(&mut self, py: Python, key: &str, value: &str) -> PyResult<()> {
```
sansyrox, line 160:
```
impl PyResponse {
    pub fn set_cookie(&mut self, name: &str, value: &str, path: Option<&str>, expires: Option<&str>, http_only: bool, secure: bool) {
        let mut cookie_value = format!("{}={}", name, value);
        if let Some(path) = path {
            cookie_value.push_str(&format!("; Path={}", path));
        }
        if let Some(expires) = expires {
            cookie_value.push_str(&format!("; Expires={}", expires));
        }
        if http_only {
            cookie_value.push_str("; HttpOnly");
        }
        if secure {
            cookie_value.push_str("; Secure");
        }

        self.headers.insert("Set-Cookie".to_string(), cookie_value);
    }
}

```

sansyrox: Why not something like the above?

VishnuSanal: @sansyrox sure, if you want it to be done so, I will update. we should probably give a `#get_cookie` function too, IMO.

(I implemented it the way I did cuz of the example code provided in the issue).

src/types/response.rs
```diff
@@ -156,4 +156,14 @@ impl PyResponse {
...
+        let kwargs = vec![("key", key), ("value", value)].into_py_dict(py);
+
+        if let Err(e) = self.headers.call_method(py, "append", (), Some(kwargs)) {
```
sansyrox, line 163: @VishnuSanal , calling methods in py handlers is usually slower. So it is preferred to alter rust handlers directly and then postpone the conversion until the end, if possible

src/types/response.rs
sansyrox: Hey @VishnuSanal , added a suggestion in near the exact line of concern.

src/types/response.rs
VishnuSanal: @sansyrox hi, I can't find another way to do this. I have removed the use of `pydict` now. please check. 

when using `headers.append` (as you mentioned in your code) leads me to the following error.

```
error[E0599]: no method named `append` found for struct `pyo3::Py` in the current scope
   --> src/types/response.rs:161:22
    |
161 |         self.headers.append(key.to_string(), value.to_string());
    |                      ^^^^^^ method not found in `Py<Headers>`

```

src/types/response.rs
```diff
@@ -156,4 +156,12 @@ impl PyResponse {
             .into_py(py);
         Ok(())
     }
+
+    pub fn set_cookie(&mut self, py: Python, key: &str, value: &str) -> PyResult<()> {
+        self.headers
+            .try_borrow_mut(py)
+            .expect("value already borrowed")
+            .append(key.to_string(), value.to_string());
+        Ok(())
```
sansyrox, line 165: Nicee

sansyrox:
@VishnuSanal , great work! Could you please update the `Response` class in `robyn.pyi` file with the updated docstring and method?

Also, could you add relevant docs to the documentation?

robyn/robyn.pyi
```diff
@@ -293,6 +295,9 @@ class Response:
     response_type: Optional[str] = None
     file_path: Optional[str] = None
 
+    def set_cookie(self, key, value) -> None:
```
sansyrox, line 298: @VishnuSanal , please add a docstring in the public functions. And also add typing here

robyn/robyn.pyi
```diff
@@ -285,6 +285,8 @@ class Response:
         headers (Union[Headers, dict]): The headers of the response or Headers directly. e.g. {"Content-Type": "application/json"}
         description (Union[str, bytes]): The body of the response. If the response is a JSON, it will be a dict.
         file_path (Optional[str]): The file path of the response. e.g. /home/user/file.txt
+
+        set_cookie (Callable): The function to set a cookie (onto headers).
```
sansyrox, line 289: This is not needed. We only document the attributes

sansyrox, line 289:
```suggestion
```

sansyrox (approved):
LGTM 👍

PR #882, opened 2024-07-04
fix: bug Binary type Reponse causes a crash#874

### LINKED ISSUE #874: Binary type Reponse causes a crash
### Bug Description

I use Robyn to write the Export Excel function, which works fine with very early versions and always crashes when updated to recent versions. I read the rust code and found that Response does a Utf8 conversion of all returned data was the cause of the crashes, then I tried to tweak the code (not knowing rust at all, AI and PyO3 docs) to differentiate between text, binary `description` Luckily the crashes disappeared!

I guessed the type of Response data by Content-Type, but there are so many MIME-Types that I thought maybe only guessing on the most common types to determine and allow users to register if the MIME-Type they use is a Text type ......

These are some of my attempts to fix this problem, I hope to get an official fix, and I wish Robyn is getting better and better, I really like it too much.

### Steps to Reproduce

1. response `Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` data

### Your operating system

Linux

3.11

### Your Robyn version

latest

### Additional Info

_No response_

asamaayako:
**Description**

This PR fixes #874  if not utf-8 data. just using Bytes. for respnose body

### DIFF
--- a/src/types/response.rs
+++ b/src/types/response.rs
@@ -68,6 +68,8 @@ impl Response {
 impl ToPyObject for Response {
     fn to_object(&self, py: Python) -> PyObject {
         let headers = self.headers.clone().into_py(py).extract(py).unwrap();
+        // The description should only be either string or binary.
+        // it should raise an exception otherwise
         let description = match String::from_utf8(self.description.to_vec()) {
             Ok(description) => description.to_object(py),
             Err(_) => PyBytes::new(py, &self.description.to_vec()).into(),
--- a/src/types/response.rs
+++ b/src/types/response.rs
@@ -68,9 +68,11 @@ impl Response {
 impl ToPyObject for Response {
     fn to_object(&self, py: Python) -> PyObject {
         let headers = self.headers.clone().into_py(py).extract(py).unwrap();
-        let description = String::from_utf8(self.description.to_vec())
-            .unwrap()
-            .to_object(py);
+        let description = match String::from_utf8(self.description.to_vec()) {
+            Ok(description) => description.to_object(py),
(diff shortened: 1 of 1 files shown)


src/types/response.rs
```diff
@@ -68,9 +68,11 @@ impl Response {
 impl ToPyObject for Response {
     fn to_object(&self, py: Python) -> PyObject {
         let headers = self.headers.clone().into_py(py).extract(py).unwrap();
-        let description = String::from_utf8(self.description.to_vec())
-            .unwrap()
-            .to_object(py);
+        let description = match String::from_utf8(self.description.to_vec()) {
+            Ok(description) => description.to_object(py),
+            Err(_) => PyBytes::new(py, &self.description.to_vec()).into(),
```
sansyrox, line 73: @asamaayako , thanks for the PR. 😄 

Are these the only two possible cases?

sansyrox (approved):
LGTM . Great work👍

src/types/response.rs
```diff
@@ -68,9 +68,11 @@ impl Response {
 impl ToPyObject for Response {
     fn to_object(&self, py: Python) -> PyObject {
         let headers = self.headers.clone().into_py(py).extract(py).unwrap();
-        let description = String::from_utf8(self.description.to_vec())
-            .unwrap()
-            .to_object(py);
+        let description = match String::from_utf8(self.description.to_vec()) {
```
sansyrox, line 71:
```suggestion
        // The description should only be either string or binary.
        // it should raise an exception otherwise
        let description = match String::from_utf8(self.description.to_vec()) {
```

goSeeFuture:
v0.57.1 No more crashes, thank you very much!

PR #883, opened 2024-07-05
fix: typo error in postgres scaffold

octaviusp:
**Description**

The app.py of postgres scaffold was started with "url" prefix, as it doesn't exists anymore, it throws an error. Now, is changed to "host", as the other scaffolds.


sansyrox (approved):
LGTM! :+1:

PR #885, opened 2024-07-08
feat: add ability to log to a file

### LINKED ISSUE #841: Add an ability to log to a file

carlosm27:
**Description**

This PR fixes #841

This PR adds the ability to create a log file.
This is an example of how to do it:

Screenshot 2024-07-07 230957

Screenshot 2024-07-07 230937


carlosm27:
Should we add the ability to add more configurations, like timestamps, levels, etc?

carlosm27:
> Hey @carlosm27 👋
> 
> How would you use this new interface?

Hey @sansyrox. I would use it to create log files with timestamps, log level and with and add rotation policy.

What features are more appropiate for this interface?

xiaojin20:
API is very practical. Does it support asynchronous operation?

Oluwaseun241 (approved):

sansyrox (commented):
Thanks for tackling this, @carlosm27 — file logging (#841) is genuinely useful and the `FileHandler` approach is the right idea. A couple of things before we can merge: the PR accidentally commits a compiled Windows binary (`robyn/robyn.cp311-win_amd64.pyd`) which needs to be removed, and the changes to `integration_tests/index.py` are demo snippets rather than an actual test. One functional concern: our log messages are wrapped with ANSI color codes via `_format_msg`, so they'd get written into the file as raw escape sequences — could you strip colors for the file handler? Finally, a quick rename to something like `log_to_file` would read more naturally. Happy to help get this over the line once those are addressed!

PR #850, opened 2024-06-12
feat: split request params

sansyrox:
**Description**

This PR fixes #

PR #894, opened 2024-07-14
Feat: split request params

kliwongan:
**Description**

This PR fixes #850

Fix is relatively simple, involves extracting Request object from Rust passing it in, and if no Request object is passed in simply call the handler function as usual.

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -830,6 +830,85 @@ def sync_router_di(request, router_dependencies):
     return router_dependencies["ROUTER_DEPENDENCY"]
 
 
+# ===== Split request body =====
+
+
+@app.get("/sync/split_request/query_params")
+def sync_split_request_basic(request: Request, query_params):
+    return jsonify(query_params.to_dict())
+
+
+@app.get("/async/split_request/query_params")
+async def async_split_request_basic(request: Request, query_params):
+    return jsonify(query_params.to_dict())
+
+
+@app.get("/sync/split_request/headers")
+def sync_split_request_headers(request: Request, headers):
+    return headers.get("server")
+
+
+@app.get("/async/split_request/headers")
+async def async_split_request_headers(request: Request, headers):
+    return headers.get("server")
+
+
+@app.get("/sync/split_request/path_params/:id")
+def sync_split_request_path_params(request: Request, path_params):
+    return path_params
+
+
+@app.get("/async/split_request/path_params/:id")
+async def async_split_request_path_params(request: Request, path_params):
+    return path_params
+
+
+@app.get("/sync/split_request/method")
+def sync_split_request_method(request: Request, method):
+    return method
+
+
+@app.get("/async/split_request/method")
+async def async_split_request_method(request: Request, method):
+    return method
+
+
+@app.post("/sync/split_request/body")
+def sync_split_request_body(request: Request, body):
+    return body
+
+
+@app.post("/async/split_request/body")
+async def async_split_request_body(request: Request, body):
+    return body
+
+
+@app.post("/sync/split_request/combined")
+def sync_split_request_combined(request: Request, body, query_params, method, url, headers):
+    return jsonify(
+        {
+            "body": body,
+            "query_params": query_params.to_dict(),
+            "method": method,
+            "url": url.path,
+            "headers": headers.get("server"),
+        }
+    )
+
+
+@app.post("/async/split_request/combined")
+async def async_split_request_combined(request: Request, body, query_params, method, url, headers):
+    return jsonify(
+        {
+            "body": body,
+            "query_params": query_params.to_dict(),
+            "method": method,
+            "url": url.path,
+            "headers": headers.get("server"),
+        }
+    )
+
+
 def main():
     app.set_response_header("server", "robyn")
     app.add_directory(
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -124,11 +124,38 @@ class Router(BaseRouter):
         exception_handler: Optional[Callable],
         injected_dependencies: dict,
     ) -> Union[Callable, CoroutineType]:
+        def wrapped_handler(*args, **kwargs):
+            # In the execute functions the request is passed into *args
+            handler_params = signature(handler).parameters
+            # If no request arg, assume no other args are present
+            requests = list(filter(lambda a: isinstance(a, Request), args))
+            if not len(requests):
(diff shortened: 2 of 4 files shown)


integration_tests/base_routes.py
```diff
@@ -830,6 +830,85 @@ def sync_router_di(request, router_dependencies):
     return router_dependencies["ROUTER_DEPENDENCY"]
 
 
+# ===== Split request body =====
+
+
+@app.get("/sync/split_request/query_params")
+def sync_split_request_basic(request: Request, query_params):
```
sansyrox, line 837: @kliwongan would this work?
```suggestion
def sync_split_request_basic(query_params, request: Request):
```

integration_tests/base_routes.py
```diff
@@ -830,6 +830,85 @@ def sync_router_di(request, router_dependencies):
...
+    return jsonify(query_params.to_dict())
```
sansyrox, line 838: `jsonify` is not needed anymore. You can just return a dictionary 😄

robyn/router.py
```diff
@@ -124,11 +124,38 @@ def add_route(
         exception_handler: Optional[Callable],
         injected_dependencies: dict,
     ) -> Union[Callable, CoroutineType]:
+        def wrapped_handler(*args, **kwargs):
+            # In the execute functions the request is passed into *args
+            handler_params = signature(handler).parameters
+            # If no request arg, assume no other args are present
+            requests = list(filter(lambda a: isinstance(a, Request), args))
```
sansyrox, line 131: How does this handle `INJECTED_DEPENDENCIES`?

sansyrox:
Hey @kliwongan 👋 

Great work! I have some follow ups 😄

integration_tests/base_routes.py
kliwongan: Yes it would. I run a search for a request object from the *args parameter, so it doesn't necessitate the request object being passed in first.

robyn/router.py
sansyrox: @kliwongan , one question what happens if there is no `requests, query_params` and just `query_params` ?

integration_tests/base_routes.py
```diff
@@ -830,6 +830,85 @@ def sync_router_di(request, router_dependencies):
     return router_dependencies["ROUTER_DEPENDENCY"]
 
 
+# ===== Split request body =====
```
sansyrox, line 833: Love the coverage! have some suggestions with the structure though 😄

robyn/router.py
kliwongan: It wouldn't work without the request object. If I did make it work without the request object it would require changes on the Rust side of things, so signatures would be checked before or during passing the args from the Rust side as the handler is being called

kliwongan: As for router/global dependencies, I can make the change to handle them properly now; I can simply check for their existence first in the handler signature and then if they exist at all in the injected dependencies

integration_tests/base_routes.py
kliwongan: Please go ahead

sansyrox: we should not use `jsonify` :D

robyn/router.py
sansyrox: >It wouldn't work without the request object. If I did make it work without the request object it would require changes on the Rust side of things, so signatures would be checked before or during passing the args from the Rust side as the handler is being called

That kills the purpose of splitting the request params, unfortunately 😓

sansyrox: The request handler is always(if there is one or more args) being passed from rust. No?

integration_tests/base_routes.py
sansyrox: @kliwongan , amazing. Could you add a test to assert that?

robyn/router.py
kliwongan: > > It wouldn't work without the request object. If I did make it work without the request object it would require changes on the Rust side of things, so signatures would be checked before or during passing the args from the Rust side as the handler is being called
> 
> That kills the purpose of splitting the request params, unfortunately 😓

I can certainly make the necessary changes, it would simply move some, if not all of the changes over to the Rust side.

sansyrox: > > > It wouldn't work without the request object. If I did make it work without the request object it would require changes on the Rust side of things, so signatures would be checked before or during passing the args from the Rust side as the handler is being called
> > 
> > 
> > That kills the purpose of splitting the request params, unfortunately 😓
> 
> I can certainly make the necessary changes, it would simply move some, if not all of the changes over to the Rust side.

Thank you! But before you proceed. How would you approach this task? Move the splitting logic before `function.call0(..)` ?

kliwongan: I believe it would have to be placed there, yes, since if the request object isn't passed in to the handler, the only time to intercept it have to be prior to the call1

sansyrox: @kliwongan 
Could you not just pass the request object to the handler from Rust and split it on the Python end?

kliwongan: Oh yeah I suppose I can do that, thanks for pointing it out, I know exactly how do to it

kliwongan: @sansyrox I have some code working and there's a bit of an ambiguity issue; for example, when there is a path whose handler function has only one unannotated argument, it's assumed that the single argument is meant for the Request object. So if we have a function that only has one argument meant to accept, say, only the request body, do we either:

1. Name it body (so we reserve specific keywords/names in the function argument section for splitting the request)
2.  OR name it whatever we want, but annotate it to identify what we want in that param (may need to create new enums or classes to represent each component of a Request for those components that don't have dedicated ones already)

sansyrox: Hey @kliwongan 👋 

> do we either: Name it body (so we reserve specific keywords/names in the function argument section for splitting the request) OR name it whatever we want, but annotate it to identify what we want in that param (may need to create new enums or classes to represent each component of a Request for those components that don't have dedicated ones already)

We should do both. First check for annotations. If there are no annotations, we use the reserved keywords. We should prioritise annotations over reserved keywords.

kliwongan: Sounds like a plan!

VishnuSanal:
hi @kliwongan, can I take over this PR? cc: @sansyrox

sansyrox:
@VishnuSanal go ahead!

VishnuSanal:
superseded by #952

PR #898, opened 2024-07-16
fix:  cors not work for missing preflight options requests

pingf:
**Description**

This PR fixes #

1. preflight return status okey
2. only allow user set one site for origin

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -456,32 +456,39 @@ class SubRouter(Robyn):
         return super().options(self.__add_prefix(endpoint))
 
 
-def ALLOW_CORS(app: Robyn, origins: List[str]|str):
-    """Allows CORS for the given origins for the entire router."""
+# keep the func params compatible with current one
+def ALLOW_CORS(app: Robyn, origins: List[str] | str):
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMultipleAllowOriginNotAllowed
-    origin = origins[0] if isinstance(origins, list) else origins
+    if isinstance(origins, list):
+        origin = origins[0]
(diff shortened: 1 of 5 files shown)


sansyrox (commented):
Hey @pingf 👋 

Thanks for the PR. I have a few suggestions :D

robyn/__init__.py
```diff
@@ -467,6 +478,23 @@ def ALLOW_CORS(app: Robyn, origins: List[str]):
         app.add_response_header("Access-Control-Allow-Headers", "Content-Type, Authorization")
         app.add_response_header("Access-Control-Allow-Credentials", "true")
 
+        # https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
+        @app.before_request()
```
sansyrox, line 482: this is a very interesting way of implementation. The first time I am seeing this approach

robyn/__init__.py
```diff
@@ -456,9 +456,20 @@ def options(self, endpoint: str):
...
-def ALLOW_CORS(app: Robyn, origins: List[str]):
+# keep the func params compatible with current one
+# use the type definition for compatible with python before 3.10
+def ALLOW_CORS(app: Robyn, origins: Union[List[str], str]):
     """Allows CORS for the given origins for the entire router."""
-    for origin in origins:
+
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMultipleAllowOriginNotAllowed
+    if isinstance(origins, list):
+        origin = origins[0]
```
sansyrox, line 466: While this may avoid the issue, we still want all the origins to be considered.

Maybe we will have to include this test in the middleware implemented below.

sansyrox:
This is fixed in the latest version @pingf 😄

PR #901, opened 2024-07-20
feat: add support to multiple authentication

### LINKED ISSUE #708: Sub-router Authentication Capabilities
This feature request proposes the implementation of authentication capabilities for sub-routers in the Robyn web framework.

- While Robyn currently supports middleware for handling authentication at the main router level, there is a growing need for more granular control, specifically at the sub-router level. 
- This would allow different authentication strategies for different sections of an application, enhancing security and flexibility.
Currently Adding 
`python
frontend = SubRouter(__name__, prefix="/frontend")
`

**Use Case**
A common use case is in a web application with both public and private sections. Public routes (like a homepage or blog) require no authentication, while private routes (like user dashboards) require user authentication. Implementing sub-router authentication streamlines this process, allowing developers to define authentication strategies at a more specific level than the entire application.

yomaaf:
**Description**

This PR fixes #708

Hello @sansyrox. This is a split PR for support for multiple authentication, however it does not include support for subrouter. Because, as previously explained. When decorating is called, it registers the route. So the authentication handler is not registered on the subrouter because the subrouter already register the route before the authentication handler is registered. I'll include support for the subrouter in the nested router PR. But it will require this PR merge first. So check out this PR first.

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -2,6 +2,7 @@ import os
 import pathlib
 from collections import defaultdict
 from typing import Optional
+from base64 import b64encode
 
 from robyn import (
     Request,
@@ -13,7 +14,7 @@ from robyn import (
     serve_html,
     WebSocketConnector,
 )
-from robyn.authentication import AuthenticationHandler, BearerGetter, Identity
+from robyn.authentication import AuthenticationHandler, BearerGetter, Identity, BasicGetter
 from robyn.robyn import Headers
 from robyn.templating import JinjaTemplate
 
@@ -794,6 +795,34 @@ async def async_auth(request: Request):
     return "authenticated"
 
 
+@app.get("/sync/auth/basic", auth_required=True, auth_middleware_name="basic")
+def sync_auth_basic(request: Request):
+    assert request.identity is not None
+    assert request.identity.claims == {"key": "value"}
+    return "authenticated"
+
+
+@app.get("/async/auth/basic", auth_required=True, auth_middleware_name="basic")
+async def async_auth_basic(request: Request):
+    assert request.identity is not None
+    assert request.identity.claims == {"key": "value"}
+    return "authenticated"
+
+
+@app.get("/sync/auth/bearer-2", auth_required=True, auth_middleware_name="bearer-2")
+def sync_auth_bearer_2(request: Request):
+    assert request.identity is not None
+    assert request.identity.claims == {"key": "value"}
+    return "authenticated"
+
+
+@app.get("/async/auth/bearer-2", auth_required=True, auth_middleware_name="bearer-2")
+async def async_auth_bearer_2(request: Request):
+    assert request.identity is not None
+    assert request.identity.claims == {"key": "value"}
+    return "authenticated"
+
+
 # ===== Main =====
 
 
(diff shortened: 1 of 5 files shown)


sansyrox:
Hey @yomaaf 👋 

Thanks for the PR. However, I do not understand the problem you are addressing here.

What do you mean by "multiple authentication" and why is it needed in Robyn?

yomaaf:
> Hey @yomaaf 👋
> 
> Thanks for the PR. However, I do not understand the problem you are addressing here.
> 
> What do you mean by "multiple authentication" and why is it needed in Robyn?

Hi @sansyrox.

"Multiple authentication" refers to the capability of a system to support different methods or mechanisms for verifying the identity of a user. This can include a variety of authentication methods such as:

- **Password-based authentication:** The most common method, where users authenticate using a username and password.
- **Multi-factor authentication (MFA):** Requires users to provide two or more verification factors (e.g., password and OTP).
- **OAuth:** A protocol that allows third-party services to exchange tokens and authenticate users without sharing passwords.

### Why is it needed in Robyn?
   
1. **Role-specific Security Requirements:**
   - Certain routes, such as those accessible to superadmins, may require more stringent security measures like JWT bearer tokens to ensure high levels of security.
   - Other routes, which may have less critical data, can use simpler authentication methods like password-based authentication.

2. **Compliance with Access Control Policies:**
   - Different routes may have varying compliance requirements. Supporting multiple authentication handlers allows Robyn to meet these diverse requirements effectively.

3. **Flexibility for Different Use Cases:**
   - Different applications or user groups may have varying security and usability requirements. Multiple authentication options provide the necessary flexibility to meet these diverse needs.
 
 
 ### Addressing fina-joy's Issue #708 

```
While Robyn currently supports middleware for handling authentication at the main router level, there is a growing need for more granular control, specifically at the sub-router level.
This would allow different authentication strategies for different sections of an application, enhancing security and flexibility.
```

sansyrox:
Hey @yomaaf 👋 

There is only one authentication handler in Robyn so the users can override the implementation. I believe your PR will serve well as a Robyn Plugin like the following - https://robyn.tech/documentation/plugins

However, I don't know enough about auth and need some time to think about it. I will get back to you after the weekend 😄 

Meanwhile, if you have any items to help me with my research, I'd highly appreciate them.

Thanks again for all the hard work :D

user-for-download (approved):

user-for-download (changes requested):
need this functionality)

sansyrox (commented):
Thanks @yomaaf — multiple named auth handlers is a feature I want, and the one-default validation in `auth_handler_configured()` plus the new `BasicGetter` are nice touches. Two things before this can land: first, `name` is currently a required positional on `AuthenticationHandler.__init__`, which breaks every existing `Handler(token_getter=...)` call — can we default it (e.g. to the class name) so it's backward compatible? Second, `main` has drifted: `add_auth_middleware` now takes `(endpoint, route_type)` for method-aware middleware and the route decorators gained `openapi_name`/`openapi_tags`, so this needs a rebase to apply. A clearer error when `auth_middleware_name` doesn't match any registered handler would also help. The test coverage across basic/bearer is great — once it's rebased and the constructor stays compatible, I'm happy to take it.

Issue #1035, opened 2024-11-17
BaseRouter and Router have different signatures for add_route


dave42w:
### Bug Description

According to mypy None of the subclasses of BaseRouter correctly implement the abstractmethod add_route.

```
class BaseRouter(ABC):
    @abstractmethod
    def add_route(*args) -> Union[Callable, CoroutineType, WebSocket]: ...
```
vs
```
class Router(BaseRouter):
...
    def add_route(
        self,
        route_type: HttpMethod,
        endpoint: str,
        handler: Callable,
        is_const: bool,
        exception_handler: Optional[Callable],
        injected_dependencies: dict,
    ) -> Union[Callable, CoroutineType]:
```
vs
```
class MiddlewareRouter(BaseRouter):
...
    def add_route(
        self,
        middleware_type: MiddlewareType,
        endpoint: str,
        handler: Callable,
        injected_dependencies: dict,
    ) -> Callable:
```
vs
```
class WebSocketRouter(BaseRouter):
...
    def add_route(self, endpoint: str, web_socket: WebSocket) -> None:
        self.routes[endpoint] = web_socket
```

The mypy errors are:
```
Signature of "add_route" incompatible with supertype "BaseRouter"
...
     Superclass:
         def add_route(*args: BaseRouter) -> Callable[..., Any] | CoroutineType[Any, Any, Any] | WebSocket
     Subclass:
         def add_route(self, route_type: HttpMethod, endpoint: str, handler: Callable[..., Any], is_const: bool, exception_handler: Callable[..., Any] | None, injected_dependencies: dict[Any, Any]) -> Callable[..., Any] | CoroutineType[Any, Any, Any]
...
     Superclass:
         def add_route(*args: BaseRouter) -> Callable[..., Any] | CoroutineType[Any, Any, Any] | WebSocket
     Subclass:
         def add_route(self, middleware_type: MiddlewareType, endpoint: str, handler: Callable[..., Any], injected_dependencies: dict[Any, Any]) -> Callable[..., Any]
...
     Superclass:
         def add_route(*args: BaseRouter) -> Callable[..., Any] | CoroutineType[Any, Any, Any] | WebSocket
     Subclass:
         def add_route(self, endpoint: str, web_socket: WebSocket) -> None
```

I can't find any use of BaseRouter in the codebase.

Options:

1. fix the signatures in the 3 subclasses to match BaseRouter
1. remove the abstractmethod add_route from BaseRouter (all tests still pass)
[...]

PR #1038, opened 2024-11-19
Fix mypy warnings in router.py

dave42w:
## Description

This PR fixes most of the mypy warnings in router.py

## Summary

This PR does not fix https://github.com/sparckles/Robyn/issues/1035 where add_route is inconsistent between the BaseRouter(ABC) and it's subclasses.

The biggest changes were in Router. _format_response
- fixing re-use of local variable names but with a different type
- using `if "x" in headers` replaced with `if headers.contains("x")`
- missing type specifiers

Beyond that a few places with a missing dict type: for injected_dependencies x 2 and self.reoutes x 1

## PR Checklist

Please ensure that:

- [x ] The PR contains a descriptive title
- [x ] The PR contains a descriptive summary of the changes
- [x ] You build and test your changes before submitting a PR.
- [x ] You have added relevant documentation
- [ x] You have added relevant tests. We prefer integration tests wherever possible

## Pre-Commit Instructions:

- [x ] Ensure that you have run the [pre-commit hooks](https://github.com/sparckles/robyn#%EF%B8%8F-to-develop-locally) on your PR.

### DIFF
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -53,11 +53,10 @@ class Router(BaseRouter):
     ) -> Response:
         headers = Headers({"Content-Type": "text/plain"})
 
-        response = {}
         if isinstance(res, dict):
             # this should change
             headers = Headers({})
-            if "Content-Type" not in headers:
+            if not headers.contains("Content-Type"):
                 headers.set("Content-Type", "application/json")
 
             description = jsonify(res)
@@ -88,16 +87,16 @@ class Router(BaseRouter):
             if len(res) != 3:
                 raise ValueError("Tuple should have 3 elements")
             else:
-                description, headers, status_code = res
-                description = self._format_response(description).description
-                new_headers = Headers(headers)
-                if "Content-Type" in new_headers:
-                    headers.set("Content-Type", new_headers.get("Content-Type"))
+                description2, headers2, status_code2 = res
+                description2 = self._format_response(description2).description
+                new_headers: Headers = Headers(headers2)
+                if new_headers.contains("Content-Type"):
+                    headers2.set("Content-Type", new_headers.get("Content-Type"))
 
                 response = Response(
-                    status_code=status_code,
-                    headers=headers,
-                    description=description,
+                    status_code=status_code2,
+                    headers=headers2,
+                    description=description2,
                 )
         else:
             response = Response(
@@ -291,7 +290,7 @@ class MiddlewareRouter(BaseRouter):
         This method adds an authentication middleware to the specified endpoint.
         """
 
-        injected_dependencies = {}
+        injected_dependencies: dict = {}
 
         def decorator(handler):
             @wraps(handler)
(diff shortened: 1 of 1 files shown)


robyn/router.py
```diff
@@ -88,16 +87,16 @@ def _format_response(
...
-                description, headers, status_code = res
-                description = self._format_response(description).description
-                new_headers = Headers(headers)
-                if "Content-Type" in new_headers:
-                    headers.set("Content-Type", new_headers.get("Content-Type"))
+                description2, headers2, status_code2 = res
+                description2 = self._format_response(description2).description
+                new_headers: Headers = Headers(headers2)
+                if new_headers.contains("Content-Type"):
+                    headers2.set("Content-Type", new_headers.get("Content-Type"))
```
sansyrox, line 94: Hey @dave42w 👋 

Thank you for the PR :D
One question - 
why are we using `description2` or `headers2` here? Is it something that mypy suggests?

dave42w: They just have to be different names to the ones earlier in _format_response. mypy objects to reusing a variable (even if it has gone out of scope) with a different type. eg this is unacceptable

```
if x:
  y: int = 5
else:
  y: str = "five"
```

dave42w: @sansyrox are we ok to consider this point resolved?

sansyrox: Hey @dave42w 👋 

Apologies for the late review.

> They just have to be different names to the ones earlier in _format_response. mypy objects to reusing a variable (even if it has gone out of scope) with a different type. eg this is unacceptable

We can disable mypy for these lines then. 

> mypy objects to reusing a variable (even if it has gone out of scope) with a different type

This is not a good reason IMO. If the code makes sense in our repo, we should not alter it according to mypy.

dave42w: H'mm, it seems risky to me. We are deep in lots of nesting at this point. It would be easy for a code change to end up trying to use these variables later in the function (especially because we don't return early) and depending on the data have 2 different types.
I would like to follow this with a more significant refactor anyway to reduce the levels of indentation/nested if statements. At that point this will not be a problem.

robyn/router.py
```diff
@@ -53,11 +53,10 @@ def _format_response(
     ) -> Response:
         headers = Headers({"Content-Type": "text/plain"})
 
-        response = {}
         if isinstance(res, dict):
             # this should change
             headers = Headers({})
-            if "Content-Type" not in headers:
```
sansyrox, line 60: Hey @dave42w , would this not be false always? I think we have a logical bug in Robyn

dave42w: I agree. I was just fixing the

robyn/router.py
sansyrox: Alright then. but can we get rid of numbers in the variable names. I get a massive ick by that(I know and I am sorry) but something else without involving numbers?

PR #1045, opened 2024-11-22
Feat: add async support for auth handler

IA-PieroCV:
## Description

This PR add the possibility to run async authenticate on AuthenticationHandler.

## Summary

This PR changes adds the typing hints for Authentication handler in order to return an Identity that comes from a Sync process or a Coroutine (Async). Also, it modifies the add_auth_middleware function to run this authenticate using async.

FYI: This allows AuthenticationHandler to support async sessions on cache (on memory db like redis).
```python
class RedisAuthHandler(AuthenticationHandler):
    """Redis Robyn auth handler using async Redis."""

    # Inheritance on abstract method allows both async and sync child methods.
    async def authenticate(self, request: Request) -> Identity | None: # Python 3.10 typing syntax
        """Authenticate Handler in Redis."""
        token = self.token_getter.get_token(request)
        if not token:
            return None

        user_id = await RedisClient.get_user_from_token(token) # <-- Important part. To the user this is just straightforward
        if user_id:
            return Identity(claims={"user_id": str(user_id)})
        return None
```

## PR Checklist

Please ensure that:

- [ ] ~~You have added relevant documentation~~ Minor feature, so current API Reference explains the usage.
- [ ] ~~You have added relevant tests. We prefer integration tests wherever possible~~ Minor feature, so current auth testing worked for both scenarios (sync and async).

## Pre-Commit Instructions:

**EDIT: Add some code context**

### DIFF
--- a/robyn/authentication.py
+++ b/robyn/authentication.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Any, Coroutine, Optional, Union
 
 from robyn.robyn import Headers, Identity, Request, Response
 from robyn.status_codes import HTTP_401_UNAUTHORIZED
@@ -64,7 +64,7 @@ class AuthenticationHandler(ABC):
         )
 
     @abstractmethod
-    def authenticate(self, request: Request) -> Optional[Identity]:
+    def authenticate(self, request: Request) -> Union[Optional[Identity], Coroutine[Any, Any, Optional[Identity]]]:
         """
         Authenticates the user.
         :param request: The request object.
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -289,21 +289,32 @@ class MiddlewareRouter(BaseRouter):
         injected_dependencies: dict = {}
 
         def decorator(handler):
-            @wraps(handler)
-            async def inner_handler(request: Request, *args):
-                if not self.authentication_handler:
-                    raise AuthenticationNotConfiguredError()
+            if inspect.iscoroutinefunction(self.authentication_handler.authenticate):
 
-                if inspect.iscoroutinefunction(self.authentication_handler.authenticate):
+                @wraps(handler)
+                async def inner_handler(request: Request, *args):
+                    if not self.authentication_handler:
+                        raise AuthenticationNotConfiguredError()
+
                     identity = await self.authentication_handler.authenticate(request)
-                else:
+                    if identity is None:
+                        return self.authentication_handler.unauthorized_response
+                    request.identity = identity
+
+                    return request
+            else:
+
+                @wraps(handler)
+                def inner_handler(request: Request, *args):
+                    if not self.authentication_handler:
+                        raise AuthenticationNotConfiguredError()
+
                     identity = self.authentication_handler.authenticate(request)
+                    if identity is None:
+                        return self.authentication_handler.unauthorized_response
+                    request.identity = identity
 
-                if identity is None:
-                    return self.authentication_handler.unauthorized_response
-                request.identity = identity
-
-                return request
+                    return request
 
             self.add_route(
                 MiddlewareType.BEFORE_REQUEST,
--- a/robyn/router.py
+++ b/robyn/router.py
@@ -290,10 +290,15 @@ class MiddlewareRouter(BaseRouter):
 
         def decorator(handler):
             @wraps(handler)
-            def inner_handler(request: Request, *args):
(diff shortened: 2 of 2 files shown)


robyn/authentication.py
```diff
@@ -64,7 +64,7 @@ def unauthorized_response(self) -> Response:
         )
 
     @abstractmethod
-    def authenticate(self, request: Request) -> Optional[Identity]:
+    def authenticate(self, request: Request) -> Union[Optional[Identity], Coroutine[Any, Any, Optional[Identity]]]:
```
sansyrox, line 67: Hey @IA-PieroCV 👋 

Thanks for the PR but why do we need this type?

robyn/router.py
```diff
@@ -290,10 +290,15 @@ def add_auth_middleware(self, endpoint: str):
 
         def decorator(handler):
             @wraps(handler)
-            def inner_handler(request: Request, *args):
+            async def inner_handler(request: Request, *args):
```
sansyrox, line 293: and what about sync support?

IA-PieroCV: I tested with both sync and async and both worked pretty well...
However, I'm kind of worried not showing  support for both sync and async. I'll make some tests and will back with the results

sansyrox:
Hey @IA-PieroCV 👋 

I will have a proper review over the weekend 😄

robyn/authentication.py
sansyrox: @IA-PieroCV 👋 

Should we use a type alias instead? The return type is not super readable

robyn/router.py
sansyrox, line 1: Nicee

sansyrox (commented):
Hey @IA-PieroCV 👋 

I really like the implementation. Do we need to add some docs for this?

Also, would love some integration tests

IA-PieroCV:
Hey @sansyrox !
I guess I can add a tab to the mdx documentation with "sync"/"async" if this is okey. I don't think is needed to be explicit, but to mention it briefly and show the capability. Are you okey with this approach?

About integration test, of course! I'll implement it in the next couple of hours, it doesn't seem like a problem.

IA-PieroCV:
Hey @sansyrox!
As the middlewares are added on runtime the evaluation of coroutine cannot be made at first hand.
I know that we were discussing how endpoints could be registered at the start method and part of the PR I made on #1056 takes this approach too. This would solve this as the configuration should be implemented before registering the auth middleware.

I'm trying to look for an approach that solves this without the change on the registration approach, but this could take me some time. By now I will convert this PR to draft until I found a solution or having the endpoint registration on the start method (if it's done).

dave42w:
> Hey @sansyrox! As the middlewares are added on runtime the evaluation of coroutine cannot be made at first hand. I know that we were discussing how endpoints could be registered at the start method and part of the PR I made on #1056 takes this approach too. This would solve this as the configuration should be implemented before registering the auth middleware.

I like the approach that we basically register routes, middleware, auth, routers after `app = Robyn(...)` and before `app.start(...)`

Then in `app.start()`, before we call `run_processes()...)` we *prepare* everything that has been registered. This makes the registration much easier to test and it avoids any restrictions on the order developers register all the parts of their application. Once applications grow it makes it easier for whole departments with their own combinations of routes, middleware etc to be plugged into the whole with one include.

I think it is going to make the whole *prepare* step much easier. We choose when to process authentication, subrouters, routes etc rather than the order the code got put together.

sansyrox (commented):
Thanks @IA-PieroCV — async `authenticate()` is definitely something Robyn should support (great Redis example in the description). I've ended up implementing the same capability in #1331, where the wrapper is always `async` and uses `inspect.isawaitable()` on the result so it also handles handlers that return an awaitable without being declared `async def`, and it additionally propagates auth to SubRouters. Since this is still a draft with no tests and conflicts with the refactored `add_auth_middleware(endpoint, route_type)` on `main`, I'd suggest we close this in favor of #1331 to avoid duplicate work — but please do review #1331 and push back there if the `isawaitable` approach misses your use case. Really appreciate the clear writeup and the redis example, which I'll reference in the docs.

sansyrox:
Thanks for this, @IA-PieroCV! 🙏 Async `authenticate()` support has landed in #1407 (merged as 3529cadc): `AuthenticationHandler.authenticate()` can now be an `async def` and the auth middleware awaits it, so async ORMs / HTTP clients work in authentication. Closing in favor of the merged PR — really appreciate the contribution.

Issue #1059, opened 2024-11-29
Route simplication


dave42w:
I have been exploring the current chain of activity around setting up routes from from creating the Robyn app instance to starting the server. Details are at the bottom of this issue.

In following this chain I believe the Robyn.add_route method (called from the various decorators eg `@app.get("/")`) is doing too much and too early.

My simplification is to achieve the following

1. Zero API changes for developers of Robyn apps
2. More flexibility for developers of Robyn apps. They should be able to assemble routes, routers, middleware etc in any order before the server is started. Because we try to do all the work with routes from the `@get` decorator we have made it more complicated to handle the developer adding routes, subrouters, middleware etc in different orders.
3. Simpler code that is more aligned with SOLID principals

So this is the refactor I'm suggesting 
:
1. Robyn.add_route only calls the master (router created inside the Robyn instance) Router.add_route
2. Router.add_route gets called by Robyn when adding `@get` at the app level and directly when adding `@get` to a subrouter (or if you add routes without using the decorator).
3. Router.add_route should do the minimum to add the route to its `routes: List[Route] ` (ensuring that everything only known at this point is done).
4. We move all the processing of routes (OpenApi, Middleware, dealing with nesting, introspecting the handler etc) to `app.start()` where currently it calls `get_routes`. To do this we have a single `prepare_routes` method in the Router to do all the steps currently triggered by add_route and end up with a full list of routes ready to be passed to the spawning of processes. This makes it easier to load OpenAPI only when it is enabled. It also means that we can nest Routers to any level (`prepare_routes` calls itself for all nested routers (so we don't need the subrouter class - it can be left deprecated and empty for compatibility). The only special thing about the master/top-level router is that it is automatically created inside the app Robyn instance.

First, this is my understanding of how routes work at present. I would appreciate it if this could be checked.

1. Robyn instance created
```
from robyn import Robyn

app = Robyn(__file__)
```
At this point in `__init__.py` the app instance of Robyn is initialised to contain an instance of Router() (from router.py)
```
self.router = Router()
```
We have app.router which contains:
```
self.routes: List[Route] = []
```
Also in router.py we have the definition of Route as
```
class Route(NamedTuple):
    route_type: HttpMethod
    route: str
    function: FunctionInfo
    is_const: bool
```
6. Next we add a route to the app.
```
@app.get("/")
def h(request):
    return "Hello, world"
```
the get decorator is in Robyn in `__init_.py`
```
def get(
        self,
        endpoint: str,
[...]

PR #1066, opened 2024-12-01
refactor: openApi made lazy. 

dave42w:
Makes openAPI lazy (only created if enabled, all methods only called if enabled). 

Fixes a potential issue in __init__ where the default argument creates a shared object at parse time rather than a separate one for each instance. As only one instance, not normally an issue but not good practice.

One FIXME is an edge case in init_openapi where neither if clause evaluates to True for openapi_file_path (should we log an error or what?)

## Description

This PR is progressing towards the larger refactor to simplify the processing of add_route. See https://github.com/sparckles/Robyn/issues/1059

## Summary

This PR does make it so that 
1. the openApi instance variable is only created if needed
2. the openApi is created if needed (and not passed as argument) at Robyn.__init__ not at parse time
3. all the calls to add_openapi_path_obj are made in app.start so after all the routes have been fully sorted (only if openApi is enabled)

Please look carefully at the TODO We should cover this edge case
```
        if self.openapi_file_path:
            self.openapi.override_openapi(Path(self.directory_path).joinpath(self.openapi_file_path))
        elif Path(self.directory_path).joinpath("openapi.json").exists():
            self.openapi.override_openapi(Path(self.directory_path).joinpath("openapi.json"))
        # TODO! what about when the elif fails?

```

## PR Checklist

Please ensure that:

- [X] You have added relevant documentation {No API changes}
- [X] You have added relevant tests. We prefer integration tests wherever possible

## Pre-Commit Instructions:

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -43,7 +43,7 @@ class Robyn:
         file_object: str,
         config: Config = Config(),
         openapi_file_path: Optional[str] = None,
-        openapi: OpenAPI = OpenAPI(),
+        openapi: Optional[OpenAPI] = None,
         dependencies: DependencyMap = DependencyMap(),
     ) -> None:
         directory_path = os.path.dirname(os.path.abspath(file_object))
@@ -51,12 +51,10 @@ class Robyn:
         self.directory_path = directory_path
         self.config = config
         self.dependencies = dependencies
+        self.openapi_file_path = openapi_file_path
         self.openapi = openapi
 
-        if openapi_file_path:
-            openapi.override_openapi(Path(self.directory_path).joinpath(openapi_file_path))
-        elif Path(self.directory_path).joinpath("openapi.json").exists():
-            openapi.override_openapi(Path(self.directory_path).joinpath("openapi.json"))
+        self.init_openapi()
 
         if not bool(os.environ.get("ROBYN_CLI", False)):
             # the env variables are already set when are running through the cli
@@ -82,6 +80,19 @@ class Robyn:
         self.event_handlers: dict = {}
         self.exception_handler: Optional[Callable] = None
         self.authentication_handler: Optional[AuthenticationHandler] = None
+
+    def init_openapi(self) -> None:
+        if self.config.disable_openapi:
+            return
+
+        if self.openapi is None:
+            self.openapi = OpenAPI()
+
+        if self.openapi_file_path:
+            self.openapi.override_openapi(Path(self.directory_path).joinpath(self.openapi_file_path))
+        elif Path(self.directory_path).joinpath("openapi.json").exists():
+            self.openapi.override_openapi(Path(self.directory_path).joinpath("openapi.json"))
+        # TODO! what about when the elif fails?
 
     def _handle_dev_mode(self):
         cli_dev_mode = self.config.dev  # --dev
@@ -240,6 +251,13 @@ class Robyn:
             raise Exception(f"Invalid port number: {port}")
 
     def _add_openapi_routes(self, auth_required: bool = False):
+        if self.config.disable_openapi:
+            return
+
+        if self.openapi is None:
+            logger.error("No openAPI")
+            return
+
         self.add_route(
             route_type=HttpMethod.GET,
             endpoint="/openapi.json",
@@ -369,7 +387,8 @@ class Robyn:
         """
 
         def inner(handler):
-            self.openapi.add_openapi_path_obj("get", endpoint, openapi_name, openapi_tags, handler)
+            if not self.config.disable_openapi and self.openapi is not None:
+                self.openapi.add_openapi_path_obj("get", endpoint, openapi_name, openapi_tags, handler)
 
             return self.add_route(HttpMethod.GET, endpoint, handler, const, auth_required)
 
@@ -392,7 +411,8 @@ class Robyn:
         """
 
         def inner(handler):
-            self.openapi.add_openapi_path_obj("post", endpoint, openapi_name, openapi_tags, handler)
+            if not self.config.disable_openapi and self.openapi is not None:
+                self.openapi.add_openapi_path_obj("post", endpoint, openapi_name, openapi_tags, handler)
 
             return self.add_route(HttpMethod.POST, endpoint, handler, auth_required=auth_required)
 
@@ -415,7 +435,8 @@ class Robyn:
         """
 
         def inner(handler):
-            self.openapi.add_openapi_path_obj("put", endpoint, openapi_name, openapi_tags, handler)
+            if not self.config.disable_openapi and self.openapi is not None:
+                self.openapi.add_openapi_path_obj("put", endpoint, openapi_name, openapi_tags, handler)
 
             return self.add_route(HttpMethod.PUT, endpoint, handler, auth_required=auth_required)
 
@@ -438,7 +459,8 @@ class Robyn:
         """
 
         def inner(handler):
-            self.openapi.add_openapi_path_obj("delete", endpoint, openapi_name, openapi_tags, handler)
+            if not self.config.disable_openapi and self.openapi is not None:
+                self.openapi.add_openapi_path_obj("delete", endpoint, openapi_name, openapi_tags, handler)
 
             return self.add_route(HttpMethod.DELETE, endpoint, handler, auth_required=auth_required)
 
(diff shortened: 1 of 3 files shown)


dave42w:
Second Commit https://github.com/sparckles/Robyn/pull/1066/commits/642eac9c6f98e955d0a383fd8fe8de4d4fec5444

This should reduce the performance problem as we no longer add openapi routes in the decorators. Instead we store the openapi_name and openapi_tags (plus auth_rerquired because it will be needed) in the Route.
So we can build all the openapi routes together in app.start (if they are needed)

All tests are passing. This will complete the openapi part of the add_routes refactor.

dave42w:
@sansyrox if you are happy with this direction then this PR is ready

Once this is in I'll start another PR for the add_router next step (which will keep routes local to the router/subrouter until we start the server).

robyn/__init__.py
```diff
@@ -369,9 +396,7 @@ def get(
         """
 
         def inner(handler):
-            self.openapi.add_openapi_path_obj("get", endpoint, openapi_name, openapi_tags, handler)
-
-            return self.add_route(HttpMethod.GET, endpoint, handler, const, auth_required)
```
sansyrox, line 374: Hey @dave42w 👋 

Could you explain this modification a bit?

robyn/router.py
```diff
@@ -108,6 +116,9 @@ def add_route(  # type: ignore
         endpoint: str,
         handler: Callable,
         is_const: bool,
+        auth_required: bool,
+        openapi_name: str,
+        openapi_tags: List[str],
```
sansyrox, line 121: why do middleware routes need to have open_api names?

robyn/router.py
```diff
@@ -19,11 +20,18 @@
...
+
+
 class Route(NamedTuple):
     route_type: HttpMethod
     route: str
     function: FunctionInfo
     is_const: bool
+    auth_required: bool
+    openapi_name: str
+    openapi_tags: List[str]
```
sansyrox, line 34: could you share some thoughts behind this implementation?

sansyrox (commented):
Hey @dave42w 👋 

Thank you for the PR 😄 

I have some questions regarding the implementations

robyn/__init__.py
dave42w: Sure.

My goal is that we do the minimum processing of routes while in the app setup stage. ie between `app = Robyn(__file__)` and `app.start(...)`

My logic is that things can happen in quite a dynamic way in this section of the code, developers might do things in an odd order (include a subrouter and then add more routes to it). Or include a subrouter in a subrouter and then include the subrouter and all it's subsubrouters. Or add middleware before the routes or in the middle of them. This could happen because we bring in a subrouter via a module and there it includes routes and middleware. Or want different middleware for the subrouter from accounting compared to sales etc.

Therefore if we call `add_openapi_path_obj` in add_route the path might not yet be fixed. So it takes more logic (and potentially doing things multiple times) to handle these variations.

When we have calls to openapi scattered around the code then it gets expensive in runtime and maintenance if in every case we want to only make the call if openapi is enabled. It gets slow to maintain if we have 10 calls to `add_openapi_path_obj` and need to change it.

Of course if we are going to collect together all the calls to `add_openapi_path_obj` and do them in one place as the server starts then we need to have the handler, openapi_name and openapi_tags available during `app.start()`. This is why I have added them to the Route class.

I'm pretty sure that when I refactor add_route it will be a lot cleaner if we create an incomplete Route instance much earlier and then pass the single object around the system rather than 7 arguments. Deep in the system we do the work to complete function_info to finish the Route instance. That will also reduce the need for parsing the arguments all over the place (maybe Route needs to be a dataclass to validate itself). All that is for a future PR.

robyn/router.py
dave42w: I don't think I've added openapi to middleware routes. MiddlewareRouter starts at line 266. These lines (120&121) are in Router.

robyn/router.py
dave42w: I hope my earlier comment https://github.com/sparckles/Robyn/pull/1066#discussion_r1866869341 explains most of this.

In Robyn.add_route we were allowing a None value for openapi_tags. Rather than deal with the Union all the way through I remove the possibility of None for easier working with types.

You can see Robyn.start already called `Robyn._add_openapi_routes` that now calls `Router.prepare_routes_openapi` for the Robyn.router The prepare then goes through every direct route and every route in all the included subrouters to call `add_openapi_path_obj`

dave42w:
> Hey @dave42w 👋
> 
> Thank you for the PR 😄
> 
> I have some questions regarding the implementations

Hi @sansyrox 

I hope my answers are clear, useful and correct :-)

robyn/router.py
dave42w: I also want to add that this is on the way to making everything that is part of add_route and include_routes much easier to unit_test. The calls to `add_openapi_path_obj` had become side effects when adding a route and so harder to unit test.

dave42w:
Hi @sansyrox 
Please can we merge this. It is passing all the tests (with 5 speed regressions all initialisation related and so not affecting runtime performance). Once it is merged. I can work on the add_route stuff which should significantly improve the code base.

Thanks

robyn/processpool.py
```diff
@@ -182,7 +182,7 @@ def spawn_process(
     server.set_response_headers_exclude_paths(excluded_response_headers_paths)
 
     for route in routes:
-        route_type, endpoint, function, is_const = route
+        route_type, endpoint, function, is_const, auth_required, openapi_name, openapi_tags = route
```
sansyrox, line 185: One question @dave42w 👋 

Where are the `auth_required`, `openapi_name` , etc used here? Could we do a better job in unpacking if now?

robyn/__init__.py
```diff
@@ -82,6 +79,20 @@ def __init__(
...
         self.exception_handler: Optional[Callable] = None
         self.authentication_handler: Optional[AuthenticationHandler] = None
+        self.included_routers: List[Router] = []
+
+    def init_openapi(self, openapi_file_path: Optional[str]) -> None:
+        if self.config.disable_openapi:
+            return
+
+        if self.openapi is None:
+            self.openapi = OpenAPI()
```
sansyrox, line 89: will this ever happen?

robyn/router.py
```diff
@@ -236,9 +247,18 @@ def inner_handler(*args, **kwargs):
...
 
+    def prepare_routes_openapi(self, openapi: OpenAPI, included_routers: List) -> None:
+        for route in self.routes:
+            openapi.add_openapi_path_obj(lower_http_method(route.route_type), route.route, route.openapi_name, route.openapi_tags, route.function.handler)
+
+        # TODO! after include_routes does not immediatelly merge all the routes
+        # for router in included_routers:
+        #    for route in router:
+        #        openapi.add_openapi_path_obj(lower_http_method(route.route_type), route.route, route.openapi_name, route.openapi_tags, route.function.handler)
+
```
sansyrox, line 261: @dave42w , what do we mean by this comment?

sansyrox (commented):
Hey @dave42w 👋 

I've gone through the changes and all looks good. I just have a few comments and suggestions 😄 

Merry shipmas btw 😉 🎄

robyn/router.py
dave42w: At the moment, when we include_routes all the routes from a subrouter get copied into the Robyn router immediately. This means we shouldn't add extra routes after calling include_routes.
The full routes refactor will move all evaluation and processing of routes and routers to when the server starts. So add_route is just adding route objects to the current router. Also include_routes just adds a router to the parent without copying any routes. This massively simplifies these processes and means that the ordering of adding routes and routers does not matter.
The server start then builds the complete route table, handling nesting of routers, open_api registration, authentication, middleware after the application developer has registered everything in any order.
The rust processes will get the same routes as they do now but we do all the route processing once in one place.
The comment is a placeholder to remind me not to forget to make sure that openapi still handles things properly (it might not be this code depending on the order that seems best for handling both include _routes and add_openapi_path_obj).

robyn/__init__.py
dave42w: It shouldn't but it (probably temporarily pending the full routes refactor) keeps mypy happy because of the use of Optional in the argument list `openapi_file_path: Optional[str]`

robyn/processpool.py
dave42w: It's preparation for the full route refactor. Then the authentication route processing happens in server start rather than in add_route.
I wanted to reduce the times I need to touch the code and so put this in before it's needed.

dave42w:
> Hey @dave42w 👋
> 
> I've gone through the changes and all looks good. I just have a few comments and suggestions 😄
> 
> Merry shipmas btw 😉 🎄

Thanks. I've replied to all the comments. Hopefully, this is just a stepping stone to the bigger router refactor where the real benefits lie (much simpler functions without side effects; support for nesting subrouters; support for building routes/routers in more flexible ways)

Issue #1148, opened 2025-03-25
This not support handle websocket binary msg


raprek:
There is not support to handle binary msg via web. Now you have echo response on binary. Also you can see it [HERE](https://github.com/sparckles/Robyn/blob/f1ab513d2640a1a853a86d3a72ad255832853ee3/src/websockets/mod.rs#L99). May you add it?

raprek:
#1149

PR #1149, opened 2025-03-25
Add ws binary

raprek:
## Description

Add support to handle binary Websocket msg

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ name = "robyn"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-pyo3 = { version = "0.20.0", features = ["extension-module",] }
+pyo3 = { version = "0.20.0", features = ["extension-module"] }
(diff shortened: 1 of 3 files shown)


sansyrox (commented):
Thanks for tackling binary WebSocket support, @raprek — this is a genuinely useful gap and #1148 is a real pain point. Since you opened this, our WebSocket internals were rewritten to a `tokio::mpsc` channel model, so the `execute_ws_function(..., Option<String>)` path this patch modifies no longer exists, which is why it now shows conflicts. I've also put up #1334, which carries a `WsPayload::Text/Binary` enum end-to-end so `receive()` returns real `bytes`, `send_bytes()` emits true binary frames, and `broadcast()` accepts bytes — plus the Python wrapper, type stubs, and docs. Because #1334 supersedes this end-to-end, I'm planning to land binary support there and would recommend we close this one — but I really appreciate you surfacing the issue and pushing on it.

PR #1184, opened 2025-06-16
feat: add a to_dict method in headers class

### LINKED ISSUE #1099: Add `to_dict` method for request headers
Hello, first of all thank you for this marvelous framework, it is really enjoyable to work with especially as I'm coming from Flask.

I noticed that there are methods for returning dictionaries for query parameters, path parameters and form data are also dictionaries. Suggestion is because currently we are losing the ability to set a default value if the header is missing, if it is - we get always None, but what if we need to have some default value?

My idea is to have something like this
```
request.headers.to_dict()
```
which will return a dictionary with all present headers
```
{
    "content-type": "application/json",
    "X-Forwarded-To": "some IP",
    ...
}
```
Is this possible to achieve?

sansyrox:
## Description

This PR fixes #1099

## Summary

This PR adds a `to_dict` method in the headers

## PR Checklist

Please ensure that:

- [ ] The PR contains a descriptive title
- [ ] The PR contains a descriptive summary of the changes
- [ ] You build and test your changes before submitting a PR.

## Pre-Commit Instructions:

- [ ] Ensure that you have run the [pre-commit hooks](https://github.com/sparckles/robyn#%EF%B8%8F-to-develop-locally) on your PR.

### DIFF
--- a/.github/workflows/release-CI.yml
+++ b/.github/workflows/release-CI.yml
@@ -195,7 +195,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           command: sdist
-          args: --out dist
+          args: -i python --out dist
       - name: Publish to PyPi
         env:
           TWINE_USERNAME: __token__
--- a/.github/workflows/release-CI.yml
+++ b/.github/workflows/release-CI.yml
@@ -181,6 +181,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [macos, windows, linux, linux-cross, merge]
     steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


k4976:
Can we like this too?
```
req.headers.get_headers().get("key", "default_value")?

# Also, like this too
req.headers.get("content-type", "text")
If have this key then it will return value else, it will return text as value
```

To add functionality in .get method of headers, query_params, etc where it can accept default value too

@sansyrox

sansyrox:
Closing as a duplicate of the headers to_dict effort; consolidating into a single fresh implementation.

PR #1205, opened 2025-07-10
Built-in Database Migration Tool For Robyn

Charlie-BU:
## Description

This PR adds a built-in database migration tool to Robyn, powered by Alembic. It introduces a new `robyn db` CLI command group that provides a full suite of migration capabilities, including `init`, `migrate`, `upgrade`, `downgrade`, `stamp`, and more.

## Summary

This PR does the following:

- Adds a fully integrated migration CLI (`robyn db`) for SQLAlchemy-based projects
- Wraps Alembic’s command interface with a consistent Robyn-style CLI structure
- Auto-generates and configures `alembic.ini` and `env.py` based on user input or intelligent detection (e.g., database URL, model metadata path)
- Supports common Alembic commands: `init`, `revision`, `migrate`, `upgrade`, `downgrade`, `merge`, `edit`, `stamp`, `show`, `history`, etc.
- Includes support for migration templates (e.g., `--template robyn`)
- Handles command-line errors gracefully with user-friendly messages
- Provides extension class `Migrate` that can be optionally used inside Robyn apps for future integration

This aims to make database versioning and schema evolution first-class citizens in the Robyn developer workflow, reducing friction for backend developers.

## PR Checklist

Please ensure that:

- [x] You have added relevant documentation  

## Pre-Commit Instructions:

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -39,12 +39,12 @@ class BaseRobyn(ABC):
     """This is the python wrapper for the Robyn binaries."""
 
     def __init__(
-            self,
-            file_object: str,
-            config: Config = Config(),
-            openapi_file_path: Optional[str] = None,
-            openapi: Optional[OpenAPI] = None,
-            dependencies: DependencyMap = DependencyMap(),
+        self,
+        file_object: str,
+        config: Config = Config(),
+        openapi_file_path: Optional[str] = None,
+        openapi: Optional[OpenAPI] = None,
+        dependencies: DependencyMap = DependencyMap(),
     ) -> None:
         directory_path = os.path.dirname(os.path.abspath(file_object))
         self.file_path = file_object
@@ -101,24 +101,21 @@ class BaseRobyn(ABC):
         is_robyn = os.getenv("ROBYN_CLI", False)
 
         if cli_dev_mode and not is_robyn:
-            raise SystemExit(
-                "Dev mode is not supported in the python wrapper. Please use the Robyn CLI. e.g. python3 -m robyn app.py --dev")
+            raise SystemExit("Dev mode is not supported in the python wrapper. Please use the Robyn CLI. e.g. python3 -m robyn app.py --dev")
 
         if env_dev_mode and not is_robyn:
-            logger.error(
-                "Ignoring ROBYN_DEV_MODE environment variable. Dev mode is not supported in the python wrapper.")
-            raise SystemExit(
-                "Dev mode is not supported in the python wrapper. Please use the Robyn CLI. e.g. python3 -m robyn app.py")
+            logger.error("Ignoring ROBYN_DEV_MODE environment variable. Dev mode is not supported in the python wrapper.")
+            raise SystemExit("Dev mode is not supported in the python wrapper. Please use the Robyn CLI. e.g. python3 -m robyn app.py")
 
     def add_route(
-            self,
-            route_type: Union[HttpMethod, str],
-            endpoint: str,
-            handler: Callable,
-            is_const: bool = False,
-            auth_required: bool = False,
-            openapi_name: str = "",
-            openapi_tags: Union[List[str], None] = None,
+        self,
+        route_type: Union[HttpMethod, str],
+        endpoint: str,
+        handler: Callable,
+        is_const: bool = False,
+        auth_required: bool = False,
+        openapi_name: str = "",
+        openapi_tags: Union[List[str], None] = None,
     ):
         """
         Connect a URI to a handler
@@ -202,11 +199,11 @@ class BaseRobyn(ABC):
         return self.middleware_router.add_middleware(MiddlewareType.AFTER_REQUEST, endpoint)
 
     def serve_directory(
-            self,
-            route: str,
-            directory_path: str,
-            index_file: Optional[str] = None,
-            show_files_listing: bool = False,
+        self,
+        route: str,
+        directory_path: str,
+        index_file: Optional[str] = None,
+        show_files_listing: bool = False,
     ):
         """
         Serves a directory at the given route
@@ -291,12 +288,12 @@ class BaseRobyn(ABC):
         self.exception_handler = exception_handler
 
     def get(
-            self,
-            endpoint: str,
-            const: bool = False,
-            auth_required: bool = False,
-            openapi_name: str = "",
-            openapi_tags: List[str] = ["get"],
+        self,
+        endpoint: str,
+        const: bool = False,
+        auth_required: bool = False,
+        openapi_name: str = "",
+        openapi_tags: List[str] = ["get"],
     ):
         """
         The @app.get decorator to add a route with the GET method
@@ -314,11 +311,11 @@ class BaseRobyn(ABC):
         return inner
 
     def post(
-            self,
-            endpoint: str,
-            auth_required: bool = False,
-            openapi_name: str = "",
-            openapi_tags: List[str] = ["post"],
+        self,
+        endpoint: str,
+        auth_required: bool = False,
+        openapi_name: str = "",
+        openapi_tags: List[str] = ["post"],
     ):
         """
         The @app.post decorator to add a route with POST method
@@ -330,17 +327,16 @@ class BaseRobyn(ABC):
         """
 
         def inner(handler):
-            return self.add_route(HttpMethod.POST, endpoint, handler, auth_required=auth_required,
-                                  openapi_name=openapi_name, openapi_tags=openapi_tags)
+            return self.add_route(HttpMethod.POST, endpoint, handler, auth_required=auth_required, openapi_name=openapi_name, openapi_tags=openapi_tags)
 
         return inner
 
     def put(
-            self,
-            endpoint: str,
-            auth_required: bool = False,
-            openapi_name: str = "",
-            openapi_tags: List[str] = ["put"],
+        self,
+        endpoint: str,
+        auth_required: bool = False,
+        openapi_name: str = "",
+        openapi_tags: List[str] = ["put"],
     ):
         """
         The @app.put decorator to add a get route with PUT method
@@ -352,17 +348,16 @@ class BaseRobyn(ABC):
         """
 
         def inner(handler):
-            return self.add_route(HttpMethod.PUT, endpoint, handler, auth_required=auth_required,
-                                  openapi_name=openapi_name, openapi_tags=openapi_tags)
+            return self.add_route(HttpMethod.PUT, endpoint, handler, auth_required=auth_required, openapi_name=openapi_name, openapi_tags=openapi_tags)
 
         return inner
 
     def delete(
-            self,
-            endpoint: str,
-            auth_required: bool = False,
-            openapi_name: str = "",
-            openapi_tags: List[str] = ["delete"],
+        self,
+        endpoint: str,
+        auth_required: bool = False,
+        openapi_name: str = "",
+        openapi_tags: List[str] = ["delete"],
     ):
         """
         The @app.delete decorator to add a route with DELETE method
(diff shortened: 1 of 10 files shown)


robyn/__init__.py
```diff
@@ -26,6 +26,14 @@
...
 from robyn.ws import WebSocket
 
+import importlib.util
+
+spec = importlib.util.find_spec("alembic")
+if spec is not None and spec.loader is not None:
+    from robyn.migrate import Migrate
+else:
+    Migrate = None
+
```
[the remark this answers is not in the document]
Charlie-BU: I was considering that user would import this `Migrate` to connect migration with his robyn app, like this:
image

But later i found a better way, which is using `robyn db ...` directly in cli. So there's no need to export `Migrate` as an extern class for users to call in their `app.py`. So i'll delete it. Thx for finding the problem!

robyn/cli.py
```diff
@@ -3,6 +3,7 @@
 import subprocess
 import sys
 import webbrowser
+import argparse
```
sansyrox, line 6: do we need this?

robyn/cli.py
```diff
@@ -71,6 +83,14 @@ def create_robyn_app():
     if docker == "N":
         os.remove(f"{final_project_dir_path}/Dockerfile")
 
+    # If database migration is needed, install the latest version of alembic
+    if db_migration == "Y":
+        print("Installing the latest version of alembic...")
+        try:
+            subprocess.run([sys.executable, "-m", "pip", "install", "alembic", "-q"], check=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
sansyrox, line 90: there should be a better way here, no?

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+
+        Args:
+            template: Template name or path, defaults to None
+
+        Returns:
+            str: Path to the template directory
+        """
+        if template is None:
+            return Path(__file__).parent / "templates" / "robyn"
+        return Path(template)
```
sansyrox, line 41: This path validation might be incorrect.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+def _auto_configure_migrations(directory: str, db_url: Optional[str] = None, model_path: Optional[str] = None) -> None:
+    """自动配置 alembic.ini 和 env.py 文件
+
+    Args:
+        directory: 迁移文件目录
+        db_url: 数据库 URL
+        model_path: 模型文件路径
+    """
```
sansyrox, line 91: please use english here.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+
+def _special_configure_for_sqlite(directory: str, model_path: Optional[str] = None):
+    # If the database is SQLite, must add render_as_batch=True in run_migrations_online() to avoid migration errors caused by SQLite's limited support for ALTER TABLE.
+    if model_path:
+        env_py_path = os.path.join(directory, "env.py")
+        if os.path.exists(env_py_path):
+            with open(env_py_path, "r") as f:
+                content = f.read()
+            try:
+                content = content.replace(
```
sansyrox, line 147: same with this replace.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+                # Try to import models module dynamically
+                spec = importlib.util.find_spec("models")
+                if spec is not None:
+                    models_module = importlib.util.module_from_spec(spec)
+                    spec.loader.exec_module(models_module)
+                    db_url = str(models_module.engine.url)
+                else:
+                    print('Cannot find models module.\nPlease provide your database URL with "--db-url=<YOUR_DB_URL>".')
+                    return
+        except Exception:
```
sansyrox, line 200: let's not use a general `Exception`.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+                spec.loader.exec_module(models_module)
+                if hasattr(models_module, "Base"):
+                    model_path = "models.Base"
+                else:
+                    print('Models module does not have Base attribute.\nPlease provide your model path with "--model-path=<YOUR_MODEL_PATH>".')
+                    return
+            else:
+                print('Cannot find models module.\nPlease provide your model path with "--model-path=<YOUR_MODEL_PATH>".')
+                return
+        except Exception:
```
sansyrox, line 219: general catch all

sansyrox:
Hey @Charlie-BU 👋 

Thanks for the PR. This is a good start :D But I have some inline comments.

robyn/migrate.py
Charlie-BU: solved :D

robyn/migrate.py
Charlie-BU: solved :D. Using regex instead.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+                # Replace the import statement
+                content = content.replace(
+                    "# add your model's MetaData object here\n# for 'autogenerate' support\n# from myapp import mymodel\n# target_metadata = mymodel.Base.metadata",
+                    f"# add your model's MetaData object here\n# for 'autogenerate' support\n{import_statement}",
+                )
+
+                # Replace the target_metadata setting
+                content = content.replace(
+                    "target_metadata = config.attributes.get('sqlalchemy.metadata', None)",
+                    "# target_metadata = config.attributes.get('sqlalchemy.metadata', None)\n# Already set by the import above",
```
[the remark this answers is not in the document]
Charlie-BU: This `env.py` file and `alembic.ini` file are automatically created by alembic when handling `alembic init alembic`, which is `robyn db init` now. And we have to modify the configuration here to adapt it to the real case.

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+    """
+    # Configure alembic.ini
+    if db_url:
+        alembic_ini_path = os.path.join(directory, "alembic.ini")
+        if os.path.exists(alembic_ini_path):
+            with open(alembic_ini_path, "r") as f:
+                content = f.read()
+
+            # Replace the database URL
+            content = content.replace("sqlalchemy.url = driver://user:pass@localhost/dbname", f"sqlalchemy.url = {db_url}")
```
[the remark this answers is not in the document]
Charlie-BU: solved :D. Using regex instead.

robyn/cli.py
Charlie-BU: I think we need this to define a sub-argument for robyn, which is `robyn db ...`. We might need `argparse` to process the argument passed after `robyn db`.
image

robyn/cli.py
Charlie-BU: I was considering to add `alembic` as a dependency when people execute `pip install robyn`. But in view that not everyone needs this database migration tool, I decided to make it an option for people when they do `robyn create`. If they said yes then `alembic` is a must in this case.
I modified this self-installing process like this:
image
Pls check if there's something that could be optimized. Thx! :D

robyn/migrate.py
```diff
@@ -0,0 +1,624 @@
...
+
+    @wraps(f)
+    def wrapped(*args, **kwargs):
+        try:
+            return f(*args, **kwargs)
+        except Exception as e:
+            print(f"ERROR: {str(e)}")
+            sys.exit(1)
+
+    return wrapped
```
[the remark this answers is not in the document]
Charlie-BU: Let me think how to refactor this...😊

robyn/migrate.py
Charlie-BU: Could you possibly explain more? I think it works.

robyn/migrate.py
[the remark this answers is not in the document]
sansyrox: @Charlie-BU  , so do we need to commit this file?

robyn/cli.py
sansyrox: @Charlie-BU , maybe we can allow it to be installed like `pip install robyn[db]` and then it installs alembic too?

robyn/cli.py
```diff
@@ -71,6 +83,36 @@ def create_robyn_app():
...
+                    # If pip API is not available, use subprocess
+                    print("Installing alembic using subprocess...")
+                    subprocess.run([sys.executable, "-m", "pip", "install", "alembic", "-q"], check=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+                    print("Successfully installed alembic.")
+            else:
+                print("Alembic is already installed.")
+        except (subprocess.CalledProcessError, ImportError) as e:
+            print(f"Failed to install alembic: {str(e)}")
+            print("Please install it manually using 'pip install alembic'.")
+
```
sansyrox, line 115: I think we can get rid of this if we decide to use alembic as an optional dependency.

robyn/cli.py
```diff
@@ -101,6 +143,70 @@ def start_app_normally(config: Config):
     subprocess.run(command, start_new_session=False)
 
 
+def handle_db_command():
```
sansyrox, line 146: we are importing a lot of things in this function. let's do that on the top

robyn/templates/robyn/env.py
sansyrox, line 1: is this file autogenerated @Charlie-BU ?

Charlie-BU: yeah, autogenerated by `alembic init alembic`.

robyn/cli.py
Charlie-BU: Sure!

robyn/cli.py
Charlie-BU: NP. Then people gotta install alembic as they do `pip install robyn`. Right?

robyn/migrate.py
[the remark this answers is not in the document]
Charlie-BU: I'm really sorrrrrrrry I'm not quite familiar about to design a separate module. I don't know if i need to redesign the whole robyn-db-migration structure to do that?

robyn/migrate.py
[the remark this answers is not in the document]
Charlie-BU: yeah we do. That's a template whose path is needed passing into `alembic init alembic`, which is `command.init(config, directory, template=template_path, package=package)` in our case. 
image

So that's why I wrote a `_get_template_path()` method in class `Config`. It's a must when we initialize the alembic using python code instead of command `alembic init alembic`.
image

robyn/cli.py
Charlie-BU: Cool! I removed this as you said. And set the `pyproject.toml` like this. So people will install alembic when they do `pip install robyn[db]`
image

image

robyn/migrate.py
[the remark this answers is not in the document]
Charlie-BU: I optimized my decorator. Let it catch specific exception while executing alembic commands with `command.*`, and inform the exception detail. Thanks for your advice!!:D

sansyrox (commented):
Thank you so much for this thorough piece of work, @Charlie-BU — there's clearly a lot of care here. My main hesitation is scope: Robyn is intentionally ORM-agnostic and doesn't ship a database layer, so bundling a SQLAlchemy/Alembic-specific migration CLI is a big surface area for us to own long-term, and it largely overlaps with Alembic's own CLI. There's also a blocking correctness issue: `robyn/cli.py` now imports `robyn/migrate.py` unconditionally, and `migrate.py` calls `exit(1)` at import time when Alembic isn't installed — that would break every `robyn` command (create/dev/run) for users who don't have the optional `db` extra. I'd also rather avoid `pip._internal` and runtime auto-installs, and the regex-based rewriting of `env.py`/`alembic.ini` is quite brittle. Given the scope question plus the missing tests, I'm inclined to recommend we close this in favor of documenting an Alembic recipe instead, but I'm very happy to discuss if you feel strongly about it being first-class.

Charlie-BU:
@sansyrox  Thanks for reviewing my unnatural idea. I'd like to discuss your suggestion but unfortunately i'm really in a busy time now. If it's okay for you i'd like to open another PR about alembic documentation in a few months.

PR #1232, opened 2025-08-07
refactor: necessary refactoring of pyproject.toml

thePromger:
* Adds requires-python tag/attributes which helps users+build_tool/pip to find minimum version supported

* Changes version constraint from exact to lowest version supported to be used

* Adding uv.lock file to be used in CI/CD or for development by some other maintainers

## PR Checklist

Please ensure that:

## Pre-Commit Instructions:

- [ ] Ensure that you have run the [pre-commit hooks](https://github.com/sparckles/robyn#%EF%B8%8F-to-develop-locally) on your PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "maturin"
 [project]
 name = "robyn"
 version = "0.82.1"
+requires-python = ">=3.10"
 description = "A Super Fast Async Python Web Framework with a Rust runtime."
 readme = "README.md"
 authors = [{ name = "Sanskar Jethi", email = "sansyrox@gmail.com" }]
--- a/test
+++ b/test
@@ -1,0 +1,1 @@
+test file, just to activate codspeed ci to run benchmark again.


thePromger:
@sansyrox 
i might need to update some dependencies, should i update them in this PR itself or create new one?
To update project/dev/test dependencies.

thePromger:
Did these change, because i saw in many big projects that they dont put upper bound limit for deps instead they just define lower bound limit supported, later i am able to understand the reason little bit.
If we dont pin the exact version, so pip can download the dep according to host's os+python requirements which have many advantages than pinning exact version.

pyproject.toml
```diff
@@ -24,17 +25,17 @@ classifiers = [
...
-  "orjson == 3.11.0",
-  "rustimport == 1.3.4",
+  "inquirerpy >= 0.3.4",
+  "multiprocess >= 0.70.14",
+  "orjson >= 3.11.0",
+  "rustimport >= 1.3.4",
   # conditional
-  "uvloop~=0.21.0; sys_platform != 'win32' and platform_python_implementation == 'CPython' and platform_machine != 'armv7l'",
-  "watchdog == 4.0.1",
+  "uvloop >= 0.21.0; sys_platform != 'win32' and platform_python_implementation == 'CPython' and platform_machine != 'armv7l'",
```
sansyrox, line 33: let's pin uvloop

pyproject.toml
```diff
@@ -24,17 +25,17 @@ classifiers = [
...
 ]
 dependencies = [
-  "inquirerpy == 0.3.4",
-  "multiprocess == 0.70.14",
-  "orjson == 3.11.0",
-  "rustimport == 1.3.4",
+  "inquirerpy >= 0.3.4",
+  "multiprocess >= 0.70.14",
+  "orjson >= 3.11.0",
+  "rustimport >= 1.3.4",
```
sansyrox, line 31: let's pin the latest working versions

pyproject.toml
thePromger: can i know the reason behing it?
bcoz if we dont pin so pip will automatically do install correct version according to host pc requirements

thePromger:
@sansyrox 
do read more about it, why codspeed performance dropped?
https://github.com/CodSpeedHQ/pytest-codspeed/releases


> ## [CodSpeed Performance Report](https://codspeed.io/sparckles/Robyn/branches/thePromger%3Achng-toml)
> ## Merging #1232 will **degrade performances by 25.45%**

what i am able to understand that this 25% performance drop is due to because we unpin the exact version, which allows pip to use latest version supported by host requirements
and from pytest-codspeed==3.0.0 something they introduced new way of measuring/benchmark method known as wall time
and from that version its set by default and to use that new method, we need to do some setup/changes like using their own runners and some other requirements too

let me test some senarious in robyn's CI inside this PR itself

thePromger:
@sansyrox 
locally both codspeed-mode={instrumentation,walltime} both are working fine today, but i don't know what is the issue in CI
note: yesterday, i think when used walltime (it took high cpu usage + every test takes very much longer time) in codespaces(i received the notification from them) and when defined mode as instrumentation it worked smoothly without any lags/delays.

sansyrox (commented):
Thanks for the contribution, @thePromger! As it stands this draft only adds the `requires-python = ">=3.10"` line (plus a stray root `test` file that we wouldn't want to merge), and the description mentions constraint and `uv.lock` changes that aren't actually in the diff. The `requires-python` addition is also already covered by the fuller Poetry→UV migration in #1297. Given it's conflicting against current main and superseded there, I'd recommend we close this in favor of #1297 — but thank you for flagging the `requires-python` gap, that was a real omission.

PR #1252, opened 2025-09-23
refactor: don't init config at class definition to allow custom config classes/argparser

### LINKED ISSUE #1246: [argparse] Don't instantiate Config on import Robyn
Hi, I noticed that argparse is deeply associated in Config():

https://github.com/sparckles/Robyn/blob/main/robyn/argument_parser.py#L6-L87

And that Config.__init__ is called many times, at least from:
```
caller name: WebSocket
caller name: <module>
caller name: BaseRobyn
caller name: SubRouter
```
at `import Robyn`, and global defined at

https://github.com/sparckles/Robyn/blob/main/robyn/__init__.py#L58

I would like to request some reconsideration/rework on allowing robyn apps without the default argparse setup, instead, let the user cli pass the appropriate parameters.

As is, my understanding is that it is rather difficult achieve a
```
my_custom_cli --help
```
that doesn't output robyn own help

Thanks,

gastmaier:
## Description

This PR fixes #1246

## Summary

Python's default arguments are evaluated once when the function is defined. Config was instantiated at BaseRobyn class definition and __init__.py body. Call Config() only inside BaseRobyn, and if not None. At SubRouter, forward the config parameter.

For config.compile_rust_path, access from the class object, and unset to ensure compile_rust_path is called once per config object. Allows to write custom argument parsers without colliding with Config.parser.

This allows to pass implement alternative configs, for example:

```python
from robyn import Robyn, Response as RobynResponse
from robyn.argument_parser import Config as RobynConfig

from .handler import my_route

class AltConfig(RobynConfig):
    def __init__(self) -> None:
        self.disable_openapi = True
        self.dev = False
        self.log_level = "DEBUG"
        self.open_browser = False
        self.fast = 1
        self.workers = 1
        self.processes = 1
        self.compile_rust_path = None

config=AltConfig()
app = Robyn(__file__, config=config)

@app.post("/my-route")
async def route_my_route(request):
    response =  my_route(app.env,  request)
    return RobynResponse(
        status_code=response.status_code,
        headers=response.headers,
        description=response.body,
    )

def main():
    app.start(port=app.env.port)
```
## PR Checklist

Please ensure that:

## Pre-Commit Instructions:

## High-level PR Summary
This PR refactors how the `Config` class is initialized in Robyn to fix a bug (#1246) related to Python's default argument evaluation behavior. Instead of initializing the config at class definition and in `__init__.py`, it now initializes the config inside the `BaseRobyn` class only when needed. This change enables users to implement custom config classes and argument parsers without colliding with the default `Config.parser`. The PR also ensures that `compile_rust_path` is called only once per config object by unsetting it after use.

⏱️ Estimated Review Time: 5-15 minutes

<details>
<summary>💡 Review Order Suggestion</summary>

| Order | File Path |
|-------|-----------|
| 1 | `robyn/__init__.py` |
</details>

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -55,28 +55,25 @@ def _normalize_endpoint(endpoint: str) -> str:
     return endpoint.rstrip("/")
 
 
-config = Config()
-
-if (compile_path := config.compile_rust_path) is not None:
-    compile_rust_files(compile_path)
-    print("Compiled rust files")
-
-
 class BaseRobyn(ABC):
     """This is the python wrapper for the Robyn binaries."""
 
     def __init__(
         self,
         file_object: str,
-        config: Config = Config(),
+        config: Optional[Config] = None,
         openapi_file_path: Optional[str] = None,
         openapi: Optional[OpenAPI] = None,
         dependencies: DependencyMap = DependencyMap(),
     ) -> None:
+        self.config = Config() if config is None else config
+        if (compile_path := self.config.compile_rust_path) is not None:
+            compile_rust_files(compile_path)
+            self.config.compile_rust_path = None
+
         directory_path = os.path.dirname(os.path.abspath(file_object))
         self.file_path = file_object
         self.directory_path = directory_path
-        self.config = config
         self.dependencies = dependencies
         self.openapi = openapi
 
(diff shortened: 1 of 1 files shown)


sansyrox:
@gastmaier , have you tested this in `--dev` mode?

gastmaier:
There shouldn't be any issues with dev mode, just resolved the multiple instantiation of Config() due to the behaviour of the Python interpreter.

PR #1260, opened 2025-10-28
fix: fixing routing, while files & api served from same base route

### LINKED ISSUE #1251: Only GET works together with serve_directory at root
### Bug Description

Hi,

I believe this is a common situation: I need my app to serve static files from the root, but I also have an API, and I can't make them work together.

### Steps to Reproduce

Example:

```python
from robyn import Robyn

app = Robyn(__name__)


@app.get("/foo")
async def get_foo():
    print("get received")
    return "get received"


@app.post("/foo")
async def post_foo():
    print("post received")
    return "post received"


@app.put("/foo")
async def put_foo():
    print("put received")
    return "put received"


@app.delete("/foo")
async def delete_foo():
    print("delete received")
    return "delete received"


@app.patch("/foo")
async def patch_foo():
    print("patch received")
    return "patch received"


@app.connect("/foo")
async def connect_foo():
    print("connect received")
    return "connect received"


@app.trace("/foo")
async def trace_foo():
    print("trace received")
    return "trace received"


@app.options("/foo")
async def options_foo():
    print("options received")
    return "options received"


app.serve_directory(
    route="/",
    directory_path="./frontend",
    index_file=None,
    show_files_listing=False,
)

app.start(host="0.0.0.0", port=8123)
```

When I try to GET, it works:

```bash
$ curl -X 'GET' http://127.0.0.1:8123/foo

get received
```

But when I try to POST, it fails:

```bash
$ curl -X 'POST' http://127.0.0.1:8123/foo

Request did not meet this resource's requirements.
```

### Your operating system

Linux

3.11

### Your Robyn version

latest

### Additional Info

_No response_

thePromger:
## Description
- When serving both static files and API routes from the same endpoint (e.g., `/`, `/app`, etc.), HTTP methods other than GET or HEAD are blocked by the Actix file service.  
- As a result, if a static file is not found for a request, the API route handler is never reached.

- This PR fixes #1251.

### Solution
- Added a `method_guard` to the Actix file service to allow requests with any HTTP method.  
- This ensures that if the file service cannot handle a request, it properly falls through to the API router.

### Impact
- All HTTP methods (GET, POST, PUT, DELETE, etc.) now correctly pass through to API routes when no matching static file exists.  
- Fixes routing issues when static file serving and API routing share the same base path.

## PR Checklist

Please ensure that:

## Pre-Commit Instructions:

## Summary by CodeRabbit

* **New Features**
  * Added a /static subrouter with an API endpoint at /static/build (GET and POST) that coexists with static file serving and is registered with the app.
* **Bug Fixes**
  * Static file handling now permissively accepts methods on directory routes but only serves actual files (directories/non-files are filtered).
* **Tests**
  * Added integration tests verifying the /static/build API and static file serving (e.g., /static/build/index.html).

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -1204,19 +1204,17 @@ def sse_status_code(request):
     return SSEResponse(event_generator(), status_code=201)
 
 
-# Serving static files at /static from ./integration_tests.
-app.serve_directory(
-    route="/static",
-    directory_path=str(current_file_path),
-)
-
-
 def main():
     app.set_response_header("server", "robyn")
     app.serve_directory(
         route="/test_dir",
         directory_path=os.path.join(current_file_path, "build"),
         index_file="index.html",
+    )
+    # Serving static files at /static from ./integration_tests.
+    app.serve_directory(
+        route="/static",
+        directory_path=str(current_file_path),
     )
     app.startup_handler(startup_handler)
     app.include_router(sub_router)


src/server.rs
```diff
@@ -155,18 +155,22 @@ impl Server {
                         if let Some(index_file) = &directory.index_file {
                             app = app.service(
                                 Files::new(&directory.route, &directory.directory_path)
+                                    .method_guard(guard::fn_guard(|_| true))
                                     .index_file(index_file)
                                     .redirect_to_slash_directory(),
                             );
```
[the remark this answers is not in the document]
thePromger: The method_guard here allows requests with any HTTP method, not just the default GET or HEAD, to be processed by the Actix file service.

By setting it to always return true, the file service now accepts all HTTP methods. If no matching static file is found, the request is passed on to the next registered service (such as the API router).

Before this fix, Actix would return early for non-default methods and never forward those requests to other services, which prevented the API routes from being reached.

integration_tests/base_routes.py
```diff
@@ -1204,6 +1204,13 @@ def event_generator():
     return SSEResponse(event_generator(), status_code=201)
 
 
+# Serving static files at /static from ./integration_tests.
+app.serve_directory(
+    route="/static",
+    directory_path=str(current_file_path),
+)
```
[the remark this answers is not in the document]
thePromger: I don't think it's a potential issue

integration_tests/base_routes.py
```diff
@@ -1204,6 +1204,13 @@ def event_generator():
...
+
+
```
sansyrox: why not add it in the `main()` function?

sansyrox:
@thePromger , is there an issue that this pr fixes specifically? in the issue list?

integration_tests/base_routes.py
thePromger: Done

integration_tests/base_routes.py
```diff
@@ -1211,9 +1211,15 @@ def main():
         directory_path=os.path.join(current_file_path, "build"),
         index_file="index.html",
     )
+    # Serving static files at /static from ./integration_tests.
+    app.serve_directory(
+        route="/static",
+        directory_path=str(current_file_path),
+    )
```
sansyrox: @thePromger , I don't think we should proceed with a special subrouter for a static file. 

What we should do is create a `serve_directory` function like `serve_file` and then remove the `app.serve_directory` function altogether.

sansyrox: or approach two

we can call the router as `directory_router`

thePromger: > @thePromger , I don't think we should proceed with a special subrouter for a static file. 


Actually static router is just to test new changes in the rust code when handling the request

That /static route is being served by both file handler and api handler too
If file handler can't serve it, then request will be sent to api handler to serve, else file handler just send the data of the file if found
Previously if file handler can't serve the request, it just return the request early
This is useful when serving file and api from common route prefix

> 
> What we should do is create a `serve_directory` function like `serve_file` and then remove the `app.serve_directory` function altogether.

I don't find any issue in current implementation, coz it's easy to serve the directory with sub router too, no need to require main app instance

PR #1328, opened 2026-03-12
fix: replace print() with logger in responses, reloader, and __init__

mango766:
## Summary

- Replace `print()` calls with proper `logger` usage in non-CLI modules
- Fix broken format strings in `reloader.py` that printed tuples instead of formatted messages

## Changes

| File | `print()` removed | Details |
|------|-------------------|---------|
| `responses.py` | 1 | Async generator error now uses `logger.error()` |
| `reloader.py` | 6 | Rust compilation logging uses `logger.info/error` |
| `__init__.py` | 1 | Rust compilation message uses `logger.info()` |

## Bug Fix: Broken Format Strings

`reloader.py` had several broken format strings using `print("msg: %s", arg)` which doesn't do string interpolation — it just prints the tuple `("msg: %s", arg)`. These are now fixed with proper `logger` calls that correctly format the arguments.

**Before:**
```python
print("Created rust file : %s", rust_file)
# Output: ('Created rust file : %s', 'foo.rs')
```

**After:**
```python
logger.info("Created rust file: %s", rust_file)
# Output: Created rust file: foo.rs
```

## Note

CLI-facing output in `cli.py` is intentionally left as `print()` since those are user-facing messages.

## Test Plan

- [ ] Verify Python syntax — passed
- [ ] `robyn --dev` correctly shows reloader logs
- [ ] Streaming responses correctly log errors

## Summary by CodeRabbit

* **Refactor**
  * Standardized error reporting by replacing direct console prints with structured logging across core runtime components, improving diagnostic output and consistency.
  * Added a default no-op log handler to prevent spurious "no handlers" warnings when no logging is configured, reducing noisy console output.

### DIFF
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -70,7 +70,7 @@ config = Config()
 
 if (compile_path := config.compile_rust_path) is not None:
     compile_rust_files(compile_path)
-    print("Compiled rust files")
+    logger.info("Compiled rust files")
 
 
 class BaseRobyn(ABC):
--- a/robyn/__init__.py
+++ b/robyn/__init__.py
@@ -70,7 +70,7 @@ config = Config()
 
 if (compile_path := config.compile_rust_path) is not None:
     compile_rust_files(compile_path)
-    logger.info("Compiled rust files")
+    print("Compiled rust files")
 
 
 class BaseRobyn(ABC):
--- a/robyn/logger.py
+++ b/robyn/logger.py
@@ -19,6 +19,7 @@ class Logger:
 
     def __init__(self):
         self.logger = logging.getLogger(__name__)
+        self.logger.addHandler(logging.NullHandler())
 
     def _format_msg(
         self,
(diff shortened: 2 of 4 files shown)


themavik (commented):
Switching `reloader.py` from `print` tuples to `logger.info/error` with `%s` args fixes the broken formatting; `responses.py` async-gen errors now hit logging instead of stdout.

sansyrox (commented):
Nice cleanup, @mango766 — and thanks especially for catching the broken format strings; calls like `print("Created rust file : %s", rust_file)` were printing the args as a tuple rather than interpolating, so moving them to proper `logger` calls is a real fix. The `NullHandler` addition is the correct library idiom and is safe here since `basicConfig` is still called in `BaseRobyn.__init__` and the default log level is `INFO`, so these messages still surface. Two small things: the PR description lists a change to `__init__.py`, but the diff actually edits `logger.py` and leaves `print("Compiled rust files")` at `__init__.py:74` in place — could you either drop that from the description or finish converting that line for consistency? Also note this overlaps with #1396 in `reloader.py`, so we'll coordinate the merge order. Once the description matches the diff, this is good to go.

PR #1382, opened 2026-04-18
fix: share contextvars Context across before/handler/after hooks (#1380)

### LINKED ISSUE #1380: contextvars.ContextVar not shared between before_request and after_request hooks
## Description

`contextvars.ContextVar` values set in a `@app.before_request()` hook are not visible in `@app.after_request()` hooks. The contextvar reads back as the default value in `after_request`, indicating the two hooks run in different contextvars contexts.

## Reproduction

```python
import contextvars
from robyn import Robyn, Request, Response

app = Robyn(__file__)
_request_ctx: contextvars.ContextVar[str] = contextvars.ContextVar("_ctx", default="")

@app.before_request()
async def before(request: Request) -> Request:
    _request_ctx.set("hello")
    print(f"before_request: {_request_ctx.get()}")  # prints "hello"
    return request

@app.after_request()
async def after(response: Response) -> Response:
    print(f"after_request: {_request_ctx.get()}")  # prints "" (default, NOT "hello")
    return response

@app.get("/test")
async def test(request: Request):
    return "ok"

app.start(host="0.0.0.0", port=8000)
```

## Expected behavior

`_request_ctx.get()` in `after_request` should return `"hello"` — the value set in `before_request` for the same request.

## Actual behavior

`_request_ctx.get()` in `after_request` returns `""` (the default). The contextvars context is not propagated between the two hooks.

## Impact

This breaks any pattern that needs to pass per-request state from `before_request` to `after_request`, such as CORS origin matching. Workaround: use `threading.local()` instead of `contextvars.ContextVar`.

## Environment

- Robyn 0.82.1
- Python 3.13
- macOS

sansyrox:
## Summary
- Each middleware phase was awaited through its own `asyncio.Task`; asyncio copies the current context for every new task, so `ContextVar` writes in `before_request` were isolated to that task and invisible to the route handler and `after_request`.
- Allocate one `contextvars.Context()` per request in `index()` and thread it through the executor calls. Async handlers are scheduled via a cached Python helper that calls `loop.create_task(coro, context=ctx)`, so asyncio uses the shared context object directly (no copy) and writes persist across phases. Sync handlers run via `ctx.run(handler, ...)`.
- Fixes #1380, so users no longer need the `threading.local()` workaround.

## Test plan
- [x] Added async + sync regression tests (`/async/contextvars/route`, `/sync/contextvars/route`) in `integration_tests/test_middlewares.py` that write a `ContextVar` in `before_request` and assert the handler and `after_request` both observe the value.
- [x] Full local run: `PYTHONPATH=. pytest integration_tests/ -q --ignore=integration_tests/test_web_sockets.py --ignore=integration_tests/test_openapi.py` — 314 passed.
- [x] Standalone reproducer from the issue now prints the set value in all three phases.

## Summary by CodeRabbit

* **New Features**
  * Context variables now propagate consistently across the full request lifecycle (before hooks, handlers, after hooks) for both async and sync requests, ensuring middleware and handlers observe the same contextual state.

* **Tests**
  * Added integration tests to verify context propagation for async and sync flows; the async variant is conditionally skipped on older Python versions.

### DIFF
--- a/integration_tests/base_routes.py
+++ b/integration_tests/base_routes.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextvars
 import json
 import os
 import pathlib
@@ -267,6 +268,45 @@ def sync_before_request_401():
 @app.get("/sync/middlewares/401")
 def sync_middlewares_401():
     pass
+
+
+# --- ContextVar propagation (regression test for #1380) ---
+
+_ctxvar_middleware_value: contextvars.ContextVar = contextvars.ContextVar("ctxvar_middleware_value", default="default")
+
+
+@app.before_request("/async/contextvars/route")
+async def contextvars_before(request: Request):
+    _ctxvar_middleware_value.set("set-in-before")
+    return request
+
+
+@app.after_request("/async/contextvars/route")
+async def contextvars_after(response: Response):
+    response.headers.set("x-ctxvar-after", _ctxvar_middleware_value.get())
+    return response
+
+
+@app.get("/async/contextvars/route")
+async def contextvars_handler(request: Request):
+    return _ctxvar_middleware_value.get()
+
+
+@app.before_request("/sync/contextvars/route")
+def contextvars_before_sync(request: Request):
+    _ctxvar_middleware_value.set("set-in-sync-before")
+    return request
+
+
+@app.after_request("/sync/contextvars/route")
+def contextvars_after_sync(response: Response):
+    response.headers.set("x-ctxvar-after", _ctxvar_middleware_value.get())
+    return response
+
+
+@app.get("/sync/contextvars/route")
+def contextvars_handler_sync(request: Request):
+    return _ctxvar_middleware_value.get()
 
 
 # ===== Routes =====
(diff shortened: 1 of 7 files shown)


sansyrox:
Addressed the review in f9557b1:

- **Python 3.10 compatibility (CodeRabbit)**: `loop.create_task(coro, context=ctx)` needs 3.11+. The helper now gates on `sys.version_info`; on 3.10 it falls back to plain `await coro` (pre-fix behavior — no crash, but `ContextVar` writes still won't propagate across async hooks on 3.10). Sync middleware continues to share context across hooks on all versions via `ctx.run(handler, ...)`. The async regression test is skipped on < 3.11.
- **Rustfmt**: `cargo fmt` on my touched files.
- **CodSpeed benchmark failure**: unrelated — CodSpeed's CDN returned 504 fetching the runner installer. Should pass on rerun.

Verified locally:
- Python 3.11: `ContextVar` writes propagate across hooks.
- Python 3.10: helper loads and runs without error (fallback path); async cross-hook test is skipped.

Issue #1394, opened 2026-05-29
Inconsistent prefix handling between HTTP routes and WebSocket routes in nested SubRouters


shyixiu:
### Description

When using nested SubRouters (e.g., SubRouter C included in SubRouter B, which is then included in the main App), HTTP routes and WebSocket routes behave differently regarding prefix accumulation:

- **HTTP routes**: Only get the prefix of the SubRouter they are directly defined in
- **WebSocket routes**: Accumulate prefixes from all parent routers in the nesting chain

This inconsistency can lead to confusion when designing API endpoints.

### Minimal Reproduction

```python
from robyn import Robyn, SubRouter

# Create main app
app = Robyn(__file__)

# Create middle router (b) with prefix /bbb
router_b = SubRouter(__file__, prefix="/bbb")

# Create nested router (c) with prefix /ccc
router_c = SubRouter(__file__, prefix="/ccc")

# Define HTTP route in router_c
@router_c.get("/hello")
def hello_from_c():
    return {"message": "Hello from router_c"}

# Define WebSocket route in router_c
@router_c.websocket("/ws")
async def ws_from_c(websocket):
    await websocket.accept()
    await websocket.send_text("Hello from WebSocket in router_c")
    await websocket.close()

# Include router_c into router_b
router_b.include_router(router_c)

# Include router_b into main app
app.include_router(router_b)

# Print registered routes
print("HTTP Routes:")
for route in app.router.routes:
    print(f"  {route.route_type} {route.route}")

print("\nWebSocket Routes:")
for endpoint, handlers in app.web_socket_router.routes.items():
    print(f"  WS {endpoint}")
```

### Actual Output

```
HTTP Routes:
  HttpMethod.GET /ccc/hello

WebSocket Routes:
  WS /bbb/ccc/ws
```

### Expected Behavior

Both HTTP routes and WebSocket routes should behave consistently. There are two possible approaches:

**Option 1: Both accumulate prefixes (like WebSocket currently does)**
```
HTTP Routes:
  HttpMethod.GET /bbb/ccc/hello

WebSocket Routes:
  WS /bbb/ccc/ws
```

**Option 2: Both use only the direct SubRouter's prefix (like HTTP currently does)**
```
HTTP Routes:
  HttpMethod.GET /ccc/hello

WebSocket Routes:
  WS /ccc/ws
```

### Root Cause Analysis

The inconsistency stems from **when and where** the prefix is applied:

[...]

PR #1398, opened 2026-06-07
Normalised routing for HTTP and Web Sockets (fix for error#1394)

oscarpeer-design:
## Description

This PR fixes inconsistent prefix handling between HTTP routes and WebSocket routes in nested SubRouter usage (see issue #1394).
This change only affects nested SubRouter inclusion and does not modify single-router prefix behaviour.
## Summary

Previously, HTTP and WebSocket routes behaved inconsistently when routers were nested:
- HTTP routes did not fully accumulate parent router prefixes
- WebSocket routes already accumulated full prefix chains

This PR updates include_router() so that HTTP routes now behave consistently with WebSocket routes, ensuring that nested router prefixes are applied uniformly.

## PR Checklist

Please ensure that:

- [X] You have added relevant tests. We prefer integration tests wherever possible

## Pre-Commit Instructions:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected subrouter inclusion so HTTP and WebSocket routes register with the proper combined prefixes and OpenAPI paths merge in the correct order.

* **Tests**
  * Added a test ensuring nested subrouter prefixes produce the expected HTTP and WebSocket route paths and that OpenAPI entries are incorporated correctly.

### DIFF
--- a/unit_tests/test_subrouter_nested_prefixes.py
+++ b/unit_tests/test_subrouter_nested_prefixes.py
@@ -1,0 +1,43 @@
+from robyn import Robyn, SubRouter
+
+#Tests nested subrouter prefixes are consistent with websocket routes
+def test_subrouter_nested_prefixes():
+    # Create main app
+    app = Robyn(__file__)
+
+    # Create middle router (b) with prefix /bbb
+    router_b = SubRouter(__file__, prefix="/bbb")
+
+    # Create nested router (c) with prefix /ccc
+    router_c = SubRouter(__file__, prefix="/ccc")
+
+    # Define HTTP route in router_c
+    @router_c.get("/hello")
+    def hello_from_c():
+        return {"message": "Hello from router_c"}
+
+    # Define WebSocket route in router_c
+    @router_c.websocket("/ws")
+    async def ws_from_c(websocket):
+        await websocket.accept()
+        await websocket.send_text("Hello from WebSocket in router_c")
+        await websocket.close()
+
+    # Include router_c into router_b
+    router_b.include_router(router_c)
+
+    # Include router_b into main app
+    app.include_router(router_b)
+
+    # List registered  HTTP routes
+    http_routes = []
+    for route in app.router.routes:
+        http_routes.append(route.route)
+
+    # List registered Websocket routes
+    websocket_routes = list(app.web_socket_router.routes.keys())
+
+    # HTTP and WebSocket routes should both be correctly prefixed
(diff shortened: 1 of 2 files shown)


unit_tests/test_subrouter_nested_prefixes.py
```diff
@@ -0,0 +1,51 @@
+from robyn import Robyn, SubRouter
+<<<<<<< HEAD
+=======
+
+
+# Tests nested subrouter prefixes are consistent with websocket routes
+>>>>>>> ef04402802eb361de3500bbf29e432f3af461f3e
```
sansyrox: @oscarpeer-design , this contains a merge conflict diff

oscarpeer-design: Any merging issues are because of whitespace changes in the test file I added.

oscarpeer-design: I resolved the CI issues.

sansyrox (commented):
Thanks for digging into this, @oscarpeer-design! There's a genuinely good fix buried in here: switching `add_subrouter_paths(self.openapi)` to `add_subrouter_paths(router.openapi)` corrects a real no-op bug where subrouter OpenAPI paths were never merged into the parent spec. However, the PR doesn't actually resolve #1394: after the change, HTTP routes still register as `/ccc/hello` while WebSocket routes register as `/bbb/ccc/ws`, so the inconsistency remains — and the new test asserts exactly that mismatched state rather than a consistent one. Per the issue's preferred resolution (Option 1), included HTTP routes should accumulate the parent prefix too, so the nested route should become `/bbb/ccc/hello`; that requires prepending the parent prefix to each included `Route` in `include_router`, not just reordering the `extend`. Could you update the logic so HTTP and WS truly match, change the test to assert `/bbb/ccc/hello`, and add an assertion on `openapi_spec["paths"]` to lock in the OpenAPI fix? If you'd prefer, splitting the OpenAPI fix into its own PR is also welcome since that part is ready to go.

sansyrox:
Thanks for this, @oscarpeer-design! 🙏 Your fix for #1394 — making `include_router()` accumulate prefixes for HTTP routes the way WebSocket routes already did — has been folded into the consolidated SubRouter PR #1404 (merged as 31e8a7df), which fixes the same nested-prefix bug (plus `add_route`, a modern prefix-first syntax, and cleanups) with unit + integration tests. Closing in favor of the consolidated PR; really appreciate the trace and the contribution!

Issue #1165, opened 2025-05-16
Missing Features


dcd894:
- Input Filters
- Input Validations & sanitizations
- Data Serialization
- Session Mangement & Authentication

k4976:
@dcd894 can you share any example of package or any api, like how you want it to be implement?

dcd894:
@k4976 Of course! Php has input filters & validations & sanitizations, Php also has Session, finally flask restful has data serialization (data from database to python to json, and vice versa). But the flask restful for data serialization should be more improved as it is lacking.

k4976:
> @k4976 Of course! Php has input filters & validations & sanitizations, Php also has Session, finally flask restful has data serialization (data from database to python to json, and vice versa). But the flask restful for data serialization should be more improved as it is lacking.

I thought you are talking about pydantic or something else

Can you share some sample codes & examples output

sansyrox:
Closing as substantially addressed. Three of the four items are now first-party: input validation (robyn/pydantic_support.py), data serialization (jsonify + pydantic response serialization), and session management & authentication (robyn/session.py + robyn/authentication.py). The remaining 'input filters / sanitization' item is too underspecified to action — please open a focused issue if you still need it.

sansyrox:
Reopening — closed prematurely. Authentication/sessions and pydantic validation do exist now, but the Input Filters / sanitization item on this checklist is not done, so this tracking issue shouldn't have been closed as complete.

PR #1411, opened 2026-06-20
feat: first-party signed-cookie sessions via request.session (#945)

### LINKED ISSUE #945: Add sessions in Robyn
https://github.com/sparckles/Robyn/issues/943#issuecomment-2308721629

sansyrox:
## What

Adds **HTTP sessions** to Robyn — the one item from the "Missing Features" wishlist (#1165) that genuinely did not exist. (Input validation, data serialization, and authentication already ship via the Pydantic integration, orjson, and `AuthenticationHandler`.)

The session is a dict-like object stored entirely client-side in a tamper-proof **signed cookie** (HMAC-SHA256 over a base64url JSON payload), so there is no server-side store to run. It is accessed via **`request.session`**, mirroring Robyn's existing `request.identity`:

```python
app.configure_sessions(secret_key="keep-me-secret")

@app.get("/visits")
def visits(request):
    request.session["count"] = request.session.get("count", 0) + 1
    return f"You have visited {request.session['count']} times"
```

## How it works

- **`src/types/request.rs`** — adds a `session` field to `Request`/`PyRequest`, exactly mirroring `identity`. It carries the Python `Session` object **by reference** across `before_request → handler → after_request`, so in-handler mutations are visible when the cookie is written back. This works on **every supported Python version** (no `contextvars` dependency).
- **`robyn/session.py`** — `Session` (a `MutableMapping` that tracks whether it was modified), a stdlib HMAC signer, and `SessionManager` (load/save + serialize/verify with expiry).
- **`Robyn.configure_sessions()`** registers a global `before_request` that loads + verifies the cookie into `request.session`, and a global `after_request` that signs and writes it back **only when modified** (read-only requests send no `Set-Cookie`).

## Behaviour / safety

- Cookie is `HttpOnly` + `SameSite=Lax` by default; all attributes (`secure`, `max_age`, `path`, `domain`, `same_site`, `cookie_name`) are configurable.
- Clearing the session expires the cookie in the browser.
- Tampered, expired, or foreign-key-signed cookies decode to an **empty** session (never an error, never trusted).
- The payload is signed, **not encrypted** — documented clearly; don't store secrets in it.

## Tests

- **30 unit tests** (`unit_tests/test_sessions.py`): signing round-trip/tamper/wrong-key, `Session` dict semantics + modified-tracking, serialization, expiry, cookie read/write, `same_site` normalization, and a real `request.session` round-trip.
- **6 integration tests** (`integration_tests/test_sessions.py`): cross-request persistence, per-client isolation, `HttpOnly`, tamper rejection, clear/logout.

`ruff check`, `ruff format`, standalone `isort`, and `cargo fmt --check` all clean. Regression-swept auth/identity, middlewares, cookies, multipart, and direct `Request(...)` construction (65 tests) — green.

## Docs

New **Sessions** page under API Reference plus a nav entry.

Closes #945

## Release Notes

* **New Features**
  * Added signed-cookie session support via `configure_sessions()`.
  * Exposed sessions to request handlers via `request.session`, persisting across requests when modified.
  * Session cookies are created with secure defaults (including `HttpOnly`) and configurable cookie settings.

* **Documentation**
  * Added a new “Sessions” API reference page and updated documentation navigation.

* **Tests**
  * Added unit and integration tests covering persistence, per-client behavior, cookie flags, tamper detection, expiration, and clearing sessions.

### DIFF
--- a/docs_src/src/components/documentation/Navigation.jsx
+++ b/docs_src/src/components/documentation/Navigation.jsx
@@ -253,6 +253,10 @@ export const navigation = [
         title: 'Authentication',
       },
       {
+        href: '/documentation/en/api_reference/sessions',
+        title: 'Sessions',
+      },
+      {
         href: '/documentation/en/api_reference/const_requests',
         title: 'Const Requests and Multi Core Scaling',
       },
--- a/docs_src/src/pages/documentation/en/api_reference/sessions.mdx
+++ b/docs_src/src/pages/documentation/en/api_reference/sessions.mdx
@@ -1,0 +1,85 @@
+export const description =
+  'On this page, you’ll learn how to store per-user state across requests using Robyn’s signed-cookie sessions.'
+
+Once the Gotham Police Department could authenticate, Batman wanted each officer's dashboard to remember small bits of state between requests — the theme they picked, a CSRF token, the last case they viewed — without standing up a database for it. So he reached for Robyn's sessions.
+
+## Sessions
+
+Robyn provides signed-cookie sessions. The session is a small dictionary that is stored on the client inside a tamper-proof cookie (an HMAC-SHA256 signature over the data). No server-side store is required, and the signature means a client cannot modify the contents without invalidating it.
+
+<b>
+  Note: the session is signed, not encrypted. The data is encoded, so the client can read it — never store secrets (passwords, raw tokens) in the session.
+</b>
+
+Enable sessions once on your app with `configure_sessions`, then read or write the session inside any handler through `request.session` — exactly like `request.identity` for authentication.
+
+<Row>
+<Col>
+Call `configure_sessions` with a `secret_key`. Keep this key secret and stable across restarts and across all your worker processes — it is what signs and verifies every session cookie.
+</Col>
+  <Col sticky>
+    <CodeGroup title="Setup">
+
+    ```python
+    from robyn import Robyn
+
+    app = Robyn(__file__)
+
+    app.configure_sessions(secret_key="keep-me-secret")
+    ```
+    </CodeGroup>
+  </Col>
+</Row>
+
+<Row>
+<Col>
+Inside a handler, `request.session` is a dictionary-like `Session` object. Read it like a `dict`, and mutate it like a `dict`. Robyn writes the updated session back to the response cookie automatically — but only when the session was actually modified, so read-only requests do not send a `Set-Cookie`.
+</Col>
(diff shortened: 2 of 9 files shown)


sansyrox:
Addressed the review feedback in cd75b1c:

- **Double registration** — `configure_sessions()` now raises if called twice instead of silently registering duplicate middleware.
- **Nested mutations silently lost** — `Session.is_modified()` now also diffs the current contents against a snapshot taken at load, so `session["items"].append(x)` is persisted even though it bypasses `__setitem__`. `save()` uses `is_modified()`.
- **SameSite=None + insecure** — `SessionManager` now raises when `same_site="None"` without `secure=True` (browsers drop those cookies).
- **Non-numeric `exp`** — `loads()` guards that `exp` is numeric before comparing, so a malformed payload can't raise during request handling.
- **Test timeouts** — every `requests` call in the integration tests now passes `timeout=`.

Added 7 unit tests covering these. Full session suite (36 unit + 6 integration) green; ruff/isort/cargo fmt clean.

sansyrox:
Thanks for the review — all comments are addressed (in `cd75b1c7b` and the new `c983d4c4b`):

- **Double `configure_sessions()` registration** — guarded; a second call now raises `RuntimeError` instead of registering duplicate session middleware (`robyn/__init__.py`).
- **Nested mutations silently lost** — `save()` no longer relies on the `modified` flag alone. `is_modified()` compares a canonical, order-independent JSON snapshot taken at load time against the current contents, so in-place edits like `session["items"].append(...)` are detected and persisted (`robyn/session.py`).
- **`SameSite=None` with an insecure cookie** — `SessionManager` now raises `ValueError` when `same_site="None"` without `secure=True`, failing loudly at configuration time rather than letting the browser silently drop the cookie.
- **`exp` type guard** — a non-numeric or `bool` `exp` is treated as an invalid (empty) session rather than raising during request handling.
- **Non-finite `exp` (NaN / Infinity)** — `json.loads` accepts these as floats and a plain `exp < now` comparison never expires them, so they slipped past the type guard. Now rejected via `math.isfinite`, with `NaN` / `Infinity` / `-Infinity` unit tests (`c983d4c4b`).
- **Missing request timeouts in tests** — every `requests` call in `integration_tests/test_sessions.py` now passes `timeout=TIMEOUT` (5s) so a hung server can't stall CI.

All 39 session unit tests pass locally; `ruff`, `ruff format`, and `isort` are clean.

PR #1412, opened 2026-06-21
fix(streaming): support bytes chunks + fix async DB access in generators (#1236, #1219)

### LINKED ISSUE #1219: RuntimeError occurred  when execute an asynchronous database access using sqlalchemy
### Bug Description

 async with AsyncSessionLocal() as session:
                stmt = select(ChatMessage).where(
                    ChatMessage.session_id == session_id,
                    ChatMessage.is_deleted == False
                ).order_by(desc(ChatMessage.created_at)).limit(limit)

                **result = await session.execute(stmt)**
                messages = result.scalars().all().     when run into result = await session.execute(stmt), a runtime error: RuntimeError("Task <Task pending name='Task-3' coro=<AsyncGeneratorWrapper._get_next_value.<locals>.get_next() running at. occurred.  the request returns a streaming response:  return StreamingResponse(
                generate(),
                media_type="text/event-stream",
                headers=Headers({
                    "Cache-Control": "no-cache",
                    "Connection": "keep-alive",
                    "X-Accel-Buffering": "no" 
                })
            ).   please look into this issue and thanks in advance if there's any solutions to this issue.

### Steps to Reproduce

_No response_

### Your operating system

None

None

### Your Robyn version

None

### Additional Info

_No response_

### LINKED ISSUE #1236: About using StreamingResponse
Hi!
I'm having trouble getting StreamingResponse to work. I need to stream large binary files (5Go to 20go), and I come from a flask framework, where I could do this : 
```
def generator():
    with my_file.open("rb") as f:
        yield from f
r = app.response_class(generator(), content_type="application/octet-stream", direct_passthrough=True)
r.headers.set("Content-Disposition", "attachment", download_name=filename, filename=filename)
r.headers.add("Content-Length", str(my_file.stat().st_size))
return r
```
Now that we are able to use yield with Robyn (thanks @sansyrox ), I tried this : 
```
def generate_file():
    with my_file.open("rb") as f:
        yield from f
return StreamingResponse(content=generate_file(), media_type="application/octet-stream")
```
Which doesn't work because the generator handler in Rust is trying (I think) to cast the value as string [here](https://github.com/sparckles/Robyn/blob/d29608d404fd7b9d8a05ecde97eef2ab5047bbf9/src/types/response.rs#L130) instead of directly streaming bytes from the file : `robyn.types.response| Failed to extract string from generator value: TypeError: 'bytes' object cannot be converted to 'PyString'`

However, the following works but is quite slow (about <2mo/sec when I could get  ~5 times this before with Flask), and I'm getting bytes -> str overhead (>100ms with my 10MB chunks).  
```
while True:
  chunk = f.read(10*1024*1024)
  yield str(chunk)
```
I'm using Robyn 0.72.0 in "fast" mode. Am I missing something ?

sansyrox:
Fixes two related `StreamingResponse` bugs that came out of user reports.

## #1236 — yielding `bytes` crashed

Streaming binary data (`yield b"..."`) raised `'bytes' object cannot be converted to 'PyString'` because the Rust stream driver only extracted `str`. It now downcasts `PyBytes` first (used as-is) and falls back to UTF-8-encoding `str`, so file downloads / `application/octet-stream` work. A value that's neither `str` nor `bytes` is logged and ends the stream instead of being silently dropped.

```python
@app.get("/download")
def download(request):
    def gen():
        with open(path, "rb") as f:
            while chunk := f.read(8192):
                yield chunk
    return StreamingResponse(gen(), media_type="application/octet-stream")
```

## #1219 — awaiting async DB inside a generator crashed

`await session.execute(...)` (async SQLAlchemy) inside a streaming generator raised `Task ... attached to a different loop`. `AsyncGeneratorWrapper` drove the generator on a **freshly-created** event loop and **swallowed every error** (`print` + `raise StopIteration`), so the stream silently truncated.

It now drives the generator on the loop that was running when the `StreamingResponse` was constructed — i.e. the handler's loop, where the async resources are bound — via `run_coroutine_threadsafe`, falling back to a dedicated background loop only for sync handlers. Generator errors now propagate (and show up in the server logs) instead of disappearing.

## Tests

- **4 unit tests** (`unit_tests/test_streaming_response.py`): generator runs on the constructing loop, errors propagate, background-loop fallback for sync handlers, bytes pass-through.
- **2 integration tests** (`integration_tests/test_binary_streaming.py`): sync and async `bytes` streaming.
- Existing SSE suite still green (the wrapper rewrite doesn't regress async SSE).

`ruff` / `ruff format` / `isort` / `cargo fmt --check` all clean.

Supersedes #1308 (which fixed only the bytes case) — happy to close that in favour of this.

Closes #1236
Closes #1219

## Summary by CodeRabbit

* **New Features**
  * Streaming responses now support yielding raw `bytes` (including async streaming) for binary downloads.
  * Added binary streaming endpoints for both sync and async byte streams.
* **Documentation**
  * Updated Server-Sent Events docs with examples for streaming arbitrary raw bytes via `StreamingResponse`.
* **Bug Fixes**
  * Async streaming now propagates exceptions to the consumer instead of silently terminating.
* **Tests**
  * Added integration tests for sync/async byte streaming and unit tests covering async wrapper behavior and cleanup.

### DIFF
--- a/docs_src/src/pages/documentation/en/api_reference/server_sent_events.mdx
+++ b/docs_src/src/pages/documentation/en/api_reference/server_sent_events.mdx
@@ -189,6 +189,68 @@ async def stream_api_events(request):
 
 ---
 
+## Streaming raw bytes (binary data & file downloads)
+
+Not everything Batman streams is an SSE event. To stream **arbitrary bytes** — a large file, a generated archive, an `application/octet-stream` body — he uses `StreamingResponse` directly and yields chunks, without loading the whole payload into memory. Each chunk may be `bytes` (sent as-is) or `str` (UTF-8 encoded).
+
+<Row>
+<Col>
+A sync generator yielding `bytes` chunks, served as a binary download:
+</Col>
+<Col sticky>
+<CodeGroup title="Request" tag="GET" label="/download">
+
+```python
+from robyn import Robyn, StreamingResponse, Headers
+
+app = Robyn(__file__)
+
+@app.get("/download")
+def download(request):
+    def file_chunks():
+        with open("large_file.bin", "rb") as f:
+            while chunk := f.read(8192):
+                yield chunk  # each chunk is `bytes`
+
+    return StreamingResponse(
+        file_chunks(),
+        media_type="application/octet-stream",
+        headers=Headers({
+            "Content-Type": "application/octet-stream",
+            "Content-Disposition": "attachment; filename=large_file.bin",
+        }),
+    )
+```
+
+</CodeGroup>
+</Col>
+</Row>
+
+<Row>
+<Col>
+`StreamingResponse` accepts both sync and async generators. When you use an **async generator**, it runs on the same event loop as your handler, so you can safely `await` async resources — an async database session, an HTTP client — **inside** the generator:
+</Col>
+<Col sticky>
+<CodeGroup title="Request" tag="GET" label="/export">
+
+```python
+@app.get("/export")
+async def export(request):
+    async def rows():
+        async with AsyncSessionLocal() as session:
+            result = await session.execute(select(Record))  # await works here
+            for row in result.scalars():
+                yield f"{row.id},{row.name}\n".encode()
+
+    return StreamingResponse(rows(), media_type="text/csv")
+```
+
+</CodeGroup>
+</Col>
+</Row>
+
+---
+
 ## What's next?
 
 Batman has mastered Server-Sent Events and can now stream real-time updates to his crime dashboard. While SSE is perfect for one-way communication from server to client, Batman realizes he needs bidirectional communication for more interactive features like real-time chat with his allies.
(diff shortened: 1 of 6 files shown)


sansyrox:
Good catch on the background-loop thread leak — fixed in bd0cf8493.

- The thread target and cleanup are now staticmethods that take the loop as an argument rather than bound methods, so they don't keep the wrapper alive (a bound-method `target=self._run_loop` would have created a self-reference that prevents GC, which would have defeated any GC-based cleanup).
- A `weakref.finalize` stops the loop when the wrapper is dropped, so the loop closes and the daemon thread exits even if the stream ends early (client disconnect, or an unsupported chunk type). `_finish()` still calls it for the normal path.

Added a unit test that abandons the stream after one chunk and asserts the background thread is joined.

Also verified the fix end-to-end against a live server: streaming a generator that `await`s a future bound to the handler's loop returns correctly on this branch, and returns an empty body on `main` (the original #1219 failure).

PR #1415, opened 2026-06-21
fix: batch of quick-win issues (#1099, #1147, #1273, #495, #470, #1192)

### LINKED ISSUE #470: Find a way to test startup and shutdown events.
Right now, the tests present to test the startup and the shutdown events are basically unit tests.

We need to find a way to verify if they've been called or not.

### LINKED ISSUE #495: Write a test to check the uploading and saving of a file
Write a test to check the uploading and saving of a file.

### LINKED ISSUE #1099: Add `to_dict` method for request headers

### LINKED ISSUE #1147: Sample app uses synchronous DB access in async route handlers
Just dropping in since I saw this shared on hackernews a little while ago and spotted something I've seen other web frameworks do in their example code in the past that I wondered about.

In the example app here https://robyn.tech/documentation/en/example_app/modeling_routes#crime-data-model-and-database-connection, there is some example code that looks like this:
```python
@app.post("/crimes")
async def add_crime(request):
    with SessionLocal() as db:
        crime = request.json()
        insertion = crud.create_crime(db, crime)  # this is the potentially problematic line

    if insertion is None:
        raise Exception("Crime not added")

    return {
        "description": "Crime added successfully",
        "status_code": 200,
    }
```
where the crud implementation is obviously using synchronous SQLAlchemy code. SQLAlchemy supports asyncio, and that's for a good reason: code like this will unnecessarily block the entire event loop while the database is being accessed, asynchronous DB access would be much better here. Technically, this isn't a bug, but unless I'm missing some background magic here, it's definitely not advisable and would probably kill performance for even simple CRUD apps. 

To compare: in FastAPI, this would be [worked around](https://fastapi.tiangolo.com/async/#path-operation-functions) by simply *not* declaring the route handers async, as FastAPI will just run the handler in a background threadpool in that case, so my question is: does robyn have some equivalent magic that runs every *async* handler in a threadpool as well?

Feel free to close if this is intended code and robyn has some hidden mechanism for handling this gracefully, but as someone just checking out the framework it struck me as a potential problem.

### LINKED ISSUE #1192: Discussion: Refactoring of Request object
Here, we gonna discuss about the how to do the refactor of request object to add few common methods like

```
# these are the common methods name in fastapi and flask for the sub-objects of request object like for headers, 
params, query_params, url_params, cookies, etc.

.get('key', default)
.to_dict()
.keys()
.items()
.values()
.__contains__() (for in operator)
.getlist()
.multi_items()
``` 

Need to discuss, which one to go with, and try to make it as much as fastapi/flask compatible + even make it lighter/efficient/less-overheaded as much as possible.

Let's discuss with our top chefs - @sansyrox @VishnuSanal

### LINKED ISSUE #1273: How to log requests in Robyn?
I want to log requests like this:
Image

I can’t get request/access logs to show in Robyn.

- Tried setting log level to DEBUG / INFO

- Followed the documentation: https://robyn.tech/documentation/en/api_reference/scaling#monitoring-and-observability (not working)

Some documented examples don’t seem to work. Is the project still actively maintained?

sansyrox:
Knocks out a batch of small, verified issues.

## Code
- **#1099** — `Headers.to_dict()` returns a flat `{name: value}` dict (duplicate values joined with `", "` per RFC 7230) so callers get normal dict ergonomics (`.get(key, default)`).
- **#1192** — dict-like accessors `keys()` / `values()` / `items()` / `multi_items()` on **QueryParams** and **Headers**. `keys/values/items` use last-value semantics (consistent with `get()`); `multi_items()` preserves duplicate keys (e.g. `?tag=a&tag=b`).

## Docs
- **#1147** — the example app called blocking sync SQLAlchemy inside `async def` handlers; switched them to `def` (Robyn runs sync handlers off the event loop) with a note on using an async driver if you prefer `async def`. (en + zh)
- **#1273** — the logging example used `Logger(app)`, which doesn't exist (`Logger.__init__` takes no args) — that's why request logs "didn't work". Replaced with a working access-log recipe using `before_request`/`after_request(request, response)` + a `ContextVar` timer. (en + zh)

## Tests
- **#495** — added a route that saves an uploaded file and a test asserting the saved bytes round-trip exactly (binary, non-UTF-8 payload).
- **#470** — added runtime lifecycle tests: `test_startup_event_fires` (asserts the startup event actually ran via an endpoint) and `test_shutdown_event_fires` (sends SIGINT to a dedicated server and asserts the shutdown handler ran).

All unit tests (9) and the new integration tests pass locally; `ruff`/`isort`/`ruff format`/`cargo fmt` clean.

Closes #1099
Closes #1147
Closes #1273
Closes #495
Closes #470
Closes #1192

## Summary by CodeRabbit

* **New Features**
  * Added query-parameter and header accessor methods (`keys`, `values`, `items`, `multi_items`) for enumerating and handling duplicates
  * Added `Headers.to_dict()` for a flat header dictionary
  * Enhanced lifecycle visibility for startup/shutdown behavior
* **Documentation**
  * Updated CRUD route examples to prefer synchronous `def` with the shown SQLAlchemy approach
  * Rewrote monitoring/logging documentation with request/access logging middleware guidance
* **Tests**
  * Added unit tests for `Headers` and query/header accessor behaviors
  * Extended integration coverage for lifecycle events and multipart file saving

### DIFF
--- a/docs_src/src/pages/documentation/en/example_app/modeling_routes.mdx
+++ b/docs_src/src/pages/documentation/en/example_app/modeling_routes.mdx
@@ -86,6 +86,8 @@ def delete_crime(db: Session, crime_id: int):
 
 Batman created various endpoints to manage crime data. These endpoints allowed the Gotham City Police Department to add, update, and retrieve crime data.
 
+> **A note on `def` vs `async def`.** The CRUD helpers above use **synchronous** SQLAlchemy, so the handlers below are deliberately plain `def` functions. Robyn runs synchronous handlers in a worker thread, which means a blocking database call there never stalls the event loop. If you would rather write `async def` handlers, pair them with an **async** database driver (for example SQLAlchemy's `asyncio` extension via `create_async_engine` and `await session.execute(...)`) so the database I/O is genuinely non-blocking. The one combination to avoid is a blocking, synchronous DB call inside an `async def` handler — that ties up the event loop for every other request.
+
 ```python {{ title: 'Setting up Routes' }}
 # __main__.py
 from robyn import Robyn
@@ -95,7 +97,7 @@ from sqlalchemy.orm import Session
 app = Robyn(__file__)
 
 @app.post("/crimes")
-async def add_crime(request):
+def add_crime(request):
     with SessionLocal() as db:
         crime = request.json()
         insertion = crud.create_crime(db, crime)
@@ -109,7 +111,7 @@ async def add_crime(request):
     }
 
 @app.get("/crimes")
-async def get_crimes(request):
+def get_crimes(request):
     with SessionLocal() as db:
         skip = request.query_params.get("skip", "0")
         limit = request.query_params.get("limit", "100")
@@ -118,7 +120,7 @@ async def get_crimes(request):
     return crimes
 
 @app.get("/crimes/:crime_id", auth_required=True)
-async def get_crime(request):
+def get_crime(request):
     crime_id = int(request.path_params.get("crime_id"))
     with SessionLocal() as db:
         crime = crud.get_crime(db, crime_id=crime_id)
@@ -129,7 +131,7 @@ async def get_crime(request):
     return crime
 
 @app.put("/crimes/:crime_id")
-async def update_crime(request):
+def update_crime(request):
     crime = request.json()
     crime_id = int(request.path_params.get("crime_id"))
     with SessionLocal() as db:
@@ -139,7 +141,7 @@ async def update_crime(request):
     return updated_crime
 
 @app.delete("/crimes/{crime_id}")
-async def delete_crime(request):
+def delete_crime(request):
     crime_id = int(request.path_params.get("crime_id"))
     with SessionLocal() as db:
         success = crud.delete_crime(db, crime_id=crime_id)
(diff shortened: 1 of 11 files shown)


sansyrox:
Addressed the review:
- **Path traversal (Major)** — the upload-save route now sanitizes the client filename via `os.path.basename(...)` (and skips ```/`.`/`..`), so absolute paths or `..` segments can't escape `upload_dir`.
- **Int coercion** — `skip`/`limit` query params are now `int(...)` before reaching the CRUD helper (en + zh).
- **Path-param syntax** — the delete route uses `:crime_id` consistently (was `{crime_id}`) (en + zh).

The #495 upload test still passes; ruff/isort/ruff format clean.

PR #1433, opened 2026-08-09
Fix robyn.env parser crash on blank lines and truncation of '=' values

### LINKED ISSUE #1426: robyn.env parser crashes with IndexError on a blank line
## Bug Description

`robyn/env_populator.py`'s `parser()` splits every non-comment line in `robyn.env` on `=` and yields the result unconditionally, and `load_vars()` unconditionally indexes `var[0]`/`var[1]`:

https://github.com/sparckles/Robyn/blob/main/robyn/env_populator.py#L17-L34

```python
for line in f:
    if line.startswith("#"):
        continue
    yield line.strip().split("=")
...
for var in variables:
    if var[0] in os.environ:
        ...
    else:
        os.environ[var[0]] = var[1]
```

A blank line (or a line that's just whitespace/a newline) produces `"".split("=")` → `['']`, a one-element list. `load_vars` then does `var[1]` → `IndexError: list index out of range`, crashing the app **before it even starts** — `load_vars()` runs unconditionally from `robyn/cli.py` (and from `BaseRobyn.__init__` when not run via the CLI) whenever a `robyn.env` file exists.

A related, quieter bug on the same line: `split("=")` has no `maxsplit`, so any value that itself contains `=` (e.g. a base64 secret like `SECRET_KEY=abc=123==`) gets truncated — `var[1]` becomes just `"abc"` instead of `"abc=123=="`.

## Steps to Reproduce

```python
import tempfile, os
from pathlib import Path
from robyn.env_populator import load_vars

d = tempfile.mkdtemp()
Path(d, "robyn.env").write_text("ROBYN_PORT=8080\n\nROBYN_HOST=127.0.0.1\n")  # note the blank line
load_vars(project_root=d)
```

```
  ...
IndexError: list index out of range
```

Reproduced directly against current `main`.

## Expected vs Actual

- **Expected:** blank lines in `robyn.env` are silently skipped (as `#`-comment lines already are), and values containing `=` are preserved in full.
- **Actual:** a blank line crashes app startup with an unhandled `IndexError`; values containing `=` are silently truncated.

## Suggested Fix

```python
for line in f:
    line = line.strip()
    if not line or line.startswith("#"):
        continue
    if "=" not in line:
        continue  # or log a warning about a malformed line
    yield line.split("=", 1)  # maxsplit=1 preserves '=' inside the value
```

## Additional Info

`unit_tests/test_env_populator.py` only covers a well-formed two-line file, so this path currently has no regression coverage. Found via a broader codebase audit while working on #485 (response compression). Robyn version: main branch.

eeshsaxena:
Fixes #1426

### Problem

`robyn/env_populator.py` `parser()` splits every non-comment line on `=` and yields the result unconditionally, and `load_vars()` indexes `var[0]`/`var[1]`:

```python
yield line.strip().split("=")
...
os.environ[var[0]] = var[1]
```

Two bugs:

1. A blank/whitespace-only line yields `['']`, and a line without `=` yields a one-element list. `load_vars` then hits `var[1]` -> `IndexError`, crashing the app **before it starts** (`load_vars` runs whenever a `robyn.env` exists, from `cli.py` and `BaseRobyn.__init__`).
2. `split("=")` has no `maxsplit`, so a value containing `=` (e.g. a base64 secret `SECRET_KEY=abc=123==`) is truncated to `abc`.

Reproduce:

```python
from pathlib import Path
from robyn.env_populator import load_vars, parser
Path("robyn.env").write_text("ROBYN_PORT=8080\n\nROBYN_HOST=127.0.0.1\n")
load_vars(variables=parser(config_path=Path("robyn.env")))   # IndexError
```

### Fix

In `parser()`: strip each line, skip blanks/comments and lines without `=`, and use `split("=", 1)` so values keep their `=` characters. `load_vars` now only ever receives well-formed 2-element pairs.

Before/after on the same input:

```
before: [['ROBYN_PORT','8080'], [''], [''], ['NO_EQUALS'], ['SECRET','abc','123','',''], ['ROBYN_HOST','127.0.0.1']]
after:  [['ROBYN_PORT','8080'], ['SECRET','abc=123=='], ['ROBYN_HOST','127.0.0.1']]
```

### Tests

Added tests in `unit_tests/test_env_populator.py` covering blank/whitespace/comment/valueless lines and `=`-in-value preservation, plus a `load_vars` blank-line smoke test. (The Robyn package import pulls in the compiled Rust extension, so I also exercised `env_populator` directly to confirm the crash before the change and clean parsing after.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment file parsing by ignoring blank lines, whitespace-only lines, and comments.
  * Added warnings for malformed entries without an equals sign or with empty keys.
  * Preserved additional equals signs within environment variable values.
  * Improved reliability when loading files containing empty or invalid lines.

* **Tests**
  * Added coverage for parsing edge cases, malformed entries, values containing equals signs, and environment cleanup.

### DIFF
--- a/robyn/env_populator.py
+++ b/robyn/env_populator.py
@@ -15,16 +15,13 @@ def parser(config_path=None, project_root=""):
         with open(config_path, "r") as f:
             for line in f:
                 line = line.strip()
-                # Skip blank lines and comments so they do not yield a
-                # malformed (key-only) pair that crashes load_vars().
                 if not line or line.startswith("#"):
                     continue
-                if "=" not in line:
+                key, sep, value = line.partition("=")
+                if not sep or not key:
                     logger.warning(" Ignoring malformed robyn.env line: %r", line)
                     continue
-                # maxsplit=1 so values containing '=' (e.g. base64 secrets)
-                # are preserved instead of being truncated at the first '='.
-                yield line.split("=", 1)
+                yield [key, value]
 
 
 # check for the environment variables set in cli and if not set them
--- a/robyn/env_populator.py
+++ b/robyn/env_populator.py
@@ -14,9 +14,17 @@ def parser(config_path=None, project_root=""):
     if config_path.exists():
         with open(config_path, "r") as f:
             for line in f:
-                if line.startswith("#"):
+                line = line.strip()
+                # Skip blank lines and comments so they do not yield a
+                # malformed (key-only) pair that crashes load_vars().
+                if not line or line.startswith("#"):
                     continue
-                yield line.strip().split("=")
+                if "=" not in line:
+                    logger.warning(" Ignoring malformed robyn.env line: %r", line)
+                    continue
+                # maxsplit=1 so values containing '=' (e.g. base64 secrets)
+                # are preserved instead of being truncated at the first '='.
+                yield line.split("=", 1)
 
 
 # check for the environment variables set in cli and if not set them
--- a/unit_tests/test_env_populator.py
+++ b/unit_tests/test_env_populator.py
@@ -38,3 +38,33 @@ def test_env_population(env_file):
     HOST = os.environ["ROBYN_HOST"]
     assert PORT == "8080"
     assert HOST == "127.0.0.1"
+
+
+def test_parser_skips_blank_and_malformed_lines(tmp_path):
+    env_path = tmp_path / "robyn.env"
+    # Blank line, whitespace-only line, comment and a valueless line must not
+    # crash the parser or produce malformed pairs.
+    env_path.write_text("ROBYN_PORT=8080\n\n   \n# a comment\nNO_EQUALS\nROBYN_HOST=127.0.0.1\n")
+    result = list(parser(config_path=env_path))
+    assert result == [["ROBYN_PORT", "8080"], ["ROBYN_HOST", "127.0.0.1"]]
+
+
+def test_parser_preserves_equals_in_value(tmp_path):
+    env_path = tmp_path / "robyn.env"
+    env_path.write_text("SECRET_KEY=abc=123==\n")
+    result = list(parser(config_path=env_path))
+    assert result == [["SECRET_KEY", "abc=123=="]]
+
+
+def test_load_vars_with_blank_line_does_not_crash(tmp_path):
+    env_path = tmp_path / "robyn.env"
+    env_path.write_text("ROBYN_PORT=8081\n\nROBYN_HOST=0.0.0.0\n")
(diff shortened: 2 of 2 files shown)


robyn/env_populator.py
```diff
@@ -14,9 +14,17 @@ def parser(config_path=None, project_root=""):
     if config_path.exists():
         with open(config_path, "r") as f:
             for line in f:
-                if line.startswith("#"):
+                line = line.strip()
+                # Skip blank lines and comments so they do not yield a
+                # malformed (key-only) pair that crashes load_vars().
```
sansyrox: let's get rid of these inline comments.

robyn/env_populator.py
```diff
@@ -14,9 +14,17 @@ def parser(config_path=None, project_root=""):
...
+                if not line or line.startswith("#"):
                     continue
-                yield line.strip().split("=")
+                if "=" not in line:
+                    logger.warning(" Ignoring malformed robyn.env line: %r", line)
+                    continue
+                # maxsplit=1 so values containing '=' (e.g. base64 secrets)
+                # are preserved instead of being truncated at the first '='.
```
sansyrox: same here

unit_tests/test_env_populator.py
```diff
@@ -38,3 +38,33 @@ def test_env_population(env_file):
     HOST = os.environ["ROBYN_HOST"]
     assert PORT == "8080"
     assert HOST == "127.0.0.1"
+
+
+def test_parser_skips_blank_and_malformed_lines(tmp_path):
+    env_path = tmp_path / "robyn.env"
+    # Blank line, whitespace-only line, comment and a valueless line must not
+    # crash the parser or produce malformed pairs.
```
sansyrox: same here

eeshsaxena:
Cleaned up per the review:

- Removed the inline comments in `env_populator.py` and the test.
- Empty-key lines like `=value` are now rejected too. Switched to `str.partition("=")` so a missing `=` and an empty key are handled by one check, which avoids the `os.environ[""]` crash CodeRabbit flagged. Added an `=NO_KEY` case to the test.
- The load_vars test now saves and restores any pre-existing `ROBYN_PORT`/`ROBYN_HOST` instead of deleting them.

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

sansyrox:
@eeshsaxena , could you please reopen this? This is still very much relevant. Just being incredibly busy with my new role

PR #1439, opened 2026-09-03
docs: add example for outbound HTTP requests from handlers

KAY53N:
## Description

Adds an example of calling another HTTP API from a Robyn handler.

Robyn does not bundle an outbound HTTP client. Handlers can use httpx, aiohttp, rusty-req, or any other async client. This PR adds `examples/http_client.py` using [rusty-req](https://github.com/KAY53N/rusty-req) (`pip install rusty-req`) and links it from the README.

## Summary

- Add `examples/http_client.py`:
  - `GET /todo/:id` — single GET via `fetch_single`
  - `POST /posts` — forward the JSON body upstream
  - `GET /fan-out` — concurrent batch via `fetch_requests` (`SELECT_ALL`)
  - `GET /fetch?url=` — GET an arbitrary URL
- Tags are only set on the batch requests, so results can be told apart
- README: point to the example and note that Robyn does not ship an outbound client

## PR Checklist

- [x] You have added relevant documentation

## Pre-Commit Instructions:

## Summary by CodeRabbit

* **New Features**
  * Updated the outbound HTTP example to use the async `httpx` client.
  * Demonstrated concurrent fan-out requests, JSON responses, timeouts, and handling of upstream HTTP failures.

* **Documentation**
  * Clarified that any async HTTP client can be used from an async handler.
  * Updated setup guidance and linked to the outbound HTTP example.
  * Removed client-specific installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ $ python3 app.py --open-browser
 
 You can add more routes to your API. Check out the routes in [this file](https://github.com/sparckles/Robyn/blob/main/integration_tests/base_routes.py) as examples.
 
+To call another HTTP API from a handler, see [`examples/http_client.py`](https://github.com/sparckles/Robyn/blob/main/examples/http_client.py). Robyn does not bundle an outbound client; that example uses [rusty-req](https://github.com/KAY53N/rusty-req) (`pip install rusty-req`).
+
 ### 🐍 Python Version Support
 
 Robyn is compatible with the following Python versions:
--- a/examples/http_client.py
+++ b/examples/http_client.py
@@ -1,0 +1,164 @@
+"""Call another HTTP API from a Robyn handler.
+
+Robyn serves incoming requests. It does not bundle an outbound HTTP client —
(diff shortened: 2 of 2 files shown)


sansyrox:
Thanks for the example. Two problems, fixed in 0f01272:

- Every error path returned a 2-tuple. Robyn only accepts `(body, headers, status_code)` and raises `Tuple should have 3 elements` otherwise, so an upstream 404 or a non-object JSON body turned into a 500 from this example. Those paths now build a `Response` with an explicit 502/400 and a JSON body, and connection errors or timeouts are caught and reported the same way.
- The README should not steer people towards one specific third-party client, so the note is now neutral and just points at the example. I also rewrote the example around httpx: most of the file was normalising rusty-req's mixed string/dict result shapes rather than showing anything about Robyn. The Robyn side is unchanged: a path-param GET, forwarding a JSON body, and a concurrent fan-out.

Verified every route against jsonplaceholder: 200 on `/todo/1`, 502 on `/todo/999999`, 400 on an array body, 200 on `POST /posts` and `/fan-out`.

PR #1442, opened 2026-09-13
Treat WARNING as a production log level

### LINKED ISSUE #1441: Production mode robyn --log-level WARN use depricated log level
### Bug Description

We set our log level in the application starting so we dont use the command line toggles, but effect is the same.

If we set the new logging target to WARNING Robyn stays in dev mode, while using the deprecated WARN log level it goes into full production mode. This should probably be fixed as its been deprecated for quite a while now.

You might also want to add some documentation on how to configure logging from Robyn, was a bit of fidling to get it working properly :D

To be precise, i understand Rust server needs WARN but you should pick up on pythons log level WARNING and send WARN to rust too, you already do for the deprecated WARN inside python logging :)

### Steps to Reproduce

1) start robyn standard
2) set logging level to WARNING in your code
3) initiate robyn startup
4) Enjoy the logs.

(if you do the same and set log level to WARN on step 2 it works as intended, check https://docs.python.org/3/library/logging.html#logging-levels for the current log levels)

### Your operating system

Linux

Other (specify below)

### Your Robyn version

latest

### Additional Info

I'm running python 3.14 has nothing to do with my python version WARN was depricated on 3.12 too.

It might be wise to support both WARN and WARNING as older pythons had WARN as the warning level while newer ones dont.

Shaisolaris:
## Description

This PR fixes #1441

## Summary

Production mode only checked for WARN. Python logging uses WARNING, and --fast already defaults to that name, so the verbose startup banner still appeared.

WARN and WARNING now both skip the banner. The message tells people to set WARNING, which is the name in the README.

Tests cover both CLI names and the --fast default.

## PR Checklist

Please ensure that:

## Pre-Commit Instructions:

- [ ] Ensure that you have run the pre-commit hooks on your PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recognition of production log-level settings, including `WARN`, `WARNING`, `ERROR`, and `CRITICAL`.
  - Ensured equivalent command-line log-level values, including numeric levels and different capitalization, behave consistently.
  - Updated verbose-mode guidance to use the standard `WARNING` log-level name.
  - Improved compatibility with logging configuration by normalizing log-level names.

- **Tests**
  - Added coverage for log-level recognition, command-line normalization, and default behavior when using fast or development modes.


sansyrox:
Thanks. Two gaps, fixed in af2e846:

- `is_production_log_level` only matched `WARN`/`WARNING`, so `--log-level ERROR` or `CRITICAL` still printed the "VERBOSE/DEBUG MODE" banner. It now resolves the name through `logging.getLevelName` and treats anything at or above `WARNING` as production, numeric levels included.
- `logging.basicConfig` rejects lower-case names, so `--log-level warning` crashed at startup before this check ever ran. `Config` now upper-cases the level.

I also replaced the `importlib` module loading in the test with a plain `from robyn.argument_parser import ...` like the other CLI tests, and moved it to pytest style with the extra levels covered.

The lint CI failure is unrelated: the workflow installs an unpinned ruff and 0.16.7 flags pre-existing issues on main.

PR #1443, opened 2026-09-13
fix: decode query and path parameters

### LINKED ISSUE #1438: Query parameters and path parameters are never percent-decoded
### What happens

No accessor on `Request` percent-decodes. Query values and path parameters are returned exactly as
they appear on the wire, so `%20` never becomes a space and `%3A` never becomes a colon.

Measured on Robyn 0.88.0, CPython 3.12, Linux:

```python
from robyn import Robyn, Request

app = Robyn(__file__)

@app.get("/q")
def q(request: Request):
    p = request.query_params
    return str({"get": p.get("v", None), "to_dict": p.to_dict(), "items": dict(p.items())})

@app.get("/p/:seg")
def p(request: Request):
    return str(request.path_params["seg"])

app.start(host="127.0.0.1", port=8080)
```

```
GET /q?v=a%20b
  get("v", None) -> 'a%20b'
  to_dict()      -> {'v': ['a%20b']}
  items()        -> {'v': 'a%20b'}

GET /q?v=2026-01-01T00%3A00%3A00Z
  get("v", None) -> '2026-01-01T00%3A00%3A00Z'

GET /p/attempt%3A1
  path_params["seg"] -> 'attempt%3A1'
```

### Expected

`a%20b` should arrive as `a b`, and `attempt%3A1` as `attempt:1`. A space is the most common
percent-encoded character in a query string, so this affects almost any client that encodes
correctly.

### Where it comes from

`src/types/request.rs` builds the query map by splitting the raw query string on `&` and `=` and
storing the resulting substrings directly, with no decoding step. The path-parameter map is
populated the same way.

For comparison, actix-web's own `web::Query` extractor decodes, so this is Robyn's layer rather
than anything inherited.

### Workaround

Call `urllib.parse.unquote` on every value at the edge of each handler.

### Environment

Robyn 0.88.0 (current release at time of filing), CPython 3.12, Linux.

KayanoLiam:
## Description

Fixes #1438.

HTTP query accessors and path parameters currently return percent escapes verbatim: `?v=a%20b` yields `a%20b`, and `/p/attempt%3A1` yields `attempt%3A1`. Decode these values once so handlers receive `a b` and `attempt:1`.

## Summary

- Decode query names and values using form encoding rules (`+` is a space), preserving repeated values, their order, bare flags, empty keys, and empty fields.
- Decode captured path values after route matching in both handlers and route middleware. Path `+` stays literal, `%2F` does not change route matching, and `%252F` becomes `%2F`.
- Add 47 real HTTP regression/control cases covering sync/async handlers, query accessors, Unicode, invalid UTF-8, catch-all routes, and middleware.
- Document the behavior in the English and Chinese request reference. The two direct decoding dependencies already exist in the lockfile; no dependency versions are upgraded.

## PR Checklist

- [x] The PR contains a descriptive title.
- [x] The PR contains a descriptive summary of the changes.
- [x] Build and test the changes before submitting.
- [x] Add relevant documentation.
- [x] Add relevant integration tests.

## Pre-Commit Instructions

- [x] Run the pre-commit hooks on the changed Python files.

## Validation

Using the project's Python 3.12.11 environment on macOS ARM64:

- `uv run --no-sync maturin develop --locked`
- `uv run --no-sync pytest -q --tb=short --show-capture=no` — 587 passed, collecting both `integration_tests/` and `unit_tests/`.
- `uv run --no-sync cargo test --locked`
- `uv run --no-sync cargo clippy --locked`
- `cargo fmt --check`
- `uv run --no-sync pre-commit run --files integration_tests/base_routes.py integration_tests/test_get_requests.py`

The initial HTTP regression cases reproduced the original bug before the runtime change. Existing PyO3/deprecation warnings remain in the Rust build.

## Summary by CodeRabbit

- **New Features**
  - Query parameters now consistently support standard form decoding, including spaces, plus signs, UTF-8 characters, repeated values, and malformed input handling.
  - Path parameters are decoded after route matching, including values received by middleware, handlers, WebSocket requests, and testing tools.
  - Encoded slashes and other special characters are preserved correctly within matched parameters.

- **Documentation**
  - Added English and Chinese guidance on decoding behavior, repeated query values, invalid input, and safely using decoded path parameters in filesystem paths.


src/routers/http_router.rs
```diff
@@ -46,8 +47,12 @@ impl Router<(FunctionInfo, HashMap<String, String>), HttpMethod> for HttpRouter
         // Trying route matching just once.
         if let Ok(res) = table_lock.at(route) {
             let mut route_params = HashMap::new();
+            // Decode captures after matching so encoded slashes stay within one segment.
```
sansyrox: let's get rid of this comment?

src/routers/middleware_router.rs
```diff
@@ -56,8 +57,12 @@ impl Router<(Vec<FunctionInfo>, HashMap<String, String>), MiddlewareType> for Mi
         let table_lock = table.read().ok()?;
         let res = table_lock.at(route).ok()?;
         let mut route_params = HashMap::new();
+        // Apply the same decoding to route middleware and handler captures.
```
sansyrox: same here

src/types/request.rs
```diff
@@ -135,13 +135,12 @@ impl Request {
...
 
         if !req.query_string().is_empty() {
-            let split = req.query_string().split('&');
-            for s in split {
-                let path_params = s.split_once('=').unwrap_or((s, ""));
-                let key = path_params.0.to_string();
-                let value = path_params.1.to_string();
-
-                query_params.set(key, value);
+            // Preserve empty fields as well as repeated keys when decoding form data.
```
sansyrox: same here?

sansyrox:
Overall, very nice PR. GG!

sansyrox:
Thanks, this is solid. I pushed d58ac22 with a few follow-ups rather than going back and forth:

- WebSocket handlers still went through the old raw `split('&')` parser, so `websocket.query_params` stayed percent-encoded while HTTP requests got decoded. Both now use a shared `QueryParams::from_query_string`. Side effect: empty fields such as `?a=1&` no longer produce a `''` key; the two test cases that locked that in are updated.
- The capture-decoding loop was duplicated in `http_router.rs` and `middleware_router.rs`. It lives in `routers::decode_route_params` now.
- `robyn.testing.TestClient` copied its regex captures verbatim, so tests written against it diverged from the real server for any encoded segment (`/items/a%2Fb` gave `a%2Fb` there and `a/b` on the server). It decodes the same way now, with a unit test.
- The `/params/decoded/:value` test passed state to the handler through `request.body`; switched to a header like the other `before_request` tests.
- Docs: added that a decoded `:param` can contain `/` or `..` (same as a catch-all) so it must be sanitised before filesystem use, and that WebSocket query params follow the same rules. The three inline comments from my earlier review are gone.

Ran the unit suite, the full integration suite, clippy and `cargo test` locally. The lint CI failure is unrelated to this PR: the workflow installs an unpinned ruff and 0.16.7 reports 313 pre-existing issues on main.

